PR #117, opened 2012-01-04
Auto Scaling

mightypenguin:
Config options aren't implemented and there are rough edges but auto scaling to max possible size should work for vnc.html and vnc_auto.html.

Tested on Chrome16/ Ipad2(IOS5) /Iphone3 (IOS4).

I figured you probably wouldn't want it just like this so it's more of a demo of what's possible.
Not tested with flash or other weirdness.

Scaling of type "fit" which uses all available width/height can't be implemented until scaling functions for the mouse have separate scaling values for x/y.

### DIFF
--- a/include/ui.js
+++ b/include/ui.js
@@ -113,6 +113,9 @@ start: function(callback) {
     //         }
     //    };
 
+    //UI.initSetting('fullsize',false);
+    UI.initSetting('scale',true);
+
     // Show mouse selector buttons on touch screen devices
     if (UI.isTouchDevice) {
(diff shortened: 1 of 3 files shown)


DirectXMan12:
@tobegit3hub : I may have some time to update this patch shortly.  We're also looking at getting #271 merged, which informs servers which support desktop resizing (e.g. [TigerVNC](http://tigervnc.org/)) about the resize.

DirectXMan12:
Closing this in favor of #451, which is an updated version of this patch.

PR #192, opened 2012-09-25
Disable IME while noVNC is working

takaha:
IME -- Input Method Editor -- will intercept all the key events and make some of the events broken.
With this patch, IME is automatically disabled when noVNC gets the focus, and you don't have to worry if some key strokes happen to activate IME again.

This patch only works for IE, FireFox and Opera. But there is no harm for Chrome and Safari.
Even after applying this patch, if you want to use Chrome or Safari, or you have overwritten the css, you need to turn off IME manually.

### DIFF
--- a/include/base.css
+++ b/include/base.css
@@ -129,6 +129,7 @@ html {
   background-color:#313131;
   border-bottom-right-radius: 800px 600px;
   /*border-top-left-radius: 800px 600px;*/
+  imeMode: disabled;
 }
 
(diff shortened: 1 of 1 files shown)


samhed:
First of all, this style is only supported by Firefox and IE, and in Firefox it seems to only apply to text fields.

Secondly, I think the general consensus is that this style could do more harm than good, and probably shouldn't have been implemented in the first place.

PR #213, opened 2012-11-08
some softkey functionalities

yamt:


samhed:
I'm assuming you wanted this on touch devices and the like where you don't have modifier keys on the on-screen keyboards. Anyway, we merged #284 instead. Closing this.

Issue #194, opened 2012-10-02
IE9 leaves "ReferenceError:  WebSocket is not defined"


takaha:
noVNC running on IE9 often fails to connect to a VNC server leaving a message "ReferenceError:  WebSocket is not defined."
And sometimes I see  a message "ReferenceError:  swfobject is not defined."

I think noVNC has to wait until both swfobject.js and web_socket.js are loaded and  INITIALIZED before calling them.

PR #217, opened 2012-12-10
Fixing ReferenceError errors

renan:
Yet Reference Error errors like #194, #202, etc.

As a rule of thumb, initialize your var before declaring or using it.

:+1:


samhed:
Not relevant anymore, closing.

PR #224, opened 2013-01-31
Allow specifying a custom default cursor

jalfd:
Makes it possible to specify a default cursor when you create the RFB object.

When Local Cursor is disabled (or unsupported), most VNC clients change the client-side cursor to a dot. With this commit, something similar can be easily done with noVNC in browsers that don't support Local Cursor.


DirectXMan12:
If you are still interested in this PR, please update it to work with the current noVNC code.

PR #271, opened 2013-06-20
Support the 'ExtendedDesktopSize' and the 'SetDesktopSize' encodings

samhed:
Somewhat related to noVNC issue #117, but instead of scaling it is using the RFB protocol message 'SetDesktopSize' from the client to the server.

__Edit:__
This patch adds support in noVNC for the encodings 'ExtendedDesktopSize' (-308) and 'SetDesktopSize' (251). If the 'ExtendedDesktopSize' encoding is supported by the server, this patch allows for automatic resizes of the session shown in noVNC to fit the browser window. Note that the server still has the option of denying the resize request from noVNC.

When supported, this is a client-initiated resize. For this to work the 'ExtendedDesktopSize' encoding is required on the server. Aside from client-initiated resizes this encoding brings support from the server for multiple monitors (Note that mulitple monitors are not supported by noVNC). Documentation can be found [here](http://sourceforge.net/p/tigervnc/code/HEAD/tree/rfbproto/rfbproto.rst#id88).

The resize is triggered by changes to the browser window's size. Note that this will have the effect that if multiple viewers are connected to one session on the VNC server the latest change will decide the size of the session as viewed on all clients.

The 'ExtendedDesktopSize' encoding is fairly new to the protocol and I know that many VNC servers have not yet implemented this, but I do know that TigerVNC's server have implemented it.

Summary:
* This requires the VNC server to support the 'ExtendedDesktopSize' and the 'SetDesktopSize' encodings.
* Automatically resizes the session to fit the browser window
* TigerVNC's server have the required encodings.


samhed:
Okay, I'm finally done with this! :tada: 

I'll give people a chance to review the patch over the weekend before I merge it.

samhed:
Had forgot to test vnc_auto.html, works now after I added two quick fixes.

DirectXMan12:
:open_mouth: I'll try to take a look at this tomorrow.  Thanks for finishing this up -- it'll be a good feature to have.

samhed:
I'd love to put this PR behind me, it's been nagging me for one and a half years now :) Did you get a chance to have a look?

vnc_auto.html
```diff
@@ -198,9 +222,10 @@
                            'local_cursor': WebUtil.getQueryVar('cursor', true),
                            'shared':       WebUtil.getQueryVar('shared', true),
                            'view_only':    WebUtil.getQueryVar('view_only', false),
-                           'onUpdateState':  updateState,
+                           'updateState':  updateState,
```
DirectXMan12: This should still be onUpdateState (otherwise the status bar doesn't function correctly)

vnc_auto.html
```diff
@@ -80,7 +80,21 @@
                            "jsunzip.js", "rfb.js", "keysym.js"]);
 
         var rfb;
+        var resizeTimeout;
 
+
+        function UIresize() {
+            var innerWidth = window.innerWidth;
+            var innerHeight = window.innerHeight;
+            var controlBarHeight = $D('noVNC_status_bar').scrollHeight;
```
DirectXMan12: Should this really be scrollHeight (which would count any non-visible content)?  Perhaps `offsetHeight` instead (rendered height including padding and borders.

vnc_auto.html
```diff
@@ -80,7 +80,21 @@
...
+            var borderHeight = 5;
```
DirectXMan12: Can you clarify where this value comes from (in a comment or something)?

include/ui.js
```diff
@@ -44,6 +57,15 @@ var UI;
             WebUtil.initSettings(UI.start, callback);
         },
 
+        onresize: function (callback) {
+            var innerWidth = window.innerWidth;
+            var innerHeight = window.innerHeight;
+            var controlBarHeight = $D('noVNC-control-bar').scrollHeight;
```
DirectXMan12: Ditto here as in vnc_auto.html

include/rfb.js
```diff
@@ -1835,12 +1866,53 @@ var RFB;
...
+            this._FBU.bytes = 1;
+            if (this._sock.rQwait("ext_desktop_size", this._FBU.bytes)) { return false; }
+
+            this._supportsSetDesktopSize = true;
+            var number_of_screens = this._sock.rQpeek8();
+
+            this._FBU.bytes = 4 + (number_of_screens * 16);
+            if (this._sock.rQwait("ext_desktop_size", this._FBU.bytes)) { return false; }
+
+            this._sock.rQshift8();  // number-of-screens
```
DirectXMan12: you should use rQskipBytes here -- it's more efficient.

include/rfb.js
```diff
@@ -1835,12 +1866,53 @@ var RFB;
...
+            this._sock.rQshift8();  // padding
+            this._sock.rQshift16(); // padding
+
+            for (var i=0; i<number_of_screens; i += 1) {
+                // Save the id and flags of the first screen
+                if (i == 0) {
+                    this._screen_id = this._sock.rQshift32();    // id
+                    this._sock.rQshift16();                      // x-position
```
DirectXMan12: same here with rQskipBytes

DirectXMan12:
@samhed : the actual functionality looks like a charm, but I have a couple of general comments, as well as some inline comments.

1. It would be nice to have an option in `vnc.html` to disable the automatic resize, since some people with small screens may still with to view a full-sized desktop.
2. Can you squash the "fix error from previous commit in this pull request" commits back into their respective original PR?  It seems to me like this commit should be broken up into around three commits: one with the changes to `display.js`, the next with the changes to `rfb.js`, and the last with the changes to `ui.js`, `vnc.html`, and `vnc_auto.html`.  This way, we can keep our history relatively clean.
3. Can you add in a couple of basic tests for this (I can help with this part, if need-be)?  We don't want to accidentally break anything in the future.

samhed:
Thank you @DirectXMan12 for taking your time to go through this! I agree with your comments and will get to work.

> It would be nice to have an option in vnc.html to disable the automatic resize, since some people with small screens may still with to view a full-sized desktop.

Yes, I was contemplating doing this.. the reason I didn't was that since not all VNC servers have this functionality. Do you think that an option called "Resize remote session to fit the browser window" would be confusing?

> Can you squash the "fix error from previous commit in this pull request" commits back into their respective original PR?

I'm not 100% sure what you mean by "original PR" here, but yes I will clean it up. :+1:

vnc_auto.html
samhed: The noVNC_container element is for some reason a few pixels higher than the noVNC_canvas. It is 4 pixels higher in Chrome and 5 pixels higher in Firefox (according to the built in HTML inspectors). I have no idea why.. Me and a colleague spent half a day trying to figure this shit out but eventually gave up :unamused:

If someone has some insight into this I'd love to know!

DirectXMan12:
> Yes, I was contemplating doing this.. the reason I didn't was that since not all VNC servers have this functionality. Do you think that an option called "Resize remote session to fit the browser window" would be confusing?

Perhaps look at it the other way around: have an option that says "Disable remote session resize to fit browser window".  That might be a bit less confusing (not checked just means we haven't explicitly disabled it, not that it's always supported).  Another option (albeit a bit more complicated) would be a dropdown with three options: "off", "ask", and "always try".  "Off" would mean "never resize"; "Ask" (the default) would mean that if a server sent ExtendedDesktopSize, ask the user if they would like to send SetDesktopSize in the future; and "always try" would mean always use it if the server supported it (we may or may not need this -- it may be sufficient just to have "ask").  This also paves the road for supporting other resize methods, such as automatic scaling (just add another dropdown option).

> I'm not 100% sure what you mean by "original PR" here, but yes I will clean it up. :+1:

whoops, I mean to say "back to their original commit" ;-)

samhed:
> Perhaps look at it the other way around: have an option that says "Disable remote session resize to fit browser window". That might be a bit less confusing (not checked just means we haven't explicitly disabled it, not that it's always supported). Another option (albeit a bit more complicated) would be a dropdown with three options: "off", "ask", and "always try". "Off" would mean "never resize"; "Ask" (the default) would mean that if a server sent ExtendedDesktopSize, ask the user if they would like to send SetDesktopSize in the future; and "always try" would mean always use it if the server supported it (we may or may not need this -- it may be sufficient just to have "ask"). This also paves the road for supporting other resize methods, such as automatic scaling (just add another dropdown option).

Hm, that's true, your first option might be a bit more clear for users in some cases. But I don't like the idea of having an option which you can check to "disable" something. I feel the second option seems a bit too much, I would personally prefer something very simple-looking.

noVNC is after all a VNC client without attachments to any specific VNC servers. The goal should be to be able to handle messages any server following the RFB-protocol sends our way. I think I lean towards simply labeling the option "Resize to window" in the same short style as the rest of the options.

DirectXMan12:
> I think I lean towards simply labeling the option "Resize to window" in the same short style as the rest of the options.

That works for me.  You may want to make it "resize desktop to window" or "resize remote to window" so it's not unclear what it does if we merge a scaling patch as well (with "resize to window" vs "scale to window", it's unclear how they're different).

samhed:
That sounds good! I will create a new PR with cleaned up commits. I will start looking on the tests on monday.

samhed:
Hm.. when I want to try to run some of my tests I get this:

    [samuel@samuel-80 noVNC]$ ./tests/run_from_console.js -t tests/test.rfb.js 
    Running tests tests/test.rfb.js using provider SpookyJS (CapserJS on PhantomJS)
    ERROR: Error: Child terminated with non-zero exit code 1

And nothing more, no hints!

If i run 'PATH=$PATH:node_modules/karma/bin npm tests' I do get more output:

    FAILED TESTS:
      Remote Frame Buffer Protocol Client
        Public API Basic Behavior
          #setDesktopSize
            ✖ should send the request with the given size
              PhantomJS 1.9.8 (Linux)
            AssertionError: expected { Object (_websocket, _rQ, ...) } to have sent the data [ Array(24) ], but it actually sent [ Array(24) ]
                at /home/samuel/noVNC/node_modules/chai/chai.js:925
                at /home/samuel/noVNC/tests/assertions.js:22
                at /home/samuel/noVNC/node_modules/chai/chai.js:3638
                at /home/samuel/noVNC/tests/test.rfb.js:226

I'm not sure what I can do to actually understand what's wrong! :worried:

DirectXMan12:
sometimes, the console test runners leave a bit to be desired.  Try `./tests/run_from_console.js -g -o -t tests/test.rfb.js`, which will open up the generated HTML file in your browser of choice (or skip the `-o` option and copy and paste the file name yourself)

samhed:
> expected to have sent the data [ Array(24) ], but it actually sent [ Array(24) ]

When the data type is the same but the assertion fails like in this case, should we not have displayed the difference in the data?

Hehe well, I guess I can resort to debugging my tests through "console.log" and the javascript console in the browser at least!

DirectXMan12:
> When the data type is the same but the assertion fails like in this case, should we not have displayed the difference in the data?

I didn't get that fancy when I wrote the assertion ;-).  The actual browsers often display the full arrays (instead of Phantom's "[ Array(24) ]"), so running with the `-g -o` options should help.

Issue #275, opened 2013-07-04
Keyboard input does not work in Chrome on Android


samhed:
Platform: Android 4.2.2; Build/JDQ39; Nexus 7
Browser: Chrome 27.0.1453.90
Used adb logcat for debugging via USB.

I ran tests/input.html and noticed that instead of sending the key event (like it does on the iPad for example) the log shows this message:

I/chromium( 8658): [INFO:CONSOLE(110)] "Key event (keyCode = 229) not found on keyDownList", source: http://<mylocaldomain>:6080/include/util.js(110)

The following keys work as intended: 1,2,3,4,5,6,7,8,9,0,BACKSPACE,ENTER

Running other tests shows that the problem might not be in noVNC but rather in Chrome on Android. I found this issue for the chromium projects that confirm this:
http://code.google.com/p/chromium/issues/detail?id=118639

Issue #251, opened 2013-04-17
Keyboard input does not work in Firefox on Android


astrand:
Keyboard input does not work in Firefox 20.0.1 on Android. Backspace seems to work, but all other keys are lost somewhere.

doggdor:
How to install the fixes?

samhed:
@doggdor see my answer in #301

Issue #126, opened 2012-02-08
Keyboard input does not work on FF/Asus EEE Pad


astrand:
Using the Android based EEE Pad Transformer. Since the native browser does not support Websockets, I'm using Firefox 9.0 instead. Works, but the keyboard is basically unusable:
- It is only possible to type numbers. No other characters works. 
- The Enter key behaves like Tab. 

On this device, you can switch between "Asus" and "Android" keyboard. Both behaves the same. 

If I connect the physical keyboard (Dock), Enter and Backspace are the only keys that works.

PR #301, opened 2013-09-17
Fixes the onscreen keyboard in Chrome on Android

samhed:
Fixes Issue #275.

As seen in #275, using the onscreen Android keyboard in noVNC in Chrome does not work as of now in noVNC. This is because keyPress events does not get sent in Chrome on Android. As seen in this [thread](https://code.google.com/p/chromium/issues/detail?id=118639) it seems like the Chromium developers do not consider this behavior a bug and they recommend using input events instead.

This patch allows noVNC to catch input events from the hidden keyboardinput field which was used to give focus in order to bring up the onscreen keyboard. Input events instead of keyPress events brings a few pros and a few cons.. some keys that does not generate any input will require special handling but on the upside you automatically support the input of more than one character at a time, from for example gesture typing or "swipe" as well as auto-complete.

In cases where keyboard input works without this patch the code will not be invoked since the keyPress events prevents input from being entered into the input field. We will therefore never get input events at the same time as keyPress events.

Below follows a list of what I have tested and which events each case will trigger:

**Android 4.3, Nexus 7:**
Chrome 29 - input events
Firefox 23 - keypress events

**iOS 6.1, iPad Mini:**
Safari 6 - keypress events
Chrome 29 - keypress events

This patch is only affecting Chrome on Android as far as I can see. Unfortunately I do not have access to any windows touch device to test on.

And by the way, issues #126 and #251 should also be closed since the problem has been fixed in newer Firefox versions.

### DIFF
--- a/include/ui.js
+++ b/include/ui.js
@@ -155,7 +155,8 @@ addMouseHandlers: function() {
     $D("noVNC_mouse_button2").onclick = function () { UI.setMouseButton(4); };
     $D("noVNC_mouse_button4").onclick = function () { UI.setMouseButton(0); };
     $D("showKeyboard").onclick = UI.showKeyboard;
-    //$D("keyboardinput").onkeydown = function (event) { onKeyDown(event); };
+
+    $D("keyboardinput").oninput = UI.keyInput;
     $D("keyboardinput").onblur = UI.keyInputBlur;
 
     $D("sendCtrlAltDelButton").onclick = UI.sendCtrlAltDel;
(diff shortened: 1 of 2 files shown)


doggdor:
How to install the patch? My Android Chrome and firefox is still not working. The keyboard doesnt appear.

samhed:
There is no patch you have to install, it is merged into noVNC. Both Chrome and Firefox are definitely working. Can you please describe what you are doing and what you are seeing? Screenshots could be helpful for example.

doggdor:
First of all thanks to kanaka for the the program. It is really useful for me. In general I love github.com! 

My client is a raspberry (raspbian). To install noVNC I just followed some old instructions from 
[sudo git clone git://github.com/kanaka/noVNC]

By typing "sudo git clone git://github.com/kanaka/noVNC" there must be downloaded the newest version isn't it?

Thank you for the fast response!

doggdor:
Ah! I read about websockify, I think I have to install it

samhed:
> By typing "sudo git clone git://github.com/kanaka/noVNC" there must be downloaded the newest version isn't it?

yes

doggdor:
The keyboard doesnt want to open.. is it a problem of tightvncserver?

doggdor:
ok everything is right, the keyboard can be activitaed manually by using "vnc.html"

samhed:
I didn't think of vnc_auto.html since I rarely use it myself, apologies! Yes, there is a few features you miss out on if you don't use vnc.html.

PR #305, opened 2013-09-25
Allow passing queryparams to websocket

grimpy:
Adds helper function to convert object into querystring
Adds passing of token to websocket in vnc_auto.html


DirectXMan12:
If you are still interested in the PR, please update it to work with the current version of noVNC.

PR #355, opened 2014-04-04
Fix rescale bug

ardevelop:


samhed:
We are not using this way of rescaling anymore. Closing

PR #362, opened 2014-05-02
Simple rpm packaging for redhat and clones.

dragz:
I've created a simple rpm build setup in redhat/.  Feel free to include it into the mainline.

(thanks for a great piece of software)


DirectXMan12:
We've removed all distro-specific packaging from noVNC, and instead are relying on the distros themselves to produce suitable packaging.

PR #369, opened 2014-05-28
Let mouse active even if there is touch

terraloader:
I ran into the problem on my laptop: it is touch-enabled, but im working with 2 external monitors. So until now only chrome is able to receive touch-events. But if there are touch-events the normal mouse wouldn work at the same time. FIXED with this little commis.

### DIFF
--- a/include/input.js
+++ b/include/input.js
@@ -355,14 +355,14 @@ that.grab = function() {
         Util.addEvent(window, 'touchend', onMouseUp);
         Util.addEvent(c, 'touchend', onMouseUp);
         Util.addEvent(c, 'touchmove', onMouseMove);
-    } else {
+    } //else {
         Util.addEvent(c, 'mousedown', onMouseDown);
         Util.addEvent(window, 'mouseup', onMouseUp);
         Util.addEvent(c, 'mouseup', onMouseUp);
         Util.addEvent(c, 'mousemove', onMouseMove);
         Util.addEvent(c, (Util.Engine.gecko) ? 'DOMMouseScroll' : 'mousewheel',
                 onMouseWheel);
-    }
+    //}
 
     /* Work around right and middle click browser behaviors */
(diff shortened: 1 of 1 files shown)


kosmasgiannis:
Any progress on this one?

samhed:
I don't know if @terraloader ever got around to do those tests. I do however believe that what @kanaka said is true.

kosmasgiannis:
Recent tests have shown that mouse click events are not triggered on devices with a touch-screen and mouse.

My tests included : 
* SONY VAIO Tap21, Windows 8.1
* HP Pavilion TouchSmart 10, Windows 8.1
* Microsoft Surface Pro 3, Windows 8.1 Pro

In all cases Chrome 40.0.2214.111 was used.

kanaka:
@kosmasgiannis I think Windows may be the exception. I'd be interested in whether iOS and Android tablets (Chrome, firefox and/or old Android browser) generate both touch and clicks

rabidaudio:
I'm having the same issue on Arch Linux x64 and Chrome 45.0.2454.101 with a touch screen, so it's not isolated to Windows/Surface. No problems on Firefox.

sedlund:
Patch fixed for me on Chromebook Flip with touch screen

zarmhast:
@kanaka,
tested #619 which has the same solution, but is based on a newer revision, on an android device. Yes, in general both events are generated touch and mouse, unless you process the touch event. noVNC prevents propagation on all touch and mouse events hence every touch generates just touchstart and touchend. (just logged every invocation of _handleMouseButton making note of when, where and whether it is a touch or mouse event, only touch events showed up). Tested using chrome and firefox.

This is the reason only one type of event is fired:
```
Util.stopEvent(e);
return false;
```

I would suggest to close this PR and once #619 is accepted all related bugs can be marked as resolved.

PR #427, opened 2015-01-08
Unregister event listeners from websock.

jswanner:
Prevents possible memory and event notification leaks when tearing down
connection and reestablishing a new one.

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -197,6 +197,7 @@ var RFB;
             } else {
                 this._fail("Server disconnected" + msg);
             }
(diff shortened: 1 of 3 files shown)


DirectXMan12:
LGTM :+1:

@kanaka, @samhed , @astrand : any comments?

samhed:
:+1:

kanaka:
@DirectXMan12 change itself looks okay. However, I do note that websock.js has gotten significantly out of sync with websockify version (ever since the big refactor). Probably want to sync that at some point.

PR #432, opened 2015-01-26
Replace custom getPosition algorithms with getBoundingClientRect.

nbibler:
The [`getBoundingClientRect()`](https://developer.mozilla.org/en-US/docs/Web/API/Element.getBoundingClientRect) call is native JavaScript and supported since Chrome 1, Firefox 3, Safari 4, IE4 (excluding width and height, which are not used here). It works more consistently than the existing implementation, especially when working with position static and fixed element hierarchies.

All tests still pass (although there are no tests covering this function). We did manually test in our own application with these changes, as well as in the packaged `vnc.html`.

### DIFF
--- a/include/util.js
+++ b/include/util.js
@@ -433,72 +433,11 @@ Util.load_scripts = function (files) {
 };
 
 
-// Get DOM element position on page
-//  This solution is based based on http://www.greywyvern.com/?post=331
-//  Thanks to Brian Huisman AKA GreyWyvern!
-Util.getPosition = (function () {
+Util.getPosition = function(obj) {
     "use strict";
-    function getStyle(obj, styleProp) {
-        var y;
-        if (obj.currentStyle) {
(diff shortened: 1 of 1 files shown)


DirectXMan12:
LGTM :+1:

@kanaka : any objections?

kanaka:
No objections at all. As long as we tested this across the major browsers we support (for example, we should make sure mobile isn't subtly different) I think this is a great change.

DirectXMan12:
I'll see if I can do a bit of testing next week.

DirectXMan12:
Ok, I did some testing on:

- Chrome 40  (on Android)
- Chromium 41 (on Linux)
- Firefox 38 nightly (on Linux)

All of which look fine.  It looks like `getBoundingClientRect` has been around for a while, and I'm not seeing anything in the way of known issues (jQuery uses it as-is).

PR #434, opened 2015-01-31
Update tests/input.html to include keysym.js instead of keysymdef.js

jalfd:
As pointed out [here](https://github.com/kanaka/noVNC/issues/21#issuecomment-72259534), the input.html test page had broken after the recent commits moving from keysymdef.js to keysym.js.

The problem simply seems to be that the page includes the wrong file.


DirectXMan12:
Actually, both `keysym.js` and `keysymdef.js` are needed.  `keysymdef.js` includes lookups for unicode codepoints, while `keysym.js` includes explicitly named key constants.

PR #443, opened 2015-02-06
keysym.js needs inclusion in debian's install list

fzylogic:
It's used but not being packaged.  Let's fix that!


samhed:
Thank you

PR #444, opened 2015-02-07
Support ExtendedDesktopSize and SetDesktopSize

samhed:
Cleaned up version of #271 

This feature is divided into 4 commits:
1: display.js
2: rfb.js
3: ui.js, vnc.html & vnc_auto.html
4: tests

### DIFF
--- a/include/display.js
+++ b/include/display.js
@@ -111,54 +111,12 @@ var Display;
 
     Display.prototype = {
         // Public methods
-        viewportChange: function (deltaX, deltaY, width, height) {
+        viewportChangePos: function (deltaX, deltaY) {
             var vp = this._viewportLoc;
-            var cr = this._cleanRect;
-            var canvas = this._target;
 
             if (!this._viewport) {
-                Util.Debug("Setting viewport to full display region");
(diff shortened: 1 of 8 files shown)


DirectXMan12:
LGTM :+1:  Just needs tests.  It will be great to have this.

P.S. For future reference, you don't have to open a new PR -- you can just edit the existing branch that the PR is based off of (`git rebase -i`, or something like that locally, and then `git push --force`)

samhed:
There we go, I think these tests should be sufficient, the changes in display.js should be covered by existing tests. I think ui changes are difficult to write unit tests for since they are closely dependent on HTML elements?

>P.S. For future reference, you don't have to open a new PR -- you can just edit the existing branch that the PR is based off of (git rebase -i, or something like that locally, and then git push --force)

Will do!

DirectXMan12:
The unit tests look good.  I caught one last thing right before I went to merge, however.  If you can fix that, I'll merge the patch.

samhed:
Good catch, thank you! Should all be good now.

PR #448, opened 2015-02-16
Remove local copies of websockify

### LINKED ISSUE #433: please sync websockify
lots of bugfixes there... ;-)

DirectXMan12:
This commit removes local copies of websockify.
Instead `utils/launch.sh` performs the following logic:

If `utils/websockify` exists, use `utils/websockify/run` (if the latter
does not exist, or is not executable, fail, since this is probably a
mistake).

Otherwise, check to see if websockify is installed somewhere (try
`which websockify`).  If it is, use that.  Otherwise, clone
websockify from github, and tell git to ignore that directory.

Packaged versions of noVNC should simply list websockify as a
requirement.  The debian packaging has been updated to reflect
this.

Closes #433


kanaka:
@DirectXMan12 any thoughts on how include/websock.js should be handled in this case since that's really provided by websockify too?

DirectXMan12:
@kanaka : that's a bit tricky -- IMO, `websock.js` belongs in the noVNC repository.

Among other things, I think the noVNC repository should have all the files required to host the noVNC client itself on a web server.  The problem then becomes that we can't create a circular dependency between noVNC and websockify.

We could fork`websock.js` out into a third repository and use git subtrees (like submodules, but always present in the repository w/o requiring recursive clones, etc).  IMO, this is a bit overkill.
This, essentially, leads to the conclusion that we should continue just copying the file over every once and a while.

What do you think?

kanaka:
@DirectXMan12 I'm fine with copying websock.js. It doesn't change that much compared to websockify itself (although when we switch to all binary types it might need some changes), so one file is probably not a big deal to copy when it changes.

DirectXMan12:
Ok, I'll copy websock.js over.  Is this good to merge?

kanaka:
@DirectXMan12 yep, looks good.

PR #451, opened 2015-02-18
Introduce Local Autoscaling

DirectXMan12:
This commit adds support for local autoscaling (both upscaling and downscaling, with the option to disable upscaling), based on #117.

Closes #117.

Credit goes to @mightypenguin for his initial work in #117 -- this is just an updated version of that with UI improvements.

### DIFF
--- a/include/display.js
+++ b/include/display.js
@@ -518,38 +518,48 @@ var Display;
             return this._fb_height;
         },
 
+        autoscale: function (containerWidth, containerHeight, downscaleOnly) {
+            var targetAspectRatio = containerWidth / containerHeight;
+            var fbAspectRatio = this._fb_width / this._fb_height;
+
+            var scaleRatio;
+            if (fbAspectRatio >= targetAspectRatio) {
+                scaleRatio = containerWidth / this._fb_width;
+            } else {
+                scaleRatio = containerHeight / this._fb_height;
+            }
+
+            var targetW, targetH;
+            if (scaleRatio > 1.0 && downscaleOnly) {
+                targetW = this._fb_width;
+                targetH = this._fb_height;
+                scaleRatio = 1.0;
+            } else if (fbAspectRatio >= targetAspectRatio) {
+                targetW = containerWidth;
+                targetH = Math.round(containerWidth / fbAspectRatio);
+            } else {
+                targetW = Math.round(containerHeight * fbAspectRatio);
+                targetH = containerHeight;
+            }
+
+            // NB(directxman12): If you set the width directly, or set the
+            //                   style width to a number, the canvas is cleared.
+            //                   However, if you set the style width to a string
+            //                   ('NNNpx'), the canvas is scaled without clearing.
+            this._target.style.width = targetW + 'px';
+            this._target.style.height = targetH + 'px';
+
+            this._scale = scaleRatio;
+
+            return scaleRatio;  // so that the mouse, etc scale can be set
+        },
+
         // Private Methods
         _rescale: function (factor) {
-            var canvas = this._target;
-            var properties = ['transform', 'WebkitTransform', 'MozTransform'];
-            var transform_prop;
-            while ((transform_prop = properties.shift())) {
-                if (typeof canvas.style[transform_prop] !== 'undefined') {
-                    break;
-                }
-            }
-
-            if (transform_prop === null) {
-                Util.Debug("No scaling support");
-                return;
-            }
-
-            if (typeof(factor) === "undefined") {
-                factor = this._scale;
-            } else if (factor > 1.0) {
-                factor = 1.0;
-            } else if (factor < 0.1) {
-                factor = 0.1;
-            }
-
-            if (this._scale === factor) {
-                return;
-            }
-
             this._scale = factor;
-            var x = canvas.width - (canvas.width * factor);
-            var y = canvas.height - (canvas.height * factor);
-            canvas.style[transform_prop] = 'scale(' + this._scale + ') translate(-' + x + 'px, -' + y + 'px)';
+
+            this._target.style.width = Math.round(factor * this._fb_width) + 'px';
+            this._target.style.height = Math.round(factor * this._fb_height) + 'px';
         },
 
         _setFillColor: function (color) {
--- a/tests/test.display.js
+++ b/tests/test.display.js
@@ -154,6 +154,84 @@ describe('Display/Canvas Helper', function () {
         });
     });
 
+    describe('rescaling', function () {
+        var display;
+        var canvas;
+
+        beforeEach(function () {
+            display = new Display({ target: document.createElement('canvas'), prefer_js: false, viewport: true });
+            display.resize(4, 3);
+            canvas = display.get_target();
+            document.body.appendChild(canvas);
+        });
+
+        afterEach(function () {
+            document.body.removeChild(canvas);
+        });
+
+        it('should not change the bitmap size of the canvas', function () {
+            display.set_scale(0.5);
+            expect(canvas.width).to.equal(4);
+            expect(canvas.height).to.equal(3);
+        });
+
+        it('should change the effective rendered size of the canvas', function () {
+            display.set_scale(0.5);
+            expect(canvas.clientWidth).to.equal(2);
+            expect(canvas.clientHeight).to.equal(2);
+        });
+    });
+
+    describe('autoscaling', function () {
+        var display;
+        var canvas;
+
+        beforeEach(function () {
+            display = new Display({ target: document.createElement('canvas'), prefer_js: false, viewport: true });
+            display.resize(4, 3);
+            canvas = display.get_target();
+            document.body.appendChild(canvas);
+        });
+
+        afterEach(function () {
+            document.body.removeChild(canvas);
+        });
+
+        it('should preserve aspect ratio while autoscaling', function () {
+            display.autoscale(16, 9);
+            expect(canvas.clientWidth / canvas.clientHeight).to.equal(4 / 3);
+        });
+
+        it('should use width to determine scale when the current aspect ratio is wider than the target', function () {
+            expect(display.autoscale(9, 16)).to.equal(9 / 4);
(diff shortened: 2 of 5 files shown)


samhed:
What is the difference between using these canvas.style.width/height and the approach seen in the display rescale function?

    canvas.style[transform_prop] =
                    'scale(' + this._scale + ') translate(-' + x + 'px, -' + y + 'px)';

kanaka:
@samhed using style affects the flow of the rest of the document, whereas transform does not. In other words, if you transform scale up an image, it starts to overlap the rest of the page instead of pushing other DOM elements out of the way like using style to change the effective size.

samhed:
@kanaka thanks for the clarification!

Is it really desirable to have both of these though, one in ui.js and one in display.js? Should we not simply replace the rescale transform-function in the display module with this method?

kanaka:
@DirectXMan12 yay! I've been wanting to do that for a while (years in fact). I'm not going to be able to test it anytime soon, but this lines up with the direction I was thinking and the code looks good from brief review. As long as it works well across all our supported browsers, I good with it.

Out of curiosity, how do you test visual correctness like this across multiple browsers? If you have a good system, it might be good worth making a wiki page with the info. My old method was to keep around a bunch of browser versions on Linux, plus latest browsers on a Windows 7 laptop. That was non-comprehensive and non-ideal to say the least.

If you don't have something awesome already, maybe we could provide links to VMs and/or containers to download. For example, Microsoft provides free VirtualBox VMs with most IE version/Windows versions at https://www.modern.ie/en-us/virtualization-tools#downloads. Another option too would be to create some captured streams of VNC data for playback, fix the playback page, and then we could use the free screenshot functionality at https://www.modern.ie/en-us/screenshots to run the playback (settings can be in the URL, so those can be tested too) across all the browsers that we support. https://saucelabs.com also gives 2 hours of automated testing free per month. Anyways, just some random thoughts I had just now. Interested to hear your thoughts on a good system for testing visual things that are hard to automate from the command line.

kanaka:
@DirectXMan12 yes, it might make sense to have it in display.js. Or maybe some of it anyways. I've tried to keep input.js, display.js, rfb.js, and websock.js fairly uncoupled from each other. I think input.js is the only module that's ever been used outside of noVNC (in gtk broadway), but by designing them so that each module _could_ be used separately I think it helps a lot to maintain a modular design.

As long as all $D references and UI.rfb accesses are replaced with function parameters e.g. display should only be aware of the canvas, but not any other DOM elements, UI settings, browser window details, or VNC details. Right now it's sort of a bunch of glue code intermingled with the resize logic, so it might take some refactoring, but the general functionality that this represents does make sense in display.js.

DirectXMan12:
> Is it really desirable to have both of these though, one in ui.js and one in display.js? Should we not simply replace the rescale transform-function in the display module with this method?

@samhed, @kanaka : I can take a look at that.  Would anyone be thrown off if we did that (is anyone who embeds noVNC in their project using the existing scale functionality currently?)

>  https://saucelabs.com also gives 2 hours of automated testing free per month. Anyways, just some random thoughts I had just now. Interested to hear your thoughts on a good system for testing visual things that are hard to automate from the command line.

@kanaka: we actually have an OpenSauce account.  It's how we run our unit tests across all the different OS/browser combos.

P.S. most of the credit for this should goes @mightypenguin -- the core functionality is based on his work on #117.

kanaka:
@DirectXMan12 actually, I think @mightypenguin is actually the most likely person to be affected by this since he is probably the one that has done the most tight integration/tweaking of the noVNC into an existing UI (at least that I'm aware of). The scale functionality has been broken since forever, so I suspect we won't make anybody particularly unhappy if we replace/fix it.

mightypenguin:
I'm fine with pulling out my auto scaling work. As long as the alternative can handle any arbitrary resolution (regardless of ratio) and "auto" scale based on browser window size then my code has no advantage. Or if those things don't matter to anyone.

I no longer maintain a software project that uses noVnc.

samhed:
I found a problem with this patch when choosing remote resize:

Either, if the server doesn't accept the resize request or if you make a server-side resize (for example with xrandr) you will not get clipping on touch devices.

I don't think clipping should be an option on touch devices. Currently we always have clipping enabled on touch devices since we do not want the browser to handle the panning. I think we need to keep it this way!

DirectXMan12:
> Either, if the server doesn't accept the resize request or if you make a server-side resize (for example with xrandr) you will not get clipping on touch devices.

@samhed : However, it seems silly to show the clipping button if we have local scaling, or if we successfully scale remotely.  How about a `onServerResize` callback that disabled clipping when it was not needed, and re-enabled it when it was needed on touch devices?

samhed:
> However, it seems silly to show the clipping button if we have local scaling, or if we successfully scale remotely.

@DirectXMan12 The clipping button is only shown when it is needed as of f8b399d, i.e. when the framebuffer is bigger than the viewport

DirectXMan12:
@samhed: whoops, forgot about that. 

I'll just move clipping back out into its own checkbox, then.

DirectXMan12:
Ok, so, in this patch I restore the 'clip' checkbox.  However, it is unchecked whenever local scaling or local downscaling is selected (the actual change is done when the settings are applied, either by closing the settings modal, or by clicking apply). 

This is because when clipping is applied, the style `position: fixed` is used.  This causes the actual canvas to be resized with the window, which causes 2 problems:

1. The content is already drawn with the correct aspect ratio to match our desired screen size.  Attempting to correct the aspect ratio by comparing it to the RFB aspect ratio results in distorted content.
2. Even if we don't attempt to correct for the aspect ratio, only the viewport is drawn on the canvas, meaning that when we scale, we just get a smaller version what we saw, not the whole framebuffer.

I'm thinking I'll whip up some javascript that immediately disables 'clipping' (and disables the checkbox) as soon as local scaling or local downscaling is selected.

DirectXMan12:
P.S. In the next version of the patch, whichever way we decide, I'll move the code into `_rescale` in `include/display.js`

samhed:
Nice!

>This is because when clipping is applied, the style position: fixed is used. This causes the actual canvas to be resized with the window, which causes 2 problems:

The canvas is actually positioned 'absolute' when clipping but I assume you get the same issue? Why does this cause the canvas to be resized?

> I'm thinking I'll whip up some javascript that immediately disables 'clipping' (and disables the checkbox) as soon as local scaling or local downscaling is selected.

I think that would be nice, to be as clear as possible to the user.

DirectXMan12:
> The canvas is actually positioned 'absolute' when clipping but I assume you get the same issue?

Whoops, mistype.  In fact, I think I may have been really tired when I wrote that sentence part.  In short, it doesn't.  In long, it doesn't ;-)  I just misread a couple of spots in the code.  In fact, what appears to be happening is that my copy of TigerVNC was continuously sending ExtDesktopSize, which was in turn calling `Display.resize`, which was in turn calling `viewportChangeSize`, which was in turn scaling the viewport to the window size when enabled.  I apologize for any confusion that may have occurred.

Anyway, the rest still stands -- unless the underlying clipping javascript is disabled, the whole `resize canvas to window` logic occurs (actually in), and we get the same two issues.

DirectXMan12:
@samhed : does this look good to you?

samhed:
Looks great! But:

> P.S. In the next version of the patch, whichever way we decide, I'll move the code into _rescale in include/display.js

DirectXMan12:
Ok, I've moved the code into `display.js`.  I've introduced a new function named `autoscale`, which takes a container width, container height, and a boolean indicated whether to only downscale.

tests/test.display.js
```diff
@@ -154,6 +154,84 @@ describe('Display/Canvas Helper', function () {
...
+        });
+
+        it('should not upscale when downscaleOnly is true', function () {
+            expect(display.autoscale(2, 2, true)).to.equal(0.5);
+            expect(canvas.clientWidth).to.equal(2);
+            expect(canvas.clientHeight).to.equal(2);
+
+            expect(display.autoscale(16, 9, true)).to.equal(1.0);
+            expect(canvas.clientWidth).to.equal(4);
+            expect(canvas.clientHeight).to.equal(3);
```
samhed: How do these tests pass? I would expect the client size to be 2x2 after the first scale?

include/display.js
```diff
@@ -520,36 +520,46 @@ var Display;
...
-            var y = canvas.height - (canvas.height * factor);
-            canvas.style[transform_prop] = 'scale(' + this._scale + ') translate(-' + x + 'px, -' + y + 'px)';
+            // NB(directxman12): If you set the width directly, or set the
+            //                   style width to a number, the canvas is cleared.
+            //                   However, if you set the style width to a string
+            //                   ('NNNpx'), the canvas is scaled without clearing.
+            this._target.style.width = targetW + 'px';
+            this._target.style.height = targetH + 'px';
+
+            this._scale = scaleRatio;
```
samhed: I guess we could call "this._rescale" here to share some code.

samhed:
Looks nice, but we need to move autoscale to be among the other public methods, currently it is where the private ones are.

tests/test.display.js
DirectXMan12: I'm unclear as to what you mean.  Scale factor is calculated based on the "logical size" -- the RFB size.  The rendered with and height is 2x2 after the first call, which is a scale factor of 0.5.  We then try to scale to 16x9, which would end up being a scale factor of 3.0 (9 / 3).  However, this would upscaling, so we clip to a scale of 1.0, meaning the rendered width and height is 4x3 (the RFB dimensions).

DirectXMan12:
> Looks nice, but we need to move autoscale to be among the other public methods, currently it is where the private ones are.

Good catch.  Will do.

tests/test.display.js
samhed: Thank you for clarifying. I was just confused

samhed:
As far as I am concerned, this is ready for merge after that!

Issue #453, opened 2015-02-18
Remove the 'debian' directory


DirectXMan12:
@kanaka: do you know if anyone is actively using this directory?  I suspect it's not actually all that useful or necessary (both Debian and Ubuntu seem to have their own version of the packaging).

kanaka:
@DirectXMan12 I think it has probably served its purpose (to serve as template for packagers). I think @abligh was the main person who proposed fixes to that directory. @abligh are you still using the novnc/debian directory for anything. Forsee any problems if we drop it?

abligh:
I suspect it may have bitrotted. I have a fixed up version somewhere that I use which you are welcome to, but I think most people would be better served by using the Debian or Ubuntu one (as I haven't tested it in every novnc situation). I'd be happy to see it go.

PR #454, opened 2015-02-18
Remove non-existent files from debian install file

drwahl:
The recent cleanup work has left some references to files that no longer
exist in the debian novnc.install file.  This patch removes those now
dead references.

### DIFF
--- a/debian/novnc.install
+++ b/debian/novnc.install
@@ -2,12 +2,9 @@ vnc.html	/usr/share/novnc
 vnc_auto.html	/usr/share/novnc
 README.md  /usr/share/doc/novnc
 LICENSE.txt  /usr/share/doc/novnc
-utils/Makefile   /usr/share/novnc/utils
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Actually, we're thinking about removing the `debian` directory entirely ( #453 ).  Please let us know if this would impact you.

drwahl:
We're currently using the novnc.install file to build debs with our build system.  If this were to go away, it would be inconvenient, but not a show stopper.  I guess my personal preference is to keep it, but I'm sure there are things to consider beyond my scope.

DirectXMan12:
@drwhahl : could you use the deb source files used by Debian itself -- they're probably more "correct".

drwahl:
Certainly could.  That shouldn't be an issue.  Closing this issue and accepting the above comment as the way forward.  Thanks @DirectXMan12.

PR #458, opened 2015-02-20
Fix disconnect/reconnect issues

### LINKED ISSUE #452: Disconnect/Reconnect Issues
Commit 155d78b39933a396d2555376f4e70f976c30b86d unregisters certain event listeners (namely, 'open') that are only registered in the RFB constructor.  Since the RFB object is reused between disconnects and reconnects, meaning that the reconnect operation doesn't continue past the websocket opening.

If we create a new RFB object each time, in certain cases, the keyboard (and mouse) do not seem to be ungrabbed.

Some profiling in chrome seems to indicate that there are a couple other event handler leaks as well (I've been unable to track down which handlers these are, so they may be from something life setInterval).

DirectXMan12:
Commit 155d78b39933a396d2555376f4e70f976c30b86d prevented reconnections
from working properly.  This fixes that by creating a new RFB object
after disconnecting or failing.

Furthermore, this ensures that a new connection cannot be opened util
we've actually disconnected (either by timer or by receiving a `close`
event).

Closes #452

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -482,6 +482,7 @@ var RFB;
                 Util.Debug("Clearing disconnect timer");
                 clearTimeout(this._disconnTimer);
                 this._disconnTimer = null;
+                this._sock.off('close');  // make sure we don't get a double event
             }
 
             switch (state) {
--- a/include/ui.js
+++ b/include/ui.js
@@ -110,13 +110,7 @@ var UI;
             UI.initSetting('path', 'websockify');
             UI.initSetting('repeaterID', '');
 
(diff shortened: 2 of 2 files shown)


samhed:
Sweet! I can test this on Monday

samhed:
Works nicely!

But how come we have to make a new rfb object in connect as well? I mean if we do it at start and at disconnect, shouldn't we be covered?

If we do in fact need this in connect, we should probably use UI.initRFB() to avoid duplicated code?

DirectXMan12:
I think that was left in from when I was testing a different approach.  Thanks for catching that (not that it really affects anything to leave it in, but there's no need to have it).  Fixed!

PR #461, opened 2015-02-26
Make Util.getPosition be relative to page

### LINKED ISSUE #459: messed up cursor coordinates when scrolled
as of 2015-02-25, the git version behaves incorrectly:

when connected to a wide screen setup (two monitors), the scrollbar at the bottom appears. this is correct.

unfortunately, when it's used. the coordinates of the cursor on the target system do not match those displayed on the client. it seems to be something like 2x the offset.

reverting to "stable" (0.5.1) solved the issue.

just a quick browse through the commits brought up 5108c4635c847de9be0edadf572f7426f351b66a as a suspect but that's just guessing.

DirectXMan12:
Commit 5108c4635c847de9be0edadf572f7426f351b66a caused a regression
in the case where scrolling is used -- getPosition return position
relative to the viewport, while getEventPosition expected a position
relative to the page.

As per
https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect,
the fix for this is simply to add the `pageXOffset` and `pageYOffset` to
the output of `getBoundingClientRect()`.

Fixes #459.

### DIFF
--- a/include/util.js
+++ b/include/util.js
@@ -436,7 +436,7 @@ Util.load_scripts = function (files) {
 Util.getPosition = function(obj) {
     "use strict";
     var objPosition = obj.getBoundingClientRect();
-    return {'x': objPosition.left, 'y': objPosition.top};
+    return {'x': objPosition.left + window.pageXOffset, 'y': objPosition.top + window.pageYOffset};
 };
 
 


DirectXMan12:
I'll merge this once Travis passes, then

samhed:
Hm, how come one of the Safari tests failed? 

>ERROR [launcher.sauce]: Can not start safari 7 (OS X 10.9)

I guess it doesn't have anything to do with that we should use scrollX and scrollY instead? According to https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX pageXOffset is simply an alias for scrollX.

DirectXMan12:
@samhed : I think it's an issue with SauceLabs.  I'm going to investigate -- I've been seeing intermitent failures to actually start the tests on Sauce over the paste few days, but haven't had time to track them down.

Issue #446, opened 2015-02-11
Please, no getElementById within the code


ibc:
https://github.com/kanaka/noVNC/blob/master/include/display.js#L229

```
var cb = document.getElementById('noVNC-control-bar');
```

No please, really not. Don't force me to have a specific HTML element with "id=noVNC-control-bar". Please not. Yo have added that because ui.js and base.css use such an HTML element, but that is just the demo application, not the library. Please don't mix them.

Consider the case in which I want to have two isolated and 100% separated noVNC displays in the same web app.

Please, don't do that.

ibc:
Note that `RFB` requires `Display`, so at the end `Display` is part of the core "library". If it needs to have a reference to an HTML element then let it to be passed via API/configuration.

samhed:
Thanks for pointing it out. Apologies for being careless!

I'm in the process of changing a few more things in display.js, I will submit a new PR where I will include a fix for this issue.

ibc:
So thanks!

ibc:
Hi, any update about this? :)

samhed:
Working on it

ibc:
Thanks a lot!

ibc:
:+1:

Issue #463, opened 2015-03-02
Firefox on Android does not resize properly


samhed:
Steps to reproduce:

1. Open noVNC in Firefox on Android
2. Log in using the "Resize remote to window" setting using portrait mode
3. After the session is done loading in, rotate the device to landscape

Expected behavior:

* The session should be resized to fit the screen.

Actual behavior:

* The session becomes roughly 200 px to high after the resize.

samhed:
This is due to the fact that the value of window.innerHeight is wrong just after a rotation. See these bugs mozilla's bugzilla:

https://bugzilla.mozilla.org/show_bug.cgi?id=923663
https://bugzilla.mozilla.org/show_bug.cgi?id=1071620
https://bugzilla.mozilla.org/show_bug.cgi?id=1125566

A fix would be to not use window.innerHeight

samhed:
This issue is unfortunately not fixed despite my efforts.. As I noted here https://bugzilla.mozilla.org/show_bug.cgi?id=1125566 it is not limited to window.innerHeight.

But since there is nothing we can do but to wait for mozilla to fix it in Firefox, I don't see any point in reopening.

samhed:
I have verified that Mozilla has now fixed this in their nightly build of Firefox on Android.

samhed:
The fix for this is now in stable version of Firefox on Android :tada:

PR #464, opened 2015-03-02
Fix resize in Firefox on Android

samhed:
This fixes issues #446 and #463 by using a different approach of getting the full size of the canvas. Instead of using window.innerWidth and window.innerHeight this would resize the canvas to fit the noVNC_container which covers the entire page.

Edit: This also fixes issue #465

### DIFF
--- a/include/ui.js
+++ b/include/ui.js
@@ -1,7 +1,7 @@
 /*
  * noVNC: HTML5 VNC client
  * Copyright (C) 2012 Joel Martin
- * Copyright (C) 2013 Samuel Mannehed for Cendio AB
+ * Copyright (C) 2015 Samuel Mannehed for Cendio AB
  * Licensed under MPL 2.0 (see LICENSE.txt)
  *
  * See README.md for usage and integration instructions.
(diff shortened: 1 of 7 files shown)


samhed:
This also fixes issue #465

DirectXMan12:
In general, looks good.  I'll try and test it soon.

samhed:
> Once #451 merges, you can just use getPosition here -- it will returns the width and height (or you could just use getBoundingClientRect directly).

You mean instead of parseInt from window.getComputedStyle? Like this

    var h = Util.getPosition(container).height

?

DirectXMan12:
> You mean instead of parseInt from window.getComputedStyle? Like this

>     var h = Util.getPosition(container).height

> ?

Yep!

samhed:
This PR is now updated to reflect the changes @DirectXMan12  made in #451

samhed:
I'm getting a bit limited in my time to work on this soon, basically I only have today and tomorrow available. Would it be possible for you @DirectXMan12 to verify that I haven't broken the scaling etc. and come with potential comments today or tomorrow?

I have tested scaling and as far as I can tell I haven't broken anything.

include/ui.js
```diff
@@ -258,6 +218,55 @@ var UI;
...
+        },
+
+        getCanvasLimit: function () {
+            var container = $D('noVNC_container');
+
+            // Hide the scrollbars until the size is calculated
+            container.style.overflow = "hidden";
+
+            var w = Util.getPosition(container).width;
+            var h = Util.getPosition(container).height;
```
DirectXMan12: This could be more effecient (`var pos = Util.getPosition(container);`, etc), but it's not a big deal.

DirectXMan12:
LGTM! :+1:  Seems to work for me as well.

samhed:
Thank you for taking the time @DirectXMan12 

> This could be more effecient (var pos = Util.getPosition(container);, etc), but it's not a big deal.

Good catch, I fixed it in 16b3ef7

PR #473, opened 2015-03-24
Create RFB object on connect

### LINKED ISSUE #467: Authentication errors are not displayed
Since e543525 which was merged as of #458 we create a new RFB object after authentication failiure and do thus loose the rfb_state. The error just flashes by in the status bar and you can see it in the javascript console.

DirectXMan12:
In e543525faa9cf0d683f41e183e89cd909f3dd229, we switched to creating
a new RFB object on disconnect.  This caused issues, however, since
any errors were only displayed briefly before the new "loaded" text
was displayed instead.

Now, we create the RFB object on connect.  This essentially removes
the usefulness of the "loaded" state, but prevents the aforementioned
problem.

To facilitate this, the code which does detection of cursor URI support
was moved from this Display constructor (which now calls the new
function) into its own function, `Util.browserSupportsCursorURIs()`.

Fixes #467

### DIFF
--- a/include/display.js
+++ b/include/display.js
@@ -86,29 +86,9 @@ var Display;
         }
 
         // Determine browser support for setting the cursor via data URI scheme
-        var curDat = [];
-        for (var i = 0; i < 8 * 8 * 4; i++) {
-            curDat.push(255);
-        }
-        try {
-            var curSave = this._target.style.cursor;
-            Display.changeCursor(this._target, curDat, curDat, 2, 2, 8, 8);
-            if (this._target.style.cursor) {
-                if (this._cursor_uri === null || this._cursor_uri === undefined) {
-                    this._cursor_uri = true;
-                }
-                Util.Info("Data URI scheme cursor supported");
-                this._target.style.cursor = curSave;
-            } else {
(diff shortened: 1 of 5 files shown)


samhed:
Sweet! Looks like a clean solution :+1:

On #467 you wrote:

> and the status bar defaults to having a class of noVNC_status_normal with some initial boilerplate text

With these changes, the statusbar is empty initially if I'm not mistaken. Maybe it should be empty? I will test this tomorrow and decide what I think.

DirectXMan12:
> With these changes, the statusbar is empty initially if I'm not mistaken

Correct.  I couldn't quite decide what to put in the status bar, so I left it empty ;-)

samhed:
> so I left it empty ;-)

I think it looks good like that!

However, explicitly specifying the class of the control bar in the HTML conflict with the way we set the style in updateState in ui.js. With this patch the class is never changed, i.e we don't get the warning/error gradient colors.

DirectXMan12:
@samhed : I just set the class on the wrong div ;-)  It should work fine now.

samhed:
God I must've been tired when writing that! It does indeed work fine now.

Let's close this in favor for #474 which includes these changes then.

PR #474, opened 2015-03-24
Throw exceptions from RFB constructor

### LINKED ISSUE #460: If no target if given to RFB all get broken
Just try to not provide the `RFB` without a proper "canvas" target. It will fail everywhere:

* It will try to call `this._sock.close()` but the WS socket does not exist due the `Display` error.
* If a try/catch is set, then it will fail due to "Uncaught TypeError: Cannot read property 'get_render_mode' of null" which happens at the end of the RFB constructor:

```
  var rmode = this._display.get_render_mode();

  this._updateState('loaded', 'noVNC ready: ' + rmode);
}
```

This is, even if "state:fatal" was already emitted, the RFB constructor continues.

This must be really more robust.

DirectXMan12:
Throw error from constructor on Display error
    
Previously, if an error was thrown from the Display constructor in the RFB constructor, we would attempt to use `RFB#updateState`` to handle this.  However, `RFB#updateState` attempts to close the WebSocket connection, which doesn't exist at this point.
    
In the constructor, it's probably just better to raise an exception instead (making sure to clean up anything relevant), both for errors from the display constructor, and for errors due to WebSocket emulation.
    
Fixes #460

### DIFF
--- a/include/display.js
+++ b/include/display.js
@@ -86,29 +86,9 @@ var Display;
         }
 
         // Determine browser support for setting the cursor via data URI scheme
-        var curDat = [];
-        for (var i = 0; i < 8 * 8 * 4; i++) {
-            curDat.push(255);
-        }
-        try {
-            var curSave = this._target.style.cursor;
-            Display.changeCursor(this._target, curDat, curDat, 2, 2, 8, 8);
-            if (this._target.style.cursor) {
-                if (this._cursor_uri === null || this._cursor_uri === undefined) {
-                    this._cursor_uri = true;
-                }
-                Util.Info("Data URI scheme cursor supported");
-                this._target.style.cursor = curSave;
-            } else {
-                if (this._cursor_uri === null || this._cursor_uri === undefined) {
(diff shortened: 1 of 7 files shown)


DirectXMan12:
NB: this includes the changes from #473 in order not to break certain things.

samhed:
This also looks good aside from the small mistake at line 419 in rfb.js.

samhed:
I can't login when trying this:

    util.js:220 Display exception: Error: Target must be set
      Util.Error	@	util.js:220
      RFB	@	rfb.js:167
      UI.initRFB	@	ui.js:164
      UI.connect	@	ui.js:782

DirectXMan12:
Did you try with the new version of the commit (0658ba7)?  The previous version had a bug due to some code I was using to test.

samhed:
That was indeed the issue, apologies. It seems to work fine now!

Let's merge this after the control-bar error-colors get fixed (as noted in #473). :+1:

PR #475, opened 2015-03-26
Follow symbolic links in launch.sh

DirectXMan12:
Previously, in launch.sh, `$HERE` was the directory of `$0`.
However, if `$0` was actually a symlink, `$HERE` would be
wherever the symlink was, which could cause issues (for
example, the script wouldn't be able to local `$WEB` or
`$WEBSOCKIFY` properly).

Now, `$HERE` looks at whatever `$0` points at instead.

Closes #447.

### DIFF
--- a/utils/launch.sh
+++ b/utils/launch.sh
@@ -22,7 +22,8 @@ usage() {
 }
 
 NAME="$(basename $0)"
-HERE="$(cd "$(dirname "$0")" && pwd)"
+REAL_NAME="$(readlink -f $0)"
+HERE="$(cd "$(dirname "$REAL_NAME")" && pwd)"
 PORT="6080"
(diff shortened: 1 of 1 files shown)


DirectXMan12:
@kanaka : what do you think about this?  While there's technically the possibility to break some people (if they had symlinked to `launch.sh` in a noVNC installation, but had the symlink in a different noVNC installation), I think it's far more likely that this is the behavior that we want (at the very least, we have one person asking for it in #447).

kanaka:
@DirectXMan12 I think that's the right approach. BTW, launch.sh was never really intended to be a primary interface. My intention was just for it to be a really easy way for people to clone and test without any fuss. Just so you know the history.

PR #483, opened 2015-05-03
Make noVNC compatible with Content Security Policies

fabian-z:
This pull request simply moves the style and onsubmit declarations from #keyboardinput in vnc.html to base.css and ui.js, making it possible to use noVNC on mobile devices with an active Content Security Policy after these small changes.

An example policy header working with noVNC (note ws://127.0.0.1:8080 needs to be replaced with the scheme, host & port noVNC should connect to):

```
default-src 'none'; font-src 'self'; script-src 'self'; img-src 'self' data:; style-src 'self'; connect-src 'self' ws://127.0.0.1:8080;
```

I admit this is not terribly useful if noVNC should connect to different clients, but for integration into specific applications this may improve security, and the needed changes are not too big.

### DIFF
--- a/include/base.css
+++ b/include/base.css
@@ -238,6 +238,7 @@ html {
   position: relative;
   left: -40px;
   z-index: -1;
+  ime-mode: disabled;
 }
 
 /*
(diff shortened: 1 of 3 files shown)


samhed:
There is a function in ui.js called "addMouseHandlers" which should be used for the onclick. Could you update your PR?

fabian-z:
Just fixed my PR, thanks for the suggestion of addMouseHandlers.
I rebased the fix into my original commit to keep the history clean.

Issue #431, opened 2015-01-21
noVNC eats CPU and hangs Firefox


abligh:
noVNC launched connecting to a qemu instance running Firefox causes extreme CPU usage and essentially hangs firefox. Essentially a lot of RFB output appears to hang firefox.

Here's the easiest way to replicate:
* Launch a cirros VM (I don't think the OS matters but that's the easiest to download) using qemu, specifying a VNC port to bind it to (therefore qemu is the rfb server)
* Use websockify to forward to this port (no SSL needed)
* Open the VNC console in Firefox
* Log in to the Cirros VM and type `ls -lR /` or indeed anything else that produces a lot of screen output
* Watch firefox die

The 'death' involves one page of output, a multisecond delay, then a second page of output, then a complete hang with the CPU pegged. Firefox is then unresponsive until the window containing VNC is closed (by hitting the close button on the window), at which point after a few seconds it recovers.

I am using the following:
* noVNC 0.5.1 straight from git on Linux. 0.5.0 did the same thing.
* Firefox 35.0 (current) on Mac OS-X. Firefox 31.0 does the same thing
* websockify 0.5.1+dfsg1-3ubuntu0.14.04.1 (Ubuntu Trusty current). My own websockets proxy which shares no code does the same thing.
* qemu 2.0.0+dfsg-2ubuntu1.10 (Ubuntu Trusty current). Guest OS tried: Cirros, Ubuntu.
* no SSL (have tried SSL using my own websockets proxy and it is also broken)

There is no problem using Chrome.

I conclude the problem is thus somewhere between Firefox and novnc, possibly when qemu is the server, when there's a lot of screen output and scrolling.

samhed:
I am not able to reproduce using TigerVnc as the vnc server. Tested Firefox 33 on Mac OS-X and Firefox 34 on Fedora 20.

abligh:
Definitely works with qemu as the vnc server :-)

samhed:
So the problem is fixed now or what do you mean?

abligh:
Sorry, I mean it can definitely be replicated with qemu as the vnc server - fixed my ambiguous comment

DirectXMan12:
I'm able to reproduce the cases of both @samhed and @abligh.  There appears to be something with the way qemu is sending the update frames.  I'll investigate further.

DirectXMan12:
[...]

PR #487, opened 2015-05-18
Introduce new zlib implementation

DirectXMan12:
***WIP: Do Not Merge***

This is an initial shot at replacing our current zlib implementation (jsunzip) with a faster one.

I ran two different tests: one was running the `sl` program using a TigerVNC server, and the other was running `ls` against a qemu-hosted VM using qemu's built-in VNC support.  Numbers reported are based on the `vnc_playback.html` perf test.

For the `sl` test, the new implementation shows a slight improvement (maybe 10%).  The test was preformed based on a recording of the `sl` program.

For the `ls` test, the new implementation cut the time nearly in half in Chrome, and even more in Firefox.  The test was performed by running `ls -lR` on my home directory in the qemu-hosted VM.

Thoughts?

### DIFF
--- a/include/inflator.js
+++ b/include/inflator.js
@@ -1,0 +1,2409 @@
+(function(f){if(typeof exports==="object"&&typeof module!=="undefined"){module.exports=f()}else if(typeof define==="function"&&define.amd){define([],f)}else{var g;if(typeof window!=="undefined"){g=window}else if(typeof global!=="undefined"){g=global}else if(typeof self!=="undefined"){g=self}else{g=this}g.inflator = f()}})(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
+'use strict';
+
+
+var TYPED_OK =  (typeof Uint8Array !== 'undefined') &&
+                (typeof Uint16Array !== 'undefined') &&
+                (typeof Int32Array !== 'undefined');
+
+
+exports.assign = function (obj /*from1, from2, from3, ...*/) {
+  var sources = Array.prototype.slice.call(arguments, 1);
+  while (sources.length) {
+    var source = sources.shift();
+    if (!source) { continue; }
+
+    if (typeof source !== 'object') {
+      throw new TypeError(source + 'must be non-object');
+    }
+
+    for (var p in source) {
+      if (source.hasOwnProperty(p)) {
+        obj[p] = source[p];
+      }
+    }
+  }
+
+  return obj;
+};
+
+
+// reduce buffer size, avoiding mem copy
+exports.shrinkBuf = function (buf, size) {
+  if (buf.length === size) { return buf; }
+  if (buf.subarray) { return buf.subarray(0, size); }
+  buf.length = size;
+  return buf;
+};
+
+
+var fnTyped = {
+  arraySet: function (dest, src, src_offs, len, dest_offs) {
+    if (src.subarray && dest.subarray) {
+      dest.set(src.subarray(src_offs, src_offs+len), dest_offs);
+      return;
+    }
+    // Fallback to ordinary array
+    for (var i=0; i<len; i++) {
+      dest[dest_offs + i] = src[src_offs + i];
+    }
+  },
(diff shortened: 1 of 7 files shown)


DirectXMan12:
Currently the only issue is that the output buffer is of fixed size -- it's currently set at 100k (it's only allocated once per instance though, so increasing the size doesn't have too much of a cost).  So far, this hasn't been an issue, but it could be, hypothetically.  We could also scale it dynamically based on screen size (which should have little impact on performance) or use a loop and join buffers together (which could have a significant impact on performance).

DirectXMan12:
CC @kanaka , @samhed , @astrand

DirectXMan12:
See also #431

kanaka:
@DirectXMan12 Was getting the playback working again difficult? From the code changes it doesn't look like it was too bad (hopefully). I think we should get those changes in even if we don't add the pako change right away. BTW, in addition to the total time reported by playback, if you enable logging=info, you'll get some more timing information (per encoding type) in the javascript console. Might give a little more insight.

Anyways, since this change gives 10% boost overall even in the worst case this seems like a good candidate for merging. I think my only concern would be to make sure that we don't have a massive regression (broken or much slower) in some of the older browser versions that we still want to support.

kanaka:
@DirectXMan12 BTW, it might be interesting to compare with an asm.js enabled zlib implementation: https://github.com/ukyo/zlib-asm That does mean getting an emscripten toolchain working, but since we probably want to switch a lot of the noVNC fast path to asm.js at some point, this might be a good place to start experimenting. I might even have some time to work on it this summer if you don't, although no promises.

kanaka:
@DirectXMan12 and I forgot to say good work! It's exciting to see this starting to happen. There is a lot of potential areas for noVNC performance improvements: binary types, asm.js, web workers (especially now that firefox supports WebSockets in workers), SIMD.js, etc. And since some of those areas represent in-progress standards efforts, we also have an opportunity to contribute to state-of-the-art. noVNC is a fairly widely used project where those technologies can actually be used and tested using real-world conditions. So I think we are somewhat uniquely positioned to be able to contribute to the development of those technologies. The same was true with canvas and websockets back when noVNC started and I was able to make a few minor contributions to the Websocket protocol and Canvas API. If only we all had several more hours in the day :-) Maybe I can convince my faculty supervisor to let me work on this as part of my PhD.

DirectXMan12:
> @DirectXMan12 Was getting the playback working again difficult? From the code changes it doesn't look like it was too bad (hopefully). I think we should get those changes in even if we don't add the pako change right away.

I concur.  It was really just a case of adapting it to work since the refactor (duplicating "test mode" outside of the RFB object's API) and changing it to create a new `RFB` object each iteration.

> BTW, in addition to the total time reported by playback, if you enable logging=info, you'll get some more timing information (per encoding type) in the javascript console. Might give a little more insight.

I'll look into this.

> Anyways, since this change gives 10% boost overall even in the worst case this seems like a good candidate for merging. I think my only concern would be to make sure that we don't have a massive regression (broken or much slower) in some of the older browser versions that we still want to support.

Yeah.  I run on the bleeding edge for Firefox (nightly) and fairly recent for Chromium (just recompiled to 44), so I was hoping some other people could give some feedback about their experience in commonly used applications with other browsers.  I'll run some basic tests in SauceLabs as well.

@kanaka: what do you think about the buffer size issue?

> @DirectXMan12 BTW, it might be interesting to compare with an asm.js enabled zlib implementation: https://github.com/ukyo/zlib-asm That does mean getting an emscripten toolchain working, but since we probably want to switch a lot of the noVNC fast path to asm.js at some point, this might be a good place to start experimenting. I might even have some time to work on it this summer if you don't, although no promises. 

Yeah, I considered this.  The issues are that asm.js is generally less performant than well-written code on Chrome (http://blog.kazuhooku.com/2013/12/benchmarking-asmjs-using-real-world.html) at the moment, unfortunately.

The other issue (not with asm.js code, but with zlib-asm) is that zlib-asm (as well as other zlib libraries) don't have APIs designed for the use case used in RFB -- having a stream and getting partial results out of that stream.  Fortunately for us, pako presents a low-level API similar to the actual zlib API, so we can use that.

> There is a lot of potential areas for noVNC performance improvements: binary types, asm.js, web workers (especially now that firefox supports WebSockets in workers), SIMD.js, etc

One of these days I'm going to get around to the major refactor for ArrayBuffer usage.  One of the big issues is that allocating new ArrayBuffer seems to be more expensive than allocating Arrays, so you want to try not to make lots of new ArrayBuffer objects.  This would probably involve a lot more passing around indicies pointing to locations inside of ArrayBuffers instead of passing around new Arrays, like we do now.

DirectXMan12:
Oh, and FYI, there's a bug fix in there that I should really pull out before merge -- there's a bug-type-thing when noVNC is used with Firefox Nightly that somehow causes the encoding handler object to not have objects defined.  It acts like a race condition (it doesn't seem to appear when run with a debugger, it only appears during times of rapid throughput, etc), but I can't seem to track down what exactly is going on.  The "solution" is essentially to check if the encoding handler is null and then try again.  I hope to have a better solution at some point.

DirectXMan12:
superseded by #488

PR #488, opened 2015-05-20
Additional Performance Improvements

DirectXMan12:
~~**WIP: Do Not Merge**~~

This commit contains a variety of additional performance improvements, currently focused around the tight indexed-to-rgb encoding.  It includes the commits from #487 

Currently, I've reduced the time a further 50% from #487 on the qemu ls test.

This makes use of Typed Arrays.  The big tradeoff here is that we preallocate, meaning noVNC starts off using more RAM (we preallocate 4 MiB for the receive queue, `4 * width * height` B for a work buffer, and 1 KiB for a palette buffer).

CC @kanaka


kanaka:
@DirectXMan12 I never replied regarding the buffer preallocation. I think that's perfectly fine. In fact it's one of the optimizations I have been assuming that we would need to make eventually. I think just reducing GC pauses pretty much mandates pre-allocated buffers for this sort of application. You would have to run the real numbers, but I actually wouldn't be that surprised if the actual average peak memory consumed by noVNC doesn't go up all that much because we have a fair number of large objects that accumulate and have to be GC'd periodically.

kanaka:
@DirectXMan12 I haven't looked at the diffs in great detail, but it looks like there might be some code paths that currently bypass the renderQ. If so, that will result in visual corruption in some context because things will get drawn out of order. The reason that the renderQ exists at all is because when we load a jpeg image from base64 data (for tight encoding), it's a non-synchronous operation so we have to delay that canvas draw and all following drawing operations until the image is loaded. By using a pending queue we allow other rects to be processed and decoded while a previous tight jpeg image is loading. So by passing the renderQ will increase the efficiency of rendering, but visible corruption will happen when tight is used with any other encoding (e.g. especially copyrect which may try to copy a region of the screen that hasn't actually been rendered yet).

DirectXMan12:
@kanaka: yeah, I realized that I skipped the render queue in some paths, and was working on a good solution to fix that without sacrificing performance when the render queue is not in use.  Creating all those objects is *not* good for performance.  What I ended up doing (I just pushed it) was to have the people call the display methods directly, and have the display methods conditionally push to the render queue if the render queue is in use/needed.

DirectXMan12:
@kanaka , @samhed, @astrand : I'd like to get some eyes on the most recent changes I've made.  I've improved a couple of the main paths significantly, but other paths should benefit as well.  If you have any data recordings of a slow session that you can send my way, I can take a look at those as well.

kanaka:
I did a high level review of your changes. I like the change to move the all the render queue logic down into display is definitely more clean in terms of modularity (rfb shouldn't be aware of how display is choosing to render things for the most part).

I think we should have a header on inflator.js with the original site and license (and also list it to LICENSE.txt).

I'll try and do a bit more in depth review later tonigh.

DirectXMan12:
> I think we should have a header on inflator.js with the original site and license (and also list it to LICENSE.txt).

Yep, will do -- it's on my list.  I'll add notes on how I generated it as well.

> I'll try and do a bit more in depth review later tonigh.

Cool, thanks.

DirectXMan12:
Ok, I added the license for Pako to LICENSE.txt and docs/LICENSE.pako.

kanaka:
@DirectXMan12 did a bit deeper review:

The from_push variable is kind of throwing me off (because both code path could be "from push" depending on your perspective). Maybe from_queue or something like that?

I suspect colourMap modification might need to be ordered with the renderQ to be fully correct so that it doesn't impact rendering of prior queue operations. Also, I think 'blit'/blitImage and finishTile should have the same renderQ logic as all the other operations. This is not a regression, but in reviewing the code, I think it's a hole in the existing logic and the restructuring just made it more obvious. These problems may not actually happen in practice (i.e. there may not be any actually servers that mix hextile and tight) but it should probably cleaned up and made consistent and correct across the board.

The refactor of messages.* functions does feel a bit more messy compared to before, in particular that they mutate the send queue but rely on the caller to mutate the sQlen variable (if it's not just the length already). Maybe passing in the queue and length as a single object and having the callee modify the length (and getting rid of the separate _mouse_arr). Is there a measurable performance impact for avoiding object creation on the send side of things? I haven't thought in depth about a clean way to do this but it does feel like there is some room for improvement here.

I think that's all for now. Otherwise, I think the changes look good and as long as cross-browser testing seems to be showing correct rendering, I'm good with the changes.

samhed:
As @kanaka mentioned in #496 , we should probably make websock.js modifications in the websockify project and afterwards sync them into noVNC.

DirectXMan12:
> The from_push variable is kind of throwing me off (because both code path could be "from push" depending on your perspective). Maybe from_queue or something like that?

Sounds good.

> I suspect colourMap modification might need to be ordered with the renderQ to be fully correct so that it doesn't impact rendering of prior queue operations. Also, I think 'blit'/blitImage and finishTile should have the same renderQ logic as all the other operations. This is not a regression, but in reviewing the code, I think it's a hole in the existing logic and the restructuring just made it more obvious. These problems may not actually happen in practice (i.e. there may not be any actually servers that mix hextile and tight) but it should probably cleaned up and made consistent and correct across the board.

Yeah.  The only reason I didn't add them in is because they didn't have renderQ support previously.

> The refactor of messages.* functions does feel a bit more messy compared to before, in particular that they mutate the send queue but rely on the caller to mutate the sQlen variable (if it's not just the length already). Maybe passing in the queue and length as a single object and having the callee modify the length (and getting rid of the separate _mouse_arr).

So, I could test performance with a persistent object that is a property of `Websock` and never gets recreated (just modified), or we could pass the actual `Websock` object to the the `messages.*` functions.  The downside with the latter is that you couldn't have an intermediary array (like mouse_arr).  I'll check the performance impact.

P.S. What was the reasoning behind `mouse_arr`.  AFAICT, it's used for grouping mouse movement and mouse click events so that they all get sent at once.

> Is there a measurable performance impact for avoiding object creation on the send side of things? 

Definitely.  In the raw performance tests (`playback.html`) it wasn't quite as significant as the read performance, but it definitely reduces real-time usage performance (i.e. performance in live sessions).

kanaka:
I think we can get rid of separate mouse_arr. To be honest, I don't remember the exact reason for it, but, yes, batching of mouse movements was probably the reason for it. But that can be done without the split, so I think it's really just a legacy thing and we can refactor it to remove it. I think either option is reasonable once we get rid of mouse_arr.

DirectXMan12:
Alright, this update should address @kanaka's concerns (render queue support added, message functions take the websock object as a parameter, `mouse_arr` has been nixed) with the exception of render queue support for `finishTile`, since we'd need to have render queue support for all the tile functions, and I'm not sure what kind of performance impact that would have.  I can take a look, but it would be helpful if someone has a HEXTILE recording to test with.  Render queue support for the tile functions can be added in later pretty easily.

@kanaka , @samhed, @astrand  : any objections to merging this (one Travis finishes up)?

include/display.js
```diff
@@ -39,6 +47,10 @@ var Display;
         this._tile_x = 0;
         this._tile_y = 0;
 
+        this._holdingBuff = new Uint8Array(800 * 600 * 4);
+        this._holdingStart = 0;
+        this._holdingEnd = 0;
+
```
kanaka: I think these this._holding* defines are extraneous.

include/rfb.js
```diff
@@ -1706,37 +1806,51 @@ var RFB;
...
                 if (rowSize * this._FBU.height < 12) {
                     raw = true;
-                    clength = [0, rowSize * this._FBU.height];
+                    cl_header = 0;
+                    cl_data = rowSize * this._FBU.height;
+                    //clength = [0, rowSize * this._FBU.height];
                 } else {
-                    clength = RFB.encodingHandlers.getTightCLength(this._sock.rQslice(3 + paletteSize,
-                                                                                      3 + paletteSize + 3));
+                    // begin inline getTightCLength (returning two-item arrays is bad, ok?)
```
kanaka: :-)

include/rfb.js
```diff
@@ -1817,28 +1938,34 @@ var RFB;
...
-                    });
+                    // skip ctl byte
+                    this._display.fillRect(this._FBU.x, this._FBU.y, this._FBU.width, this._FBU.height, [rQ[rQi + 3], rQ[rQi + 2], rQ[rQi + 1]], false);
+                    this._sock.rQskipBytes(4);
                     break;
                 case "png":
                 case "jpeg":
-                    clength = RFB.encodingHandlers.getTightCLength(this._sock.rQslice(1, 4));
-                    this._FBU.bytes = 1 + clength[0] + clength[1];  // ctl + clength size + jpeg-data
+                    // begin inline getTightCLength (returning two-item arrays is bad, ok?)
```
kanaka: Is it worse than duplicating code 3 times? :-)

include/websock.js
```diff
@@ -321,26 +341,16 @@ function Websock() {
...
         _encode_message: function () {
-            if (this._mode === 'binary') {
-                // Put in a binary arraybuffer
-                return (new Uint8Array(this._sQ)).buffer;
-            } else {
-                // base64 encode
-                return Base64.encode(this._sQ);
-            }
+            // Put in a binary arraybuffer
+            // according to the spec, you can send ArrayBufferViews with the send method
```
kanaka: I have a vague recollection like I tried that and there were a few browsers that supported arraybuffers but didn't support send of Views. Certainly, that was a long time ago so probably irrelevant, but that's why it wasn't already doing that.

kanaka:
Looks good. A few minor comments inline.

include/rfb.js
DirectXMan12: I *may* occasionally get really sarcastic in my code comments :-P

include/rfb.js
DirectXMan12: Yeah, in this case.  Firefox's GC *really* didn't like all those objects.

include/websock.js
DirectXMan12: I can test around on Sauce Labs, but it works this way in my versions of FF and Chrome.

include/display.js
DirectXMan12: Yep.  I think they were left over from some testing

include/rfb.js
kanaka: Ah, okay. Just note that it was for performance in the comment and I'll be happy :-)

include/websock.js
kanaka: Yeah, I'm sure it hasn't been relevant for a long time. But doesn't hurt to make sure everywhere we are interested in supporting.

DirectXMan12:
For Firefox, I tested on Firefox 26 (which is older than the current ESR, so that should be a safe minimum version to test) and things seemed to work fine.

Chrome's release cycle means that we don't have to worry too much about older versions, but I tested on Chromium 44.0.2403.89, and it works fine there as well.

kanaka:
IE10 would be nice (some analysis seems to show it fairly steady around 5%) but certainly not necessary. What about mobile browsers? We should probably try and support current released versions of iOS Safari and Chrome Android.

Anyways, regardless, we should probably update the main README and the [browser support matrix wiki](https://github.com/kanaka/noVNC/wiki/Browser-support) to reflect what we actually intend to support going forward.

DirectXMan12:
Ok, I tested on iOS 7.1 and the native Android browser for Android 4.4, and they both work fine.  Chrome for Android should be good as well.  The most recent changes are just to pull out web-socket-js (which wouldn't have worked with the removal of base64 support).  Once Travis finishes, I'll merge

Issue #442, opened 2015-02-04
noVNC fork to make it suitable for Node/browserify 


ibc:
Hi developers,

Current noVNC design is not very "friendly" in the sense that it exports really lot of stuff to `window` (for example `Utils` and lot of functions), and that is not fine when it comes to avoid namespace conflict with other JS libraries.

Even more, the library is designed in a way that it loads its dependencies (which export more and more stuff into `window`).

So I've forked it and make it suitable for Node/browserify based applications:
- https://github.com/eface2face/noVNC-node
- [API documentation](https://github.com/eface2face/noVNC-node/blob/master/docs/index.md)

As explained in the [Overview](https://github.com/eface2face/noVNC-node/blob/master/docs/Overview.md), basic features/changes are:
- Can integrate **noVNC** into a Node/browserify application.
- Don't pollute the `window` namespace. When loaded via a `<script>` tag it just exposes the `noVNC` module/Object.
- Removed all the non-library stuff (for example `include/ui.js` or `include/webutil.js`).
- Removed Flash feature.
- Removed "no native WebSocket" feature.
- Use the Node [debug](https://github.com/visionmedia/debug) library for debugging.
- Use [gulp](http://gulpjs.com/) to lint, test and build the browserified library.

This fork lacks all the utils in the "utils/" folder (it would require some effort to migrate them into the fork's design).

So, would you interested in adopting this new design? Of course what I'd love is that the **noVNC** project accepts my fork :)

julien-f:
Good initiative!

I am using noVNC with a browserified project and it is very annoying to handle loading manually and to avoid concurrency problems.

DirectXMan12:
I'll certainly take a look when I get some time.

You're correct in that there are really two parts to the noVNC library itself: the JS files required for a working implementation, and the JS files used by the hosting pages (e.g. `ui.js` and `webutil.js`).  Since quite a few projects embed the noVNC, I definitely think it would certainly be worthwhile to make it easier on those users.

ibc:
Hi developers. Would you be interested in moving to my [noVNC fork](https://github.com/eface2face/noVNC-node) design so noVNC would become a library suitable for browserify/requireJS/commonJS/etc?

samhed:
> Hi developers. Would you be interested in moving to my noVNC fork design so noVNC would become a library suitable for browserify/requireJS/commonJS/etc?

Have some patience, @DirectXMan12 said he could take a look

DirectXMan12:
[...]

PR #496, opened 2015-06-12
Enable noVNC to become Browserifiable

DirectXMan12:
***WIP: Do Not Merge***

This commit restructures noVNC, splitting it into the core directory
and the app directory, with the former containing core noVNC parts,
and the latter containing parts specific to the application.

Additionally, it introduces a new utility for transforming the noVNC
javascript files into a browserifiable application.  Module names and
requirements are listed at the top of each file, and the script
`use_require.js` may be used to convert them into a form that uses
`module.exports` and `require`.  If you pass the '-b' flag to
the script, it will aslo run browserify, and copy over sufficient
resources to use 'vnc.html'.

The general idea here is that:

1. The current noVNC code doesn't need to change much
2. The ability to simply use script tags to include the files individually is preserved
3. noVNC parts (or the whole) can be used with Browserify
4. Files are marked with clear dependencies

#### TODO ####

- [x] remove Array modifications (much easier once #488 lands)
- [x] make playback.html work with the new directory structure (needs #488 parts to land)
- [x] make `Util.load_scripts` work with new directory structure
- [x] Move non-core functionality from `Util` to `WebUtil` (e.g. `Util.load_scripts`).
- [x] Overhaul logging?


DirectXMan12:
See also #442 
cc @ibc , @kanaka

kanaka:
@DirectXMan12 did a bit deeper review:

The from_push variable is kind of throwing me off (because both code path could be "from push" depending on your perspective). Maybe from_queue or something like that?

I suspect colourMap modification might need to be ordered with the renderQ to be fully correct so that it doesn't impact rendering of prior queue operations. Also, I think 'blit'/blitImage and finishTile should have the same renderQ logic as all the other operations. This is not a regression, but in reviewing the code, I think it's a hole in the existing logic and the restructuring just made it more obvious. These problems may not actually happen in practice (i.e. there may not be any actually servers that mix hextile and tight) but it should probably cleaned up and made consistent and correct across the board.

The refactor of messages.* functions does feel a bit more messy compared to before, in particular that they mutate the send queue but rely on the caller to mutate the sQlen variable (if it's not just the length already). Maybe passing in the queue and length as a single object and having the callee modify the length (and getting rid of the separate _mouse_arr). Is there a measurable performance impact for avoiding object creation on the send side of things? I haven't thought in depth about a clean way to do this but it does feel like there is some room for improvement here.

I think that's all for now. Otherwise, I think the changes look good and as long as cross-browser testing seems to be showing correct rendering, I'm good with the changes.

kanaka:
It's looking good and I like the overall structure and approach. A few comments:

It would be good to split the dropping of web-socket-js/tcp-socket.js and the rest into separate commits.

I like the split into app and core but I'm not super thrilled about having "../core" in script paths. I feel like that may be asking for trouble. I know, at least in the past, that some servers aren't too happy with "/../" in URLs. Also, if somebody happens to start their web server in `app/` rather than at the top (e.g. in order to maintain the same paths to vnc.html and vnc_auto.html as before), that will almost certainly break in most default web server configurations (denial of navigating up the path). Also, this will require all users of noVNC to update their own integration of noVNC to add "app" to the path (minor, but affects everyone). I think I would be more comfortable leaving the HTML files at the top for now.

Have we decided that noVNC owns websock.js and we sync over to websockify, or should websockify continue to own websock.js and sync into noVNC? I'm fine with whatever works most efficiently for development, but websock.js is used with websockify by other projects unrelated to noVNC, so project-wise it probably makes more sense for websockify to own it. Again, this is more dev process than anything, so I'm fine with whatever as long as everybody is on the same page.

samhed:
I agree with @kanaka's comments regarding the /core path.

Otherwise, I think this is great! I'm happy that we don't have to make such big changes to achieve this!

DirectXMan12:
> It would be good to split the dropping of web-socket-js/tcp-socket.js and the rest into separate commits. 

Will do.

> I think I would be more comfortable leaving the HTML files at the top for now.

The reasoning was to keep the top directory free from any of the main files (be they app or core), but I can move the HTML files back to the root directory.

> Have we decided that noVNC owns websock.js and we sync over to websockify, or should websockify continue to own websock.js and sync into noVNC? I'm fine with whatever works most efficiently for development, but websock.js is used with websockify by other projects unrelated to noVNC, so project-wise it probably makes more sense for websockify to own it.

So, I think that, development-wise, it makes more sense for noVNC to own websock.js, seeing as the majority of the changes to websock.js are motivated by changes in noVNC.

kanaka:
If we want to have a completely clean top-level we can move app, core and *.html into a src directory. Although, I'm not too broken up about a couple of html files at the top.

I'm okay with making websock.js in noVNC the primary place for development and then syncing back to websockify as long as we continue to support non-noVNC (ack, double negative) usage.

samhed:
@DirectXMan12 how does your plans look for continuing with this refactor? Is there anything you need help with?

DirectXMan12:
@samhed what I really need is some time ;-)  I'll see if I can get to it soon.  Worst comes to worst, I'll do some work over the holiday break.

DirectXMan12:
I've finally gotten a chance to work on this again.  It should address the comments above, and it's been rebased to reflect the most recent master.

@kanaka, @samhed PTAL

app/ui.js
```diff
@@ -8,18 +8,25 @@
...
-/* global window, $D, Util, WebUtil, RFB, Display */
+/* global window, document.getElementById, Util, WebUtil, RFB, Display */
 
+/* [module] name: UI; requires: Util, WebUtil, KeyTable, RFB, Display */
 var UI;
 
 (function () {
     "use strict";
 
+    //var resizeTimeout;
```
samhed: I assume this was a miss when rebasing, this variable has been moved inside the UI object.

samhed:
Except for the minor in-line comment and the failing test, it looks fine to me!

app/ui.js
DirectXMan12: Yep, will fix

kanaka:
I did a brief review again. I think it's looking good. Two small comments:

- Maybe consider splitting the $D to document.getElementById into a separate commit so that the more meaty changes are more concisely captured in a separate commit.
- most of these changes are in the UI/app space, but we should still make sure to do a before and after perf test across various browsers. Some of the object access changes look like the type of silly things that could have the potential to affect performance or GC behavior.

DirectXMan12:
> * Maybe consider splitting the $D to document.getElementById into a separate commit so that the more meaty changes are more concisely captured in a separate commit.

:+1:

> * most of these changes are in the UI/app space, but we should still make sure to do a before and after perf test across various browsers. Some of the object access changes look like the type of silly things that could have the potential to affect performance or GC behavior.

I tested locally on FF and Chromium, but I'll see if I can do some more broad performance testing.

ibc:
Honestly, this is not the way a modern library must be designed. You say that this new approach "provides browserify capabilities" but it is not true, it provides an "already browserified" file, and that's not the Node/browserify way at all (you don't load already browserified libraries into a Node/browserify based project).

Node based applications (and hence browserify based projects) are not designed by importing *already* browserified libraries, but instead it is the final application who decides whether it wants a browserified version of all the app or not (useful to run in a browser, of course).

If you still want to allow legacy `<script>` usage you can provide an already browserified file within the project (under a `dist/` directory) as I did in my [noVNC-node](https://github.com/eface2face/noVNC-node) fork. Then you can even create a bower.json conf file pointing to such a `dist/xxxx.js` so Bower users are also happy.

But if you still want to allow the user to load "just a few files" via separate `<script>` tags, then I cannot understand any rationale for that. If this library can be splitted into separate "modules" then it SHOULD be separated into different libraries, each one providing its own browserified file for those users using `<script>` tags or Bower. Then CommonJS/AMD/Node/Browserify/WebPack users could include those separate libraries as project dependencies and load them on demands.

No JS library in the world should provide a so complex "build system" as the provided in this proposal. Do a single thing and do it well, rather than including "addEventListeners" wrapper within a library, HTML selectors (`$D`) or any out-of-the-scope stuff.

Sorry guys, I think you are completely wrong in your approach.

kanaka:
@ibc I understand your desire to have noVNC be a "modern" JavaScript library. I would like the core JavaScript part of noVNC to eventually get to that point too but not at the cost of disrupting thousands of downstream users and integrators. This change makes a lot of the structural changes needed to move towards better modularity without foisting major disruption to the existing usage and integration patterns. 

noVNC is not just a JavaScript library for handling the RFB protocol. It's primary goal (the one thing it tries to do well) is to be a well performing VNC web application that can be used standalone but that is also fairly straightforwards to integrate into other larger web applications. A secondary goal (that supports the main goal) is for the code to be designed as modules/libraries that have clear API boundaries and are easily testable. If that is done well, that means the individual modules/libraries can be used separately. Having those modules/libraries packaged and deployed separately (e.g. as node modules) is a tertiary goal. It's a good thing, but it's not high on my personal priority list. Obviously, for your personal use case that goal a high priority for you. I want to support you in that but not at the cost of the primary goals/existing users. The change that @DirectXMan12 has proposed moves the code base significantly in that direction.

kanaka:
@DirectXMan12 BTW, one thing I thought of related to this discussion is that it's relatively easy to create JavaScript files that support node.js style require/exports and can still be used directly in the browser, for example, in mod1.js that depends on mod2.js:

For example, mod1.js:

```javascript
var mod1 = {};
if (typeof module === 'undefined') {
    exports = mod1;  // browser
}

exports.func1 = function(name) { console.log('hello ' + name); }
```

And mod2.js:

```javascript
var mod2 = {};
if (typeof module === 'undefined') {
    exports = mod2;  // browser
} else {
    var mod1 = require('./mod1');  // node/CommonJS
}

exports.func2 = function(name) {
    for (var i=0; i<10; i+=1) { mod1.func1(name); }
};
```

The browser then has script tags with both mod1.js and mod2.js (no change), and from node you can now require mod2.js as a normal node module and it will pull in the mod1 dep. I've used that technique a fair bit in the past to good effect. Might get us to node modules without needing to mangle things in either direction with a build script or conversion script. Certainly the restructuring changes you've already posed would make doing that easier and cleaner. Just a thought I had while mulling this over.

DirectXMan12:
> Honestly, this is not the way a modern library must be designed. You say that this new approach "provides browserify capabilities" but it is not true, it provides an "already browserified" file,

Please take a look more closely at `utils/use_require.js` -- if you simply run `utils/use_require.js` without the `-b` flag, it will create a bunch of separate modules suitable for use individually as part of a browserified app (they get written to `build/`).  These output modules are individual node-style modules, *not* pre-bundled.  If you do use the `-b` flag, it will create a browserified bundled app, but this is mainly for testing convinience.

> No JS library in the world should provide a so complex "build system" as the provided in this proposal. Do a single thing and do it well, rather than including "addEventListeners" wrapper within a library, HTML selectors ($D) or any out-of-the-scope stuff.

Yes, and I've been trying to remove some of this compatibility cruft, since most of it is not needed any more -- we've removed `$D` entirely in the pull request, for instance.

Please keep in mind, as @kanaka said, that noVNC is first an application, and it's an application with quite a bit of legacy code in it.

DirectXMan12:
> @DirectXMan12 BTW, one thing I thought of related to this discussion is that it's relatively easy to create JavaScript files that support node.js style require/exports and can still be used directly in the browser, for example ... in mod2.js:

> ```javascript
var mod2 = {};
if (typeof module === 'undefined') {
    exports = mod2;  // browser
} else {
    var mod1 = require('./mod1');  // node/CommonJS
}

> exports.func2 = function(name) {
    for (var i=0; i<10; i+=1) { mod1.func1(name); }
};
```

@kanaka While I like the idea of not having to do mangling, this is fragile, in that it immediately breaks if someone tries to use noVNC in a page with defines `module` at the top level, for some reason or another.   I'd like to remain compatible with non-browserified apps (although, once there's browser support for ECMAScript modules, I'd strongly consider using those natively), and I can see that breaking by accident pretty easily.

**Comment by kanaka** [comment]
@DirectXMan12 Yep, that's a good point. We could do a more sophisticated check such as for existence of require, module and module.exports and that module.exports is an empty map. That would give pretty high confidence, but you're right that it could still cause collisions. Just thought I would throw it out there as food for thought.

**Comment by DirectXMan12** [comment]
Alright, I think this is entering the Final Comment Period.

@samhed, @kanaka do you see anything I missed?  I'll probably do one final performance evaluation.

**Comment by samhed** [comment]
No, I can't find anything. Looks very good to me, thanks for doing this!

**Comment by samhed** [comment]
Let's get this merged! Do you want me to resolve the merge conflicts?

**Comment by last-partizan** [comment]
Any news on when this will be merged?

**Comment by olivierlambert** [comment]
Hi @DirectXMan12 !

Any news on that?

**Comment by DirectXMan12** [comment]
I started updating the PR this weekend to be based off of the latest code (but got interrupted due to forgetting my laptop charger ;-)).

**Comment by kanaka** [comment]
@DirectXMan12 just think how much important work the world has lost due to not having a charger a key moments :-)

**Comment by olivierlambert** [comment]
@DirectXMan12 cool thanks!

**Comment by olivierlambert** [comment]
@DirectXMan12 Have you found back your laptop charger? ;)

**Comment by DirectXMan12** [comment]
> @DirectXMan12 Have you found back your laptop charger? ;)

~~Oh, boy, did I ever!~~ (wait, that sounds weird)

This should include all of the latest up-to-date noVNC changes, plus a few differences from the last revision:

1. The "module description in comments" is now basically just commented out ES6 import/export syntax.  This should make it easy for us to transition to native ES6 import/export once enough browsers support it (of course, at least one browser has to support it to start along that path :wink:).  It also makes it much easier for people to write and understand.

2. The transformation is now mainly done using browserify and babel.  A custom browserify transform uncomments the ES6 import/export code, which can then be passed to the babelify transform with at least the es2015 preset.  This significantly simplies the build code.

3. package.json now actually builds a browserify-compatible set of scripts on pack/publish/install/etc!  They'll get built into `lib/`, which is the only source folder which gets included in the npm package.

To reiterate, the tool to build browserify- and/or ES6-compatible files is `utils/use_require.js`.  When called on its own, it will generate ES6-compatible files (from `core/`) into `lib/`.  Things get a bit funky with Pako (since it's a CommonJS module), but that's left as an exercise to the reader.  When run with `--as-require`, it will instead generate CommonJS/browserify-compatible files (Pako works fine here) into `lib/` (from `core/`).  Finally, when run with the `-b` option, it will generate a fully-bundled app including `vnc.html` and all the app UI and whatnot.  This is mainly intended as an example, but can also be used as a convenient way to deploy a bundled copy of noVNC.

Otherwise, the same changes as in the previous revision apply (things got shuffled around, cruft was removed, etc).

@olivierlambert @131 PTAL and let me know how this works for you

@kanaka and @samhed did I miss anything?

If this looks good to folks, I'll merge, and we can even set up to publish "noVNC, the library parts" on NPM.

### Future Work ###

So, this uses ES6 modules, but only kind of.  Most of the modules simply export a default object containing most of the code.  This is mainly so that things easily continue working when the files are loaded one-by-one.  However, in the future, I'd like to have more proper module organization, so that instead of `import Base64 from "./core/base64"; Base64.encode and Base64.decode` we have `import * as Base64 from "./core/base64"`, etc.  Also, it would be nice to turn some of the objects into ES6 classes, just for readability's sake.

File: app/webutil.js [hunk 2]
```diff
@@ -278,4 +278,73 @@ WebUtil.injectParamIfMissing = function (path, param, value) {
...
+
+                // Call window.onscriptsload after last script loads
+                if (ps.length === 0 && window.onscriptsload) {
+                    window.onscriptsload();
+                }
+            }
+        }
+    };
+
+    var root_dirs = Object.Keys(files_by_dir);
```
samhed: > Uncaught TypeError: Object.Keys is not a function

samhed:
You forgot specifying the module when accessing key symbols from keysym.js in ui.js. For example, XK_Control_L should be KeyTable.XK_Control_L.

131:
I'm very sorry, but i keep thinking this "half" step is a mistake
* As babel a browserify transform, that mean that "legacy usage" of the standalone files (i.e. loading "base64.js only) will not be proceed by babel, hence, cannot use any es6/babel features
This syntax / comment design pattern is nowhere else to be found, and i do not think at it as a smart move (what problem is it solving ??)

@DirectXMan12  , just look how close you are to a fully "standard" [es6 design](http://www.mergely.com/BBomUaiT/) (this is taken from my [rage-fork](https://github.com/131/noVNC2/tree/master/lib) ) . I really think you should rethink this and go a little step further.

app/webutil.js
DirectXMan12: fixed.  I swear I fixed that one, must have gotten mangled when I went to stick it in the right commit.

DirectXMan12:
> You forgot specifying the module when accessing key symbols from keysym.js in ui.js. For example, XK_Control_L should be KeyTable.XK_Control_L.

Good catch, fixed

DirectXMan12:
> As babel a browserify transform, that mean that "legacy usage" of the standalone files (i.e. loading "base64.js only) will not be proceed by babel, hence, cannot use any es6/babel features

People could always run babel separately if they wanted to (from the Babel CLI/their build tool of choice).  I'd like to (and I imagine @kanaka would like to as well) keep the set of Javascript that we use limited to what can be natively supported by our target browsers.  ES6 classes look like they're supported in Firefox ESR now (at a quick glance, I'll have to do a bit more looking to double check), so that's definitely something I'd like to consider for another PR.

> This syntax / comment design pattern is nowhere else to be found, and i do not think at it as a smart move (what problem is it solving ??)

It means we can easily build the code into browserify-compatible files and/or ES6 modules, and publish those on NPM, etc, without losing the ability to work natively in browsers without any processing.  For the time being, that's something we'd like to keep.  It also means that when ES6 module support finally lands in browsers, we just have to remove the `/* [module]` and `*/` to convert the code to using that.

AFAICT (correct me if I'm wrong), your "rage-fork" primarily does three things:

1. It moves to ES6 classes (which I'd like to consider in a follow-up PR)
2. It moves to using `require` (which I address in this PR)
3. It moves input-related code into a subdirectory (which I actually really like, and will probably try and do that here).

Number 1 should come in a follow-up PR, number 2 is the output of this PR (e.g. the output of `utils/use_require.js --as-require`, or `npm pack`, which will most likely be published on NPM for easy consumption), and number 3 should be coming along shortly.  I noticed you also switched to using an external debug library, but it should be fairly easy to get noVNC's lightweight wrappers to work with that (set `Util.Debug = require('debug')('IDENTIFIER')`).

DirectXMan12:
Alright, moved input-related stuff into `input/` (makes core a bit easier to read), and trimmed down the list of transformations applied by babel to just the import/export-related ones.

DirectXMan12:
@samhed @kanaka assuming it passes tests, ok to merge?

samhed:
Ok for me!

kanaka:
@DirectXMan12 Apart from the test failures, LGTM.

And yes, I still definitely still in favor of maintaining the ability to clone the repo and go right away without any processing (by end-users or by us to generate artifacts in the repo). noVNC is both end-user product and dev library which is a bit unique. The end-users are first (IMO). Also, selfishly, occasional complaints by developers about weird structure I can deal with. I'd rather not deal with numerous users complaining about "noVNC not loading in the browser", or "node/npm install issues", or "some obscure npm error because their version of npm is too old". I think you've arrived at a solution that allows reasonable packaging solution without losing that goodness. With browsers rapidly converging towards module support, I suspect the code will look pretty standard soon enough and still be usable without any pre-processing.

olivierlambert:
\o/

Issue #498, opened 2015-06-17
noVNC in fullscreen-mode


samhed:
Especially on devices with small screens it can be very useful to be able to run noVNC in fullscreen.

Fullscreen is not supported in Safari on iOS. The only way to achieve this is through adding to homescreen and in that way run the page in WebApp mode.

samhed:
http://cubiq.org/add-to-home-screen could possibly be interesting wrt Safari on iOS.

samhed:
Fullscreen is not supported in Chrome on iOS as well. But Chrome on iOS isn't really Chrome, just a re-skinned safari.

samhed:
Fullscreen not supported on IE 10:

http://caniuse.com/#feat=fullscreen

PR #499, opened 2015-06-17
Add option to toggle fullscreen mode

samhed:
An alternative for solving #498


DirectXMan12:
Is it possible to do feature sniffing instead of UA detection?  Do the appropriate API bits still show up on Safari for iOS?

DirectXMan12:
(Otherwise, I like the idea)

samhed:
Yes it is! I was on my way updating the PR for that yesterday, but got hung up on problems with IE 11 and Safari on OS X:

* Safari disables keyboard input while in fullscreen (as a security measure apparently)
* The canvas goes blank if I toggle fullscreen while a session is running in IE 11 and the only way to get it back to working is to disconnect-connect.. I'm not sure what causes this behavior. Apparently scrollbars are not shown in IE 11 while in fullscreen, the issue could possibly be related.

samhed:
The problem with the blank canvas in IE 11 seems to be resolved if I use

    document.body.msRequestFullscreen();

instead of

    document.documentElement.msRequestFullscreen();

According to http://caniuse.com/#feat=fullscreen the scrollbars problem is limited to when document.documentElement is set to fullscreen, but I still can't get scrollbars when using the body element..

samhed:
I have updated the commit. The remaining problem is that we don't get scrollbars in fullscreen mode on IE 11.

The question is whether we should disable the fullscreen toggle button on IE 11 when not in clipping-mode. Scrollbars work in fullscreen in every other desktop browser that I have tested. What do you think @DirectXMan12 ?

DirectXMan12:
> The question is whether we should disable the fullscreen toggle button on IE 11 when not in clipping-mode.

The other option would be to enable clipping when we go full screen on IE 11.  Either option works, as long as we provide a clear documentation (perhaps a tooltip or something) as to why fullscreen is disabled.

> The other option would be to enable clipping when we go full screen on IE 11.

So... if someone is in read-only mode, sharing the desktop, it's not *that* useless (also, apparently you can use `document.body.webkitRequestFullScreen(Element.ALLOW_KEYBOARD_INPUT)` to allow keyboard input -- see https://developer.mozilla.org/en-US/docs/Web/Guide/API/DOM/Using_full_screen_mode)

samhed:
>The other option would be to enable clipping when we go full screen on IE 11.

The noVNC settings are locked while in a session but yes, I guess we make an exception for this. I'll look into the practical details on Monday.

I am already using the Element.ALLOW_KEYBOARD_INPUT flag, but unfortunately it doesn't seem to be respected by Safari. The only word from Apple I can find with regards to this is a private reply to a bug report which is quoted here: http://stackoverflow.com/questions/16150601/substitute-for-allow-keyboard-input-javascript-full-screen  

Also, see: https://bugs.webkit.org/show_bug.cgi?id=121496

It should be clarified that it is "only" alphanumerical input that is blocked in fullscreen mode in Safari. Arrow-keys, space, backspace etc. works.

kanaka:
@samhed actually, module properties have a setting that indicates whether they can be changed after the object is constructed or not (second element of table passed to Util.make_properties), but for settings that actually have an effect when changed after connecting, you might need custom setters for the property. For example, search for "set_local_cursor" in rfb.js

Also, as I recall on some mobile devices (especially Android), in order to get regular keyboard input, we had to focus the cursor in a _nearly_ invisible text box (search for keyboardinput and keyInput). I suspect that you might be running into complications with that with respect to full screen. This (textarea focus on something outside fullscreen element) was probably never really considered deeply by browser vendors so the exact behavior probably varies.

samhed:
> Also, as I recall on some mobile devices (especially Android), in order to get regular keyboard input, we had to focus the cursor in a nearly invisible text box (search for keyboardinput and keyInput). I suspect that you might be running into complications with that with respect to full screen. This (textarea focus on something outside fullscreen element) was probably never really considered deeply by browser vendors so the exact behavior probably varies.

There doesn't seem to be any issues in Firefox or Chrome on Android. But since this patch sets the entire page (documentElement) to fullscreen, the keyboardinput element is included.

samhed:
As you can see in the latest commit, I have made it so that you always get clipping mode when entering fullscreen (eventually we only want this on IE). But I have encountered an issue with scaling and it's easier to test if I don't have to switch to my windows machine for testing.

If I have a VNC session running with a size smaller than the browser window and exit fullscreen, the session is stretched vertically.. I can't see why! Do you have any ideas @DirectXMan12 ? 

Added WIP label for now, since this is not the intended final behavior.

samhed:
I fixed the issue and I am satisfied with this PR now. Any comments?

To summarize limitations with javascript fullscreen:
* Fullscreen is not supported in any browser on iOS (you have to add to homescreen).
* Alphanumerical keyboard input is not supported in Safari while in fullscreen.
* Scrollbars do not show in IE 11 while in fullscreen.

This PR:
* Adds a fullscreen toggle button to the controlbar.
* The fullscreen button is not shown when fullscreen support is missing.
* The fullscreen button is not shown in Safari.
* Clipping mode is forced on IE when scaling is disabled and we are in fullscreen.
* A popup status message is shown when entering fullscreen mode in IE to clarify what is happening.

I have tested with the latest versions of the browsers on:
* IE on Windows 8
* Firefox on Windows 8, Android 5.2, Fedora 20 and OSX 10.10
* Chrome on Windows 8, Android 5.2, iOS 8.3, Fedora 20 and OSX 10.10
* Safari on iOS 8.3 and OSX 10.10

samhed:
Unless you have any objections, I will go ahead and merge this tomorrow

PR #511, opened 2015-07-15
onInteraction callback: makes it possible to attach callbacks to client machine interaction events

robertsoakes:
This a patch which make it possible to attach onInteraction events. Guru Labs uses noVNC in a remote collaboration and instruction product (rlabs.gurulabs.com), and the ability to notify users of interaction is important to syncing other UI state.

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -142,6 +142,7 @@ var RFB;
             'onFBResize': function () { },          // onFBResize(rfb, width, height): frame buffer resized
             'onDesktopName': function () { },       // onDesktopName(rfb, name): desktop name received
             'onXvpInit': function () { },           // onXvpInit(version): XVP extensions active for this connection
+            'onInteraction': function () { },       // onInteraction(): Callback invoked on mouse button press or keyboard press
         });
 
         // main setup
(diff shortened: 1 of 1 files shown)


DirectXMan12:
So, the intent is to be able to do extra work whenever noVNC detects a keyboard or mouse event? it looks pretty straightforward, but it doesn't actually provide any information to the callback, so the use seems to be limited outside of your specific use case.

@kanaka @samhed @astrand what do you think?

samhed:
I can't say I find this particularly useful

kanaka:
@robertsoakes can you clarify what this accomplishes that you can't do by just directly hooking into the mouse and keyboard events for the noVNC target canvas element?

robertsoakes:
It's mostly a matter of convenience. We have a loose wrapper (implemented as a Backbone.js view), that we use to translate noVNC events (via the callbacks) to Backbone.js events. Having the onInteraction method makes it very easy to know when data is being explicitly passed to the connection without having to filter out stray clicks to not yet active connections. Binding event handlers to the canvas would also be pretty straightforward, but we opted to go with the most direct approach.

Our biggest use case for the interaction callback right now is signalling. We've got one interface which is used for technology training that can potentially have a very large number of active and shared connections (frequently above 30). To manage that many sessions in the UI, it's very useful to know if there has been a significant amount of time since last the last time the user did something. If the connection is "inactive" for 5 minutes or so, we can drop it for the collaborators and then bring it back when someone begins to do something again. (The collaboration sync happens over a separate socket connection which is attached to a message server.)

robertsoakes:
Based on your comments, I found a workaround which meets our needs. Instead of adding a direct callback, we instead create a mouse/keyboard that includes the signalling we need. For that reason, I'm closing this pull request.

PR #513, opened 2015-07-30
Fix occasional error with sending while readyState !== 1

SirCmpwn:
Under certain conditions, noVNC would attempt to flush the web socket
while it was disconnected, before the disconnected state was picked up.
This casues noVNC to crash ungracefully and the parent window is not
notified - leading to no chance at recovery without a page refresh.

### DIFF
--- a/include/websock.js
+++ b/include/websock.js
@@ -175,7 +175,7 @@ function Websock() {
             }
 
             if (this._websocket.bufferedAmount < this.maxBufferedAmount) {
-                if (this._sQ.length > 0) {
+                if (this._sQ.length > 0 && this._websocket.readyState === 1) {
                     this._websocket.send(this._encode_message());
                     this._sQ = [];
                 }
--- a/include/websock.js
+++ b/include/websock.js
@@ -175,7 +175,7 @@ function Websock() {
             }
 
             if (this._websocket.bufferedAmount < this.maxBufferedAmount) {
-                if (this._sQ.length > 0 && this._websocket.readyState === 1) {
+                if (this._sQ.length > 0 && this._websocket.readyState === WebSocket.OPEN) {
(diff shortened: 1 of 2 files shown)


SirCmpwn:
Looks like this is failing due to a problem in the test itself. If the readyState is not 1, then we shouldn't be sending anyway. It would fail in a real browser.

SirCmpwn:
Alright, let's see if Travis likes this.

SirCmpwn:
:tada:

include/websock.js
```diff
@@ -175,7 +175,7 @@ function Websock() {
             }
 
             if (this._websocket.bufferedAmount < this.maxBufferedAmount) {
-                if (this._sQ.length > 0) {
+                if (this._sQ.length > 0 && this._websocket.readyState === 1) {
                     this._websocket.send(this._encode_message());
```
kanaka: It would be better to compare against the ready state constants rather than a literal 1. E.g. WebSocket.OPEN

SirCmpwn: Fixed by https://github.com/SirCmpwn/noVNC/commit/5122a8a4ce7cb4958c62ef1c84ce4220f3a3ab0d.

tests/test.websock.js
```diff
@@ -180,6 +180,7 @@ describe('Websock', function() {
             it('should actually send on the websocket if the websocket does not have too much buffered', function () {
                 sock.maxBufferedAmount = 10;
                 sock._websocket.bufferedAmount = 8;
+                sock._websocket.readyState = 1;
```
DirectXMan12: Shouldn't this also be `WebSocket.OPEN`?

DirectXMan12:
LGTM (other than the above comment) :+1:

samhed:
Merged in 3120c9b

PR #524, opened 2015-08-26
Fix buffer over-reads in handle_tight

### LINKED ISSUE #522: How to securely integrate noVNC inside an authetication system ?
Hi,

1.)
I have an authentication site (let's say "https://www.abcd.com/login.php") where a user puts in his login credentials and logs in to his account website. Now I want to put there in his account a link or an iframe which lets him auto connect to his VNC server. At the moment the only way I know is to point to the URL "https://www.abcd.com:6080/vnc_auto.html?password=1234". But here I would like to have a solution where the URL should be only reacheable inside / embedded in the authentication system that means that the URL should be something like "https://localhost/vnc_auto.html?password=1234" so that the URL is only accessible for the user which logged in his account site and not for others users from outside which could guess an URL and port with a password. Can you please give some advice how to do that? Is it also sensful to put the "?password=1234" in the URL in terms of security or is there another possibility to set the password for an auto connection?

2.)
Besides my apache server is configured for ssl with listen port 443. Is it also possible to "mix" or "redirect" the websockify webserver with the apache server on the same port 443?

3.)
I have problems using the option "--target-config". 

Here is the command:
root@testserver:~/noVNC# ./utils/websockify --verbose --cert ./self.pem --web ./ 6080 --target-config ./token.list

The file "token.list" contains the line:
abcd: 192.168.124.123:5900

The URL is:
http://testserver:6080/vnc_auto.html?token=abcd

And here is the error message:
Traceback (most recent call last):
  File "/root/noVNC/utils/websocket.py", line 874, in top_new_client
    client = self.do_handshake(startsock, address)
  File "/root/noVNC/utils/websocket.py", line 809, in do_handshake
    self.RequestHandlerClass(retsock, address, self)
  File "/root/noVNC/utils/websocket.py", line 112, in __init__
    SimpleHTTPRequestHandler.__init__(self, req, addr, server)
  File "/usr/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/root/noVNC/utils/websocket.py", line 540, in handle
    SimpleHTTPRequestHandler.handle(self)
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
  File "/root/noVNC/utils/websocket.py", line 506, in do_GET
    if not self.handle_websocket():
  File "/root/noVNC/utils/websocket.py", line 494, in handle_websocket
    self.new_websocket_client()
  File "./utils/websockify", line 48, in new_websocket_client
    (self.server.target_host, self.server.target_port) = self.get_target(self.server.target_cfg, self.path)
  File "./utils/websockify", line 94, in get_target
    raise self.EClose("Token not present")
AttributeError: ProxyRequestHandler instance has no attribute 'EClose'
Reaing zombies, active child count is 6
Ignoring interrupted syscall

What happens here?

Thanks.

I'm using this release:
https://github.com/kanaka/noVNC/releases/tag/v0.5.1

DirectXMan12:
For performance reasons, the `handle_tight` function skips the
use of the receive queue API and uses the raw receive queue directly.
Because of the way that typed array receive queue gets reused, this
introduced the potential for buffer over-reads.

To address this, a new function, `rQwhole`, was introduced.
`rQwhole` simply returns a new view into the receive queue that
starts at 0 and ends at the current recorded end of the queue.

`handle_tight` now makes use of this function.

Fixes #522

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -1782,8 +1782,8 @@ var RFB;
                 return dest;
             }.bind(this);
 
-            var rQ = this._sock.get_rQ();
             var rQi = this._sock.get_rQi();
+            var rQ = this._sock.rQwhole();
             var cmode, data;
             var cl_header, cl_data;
 
--- a/include/websock.js
+++ b/include/websock.js
@@ -154,6 +154,10 @@ function Websock() {
             this._rQi += len;
         },
 
+        rQwhole: function () {
+            return new Uint8Array(this._rQ.buffer, 0, this._rQlen);
+        },
+
         rQslice: function (start, end) {
             if (end) {
                 return new Uint8Array(this._rQ.buffer, this._rQi + start, end - start);


DirectXMan12:
@kanaka @samhed any objections to this?

kanaka:
Looks fine to me although I'm seeing travis errors.

DirectXMan12:
That was just Travis having a moment, I think

PR #537, opened 2015-10-01
Pass token into the path variable

### LINKED ISSUE #536: vnc_auto.html - Token not being passed in the path
It seems that the token is not being passed to the path variable which is causing token connections to fail. Is this a bug or is there something I'm missing? 

I added a line concatenating the token variable into the path variable to test and that seemed to fix the issue, if this is indeed a bug I'd be happy to create a pull request to fix it.

miguelxpn:
Fixes #536

### DIFF
--- a/include/ui.js
+++ b/include/ui.js
@@ -96,6 +96,7 @@ var UI;
             UI.initSetting('view_only', false);
             UI.initSetting('path', 'websockify');
             UI.initSetting('repeaterID', '');
+            UI.initSetting('token', '');
 
             var autoconnect = WebUtil.getQueryVar('autoconnect', false);
             if (autoconnect === 'true' || autoconnect == '1') {
@@ -519,6 +520,7 @@ var UI;
                 UI.connSettingsOpen = false;
                 UI.saveSetting('host');
                 UI.saveSetting('port');
+                UI.saveSetting('token');
                 //UI.saveSetting('password');
             } else {
                 $D('noVNC_controls').style.display = "block";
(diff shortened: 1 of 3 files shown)


DirectXMan12:
@miguelxpn : I think this could have issues if someone already has `?abc=xyz` on the end of `path` -- you should check to see if we're already in a query string and then append `&token=blah` instead of `?token=blah`

miguelxpn:
@DirectXMan12 You are absolutely right, I changed my commit and it should be ok now.

miguelxpn:
Made a new commit adding a separate token variable in vnc.html and ui.js, if the token is already present in the path then this new variable is ignored so it doesn't break anything for people using it in the old way.

include/ui.js
```diff
@@ -810,7 +812,17 @@ var UI;
             var host = $D('noVNC_host').value;
             var port = $D('noVNC_port').value;
             var password = $D('noVNC_password').value;
+            var token = $D('noVNC_token').value;
             var path = $D('noVNC_path').value;
+
+            //if token is in path then ignore the new token variable
+            if (token && !(path.indexOf("token") > -1)){
```
DirectXMan12: Doesn't this cause the token insertion not to trigger if I have
`path = "/path"`, in which case a path of `path?path=[foo]` is perfectly valid.

DirectXMan12: We should be able to fix this by using the an `<a>` element in memory, and using that to manipulate the path.

DirectXMan12:
Besides the issue I noted, this LGTM :+1:.  I'll make the change, rebase, and merge.

PR #543, opened 2015-10-16
Fixes VNC standard auth using onPasswordRequired callback

### LINKED ISSUE #542: Standard vnc auth broken
Standard VNC auth with a password is broken, because a missing return in_negotiate_std_vnc_auth makes it proceed with a blank password instead of waiting for the password dialog. The following 1-liner fixes it:

    --- a/include/rfb.js
    +++ b/include/rfb.js
    @@ -739,6 +739,7 @@ var RFB;
                     // an RFB state change and a UI interface issue
                     this._updateState('password', "Password Required");
                     this._onPasswordRequired(this);
    +                return false;
                 }
 
                 if (this._sock.rQwait("auth challenge", 16)) { return false; }

DvdGiessen:
Should return and wait for the sendPassword-call to trigger the actual authentication. Fixes #542.

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -739,6 +739,7 @@ var RFB;
                 // an RFB state change and a UI interface issue
                 this._updateState('password', "Password Required");
                 this._onPasswordRequired(this);
+                return false;
             }
 
             if (this._sock.rQwait("auth challenge", 16)) { return false; }


DirectXMan12:
LGTM :+1:

Issue #544, opened 2015-10-17
Use hash fragment to supply parameters instead of query string


ganeshv:
For privacy, it is better to supply parameters like `host` and `port` in the hash fragment rather than the query string. The query string is visible to the server which hosts the noVNC HTML/JS and intermediate proxies, while the hash fragment is not. Javascript can as easily use the hash fragment as the query string as a source of parameters.

I'm developing a [VNC backend for Qt 5](https://github.com/pigshell/qtbase), where one can run a Qt 5 app headless, connect to it on a browser http://app.host:5900 which redirects to http://novnc.host/qtvnc.html#host=app.host&port=5900

I have modified my noVNC fork to to take parameters in the hash fragment. You can see the code [here](https://github.com/pigshell/noVNC/commits/qtvnc).

kanaka:
@ganeshv I'm not sure how useful that is host/port, but certainly for token or password I can see why that would be desirable. Adding that as an optional way to provide variable values is fine with me. Please send a PR with just that part (not the qtvnc) portion.

ganeshv:
@kanaka IMHO, from a security and privacy standpoint, the less information leaked the better, even if we have no idea how it might be exploited. host/port leaks the location of a live VNC server to whoever is hosting the viewer, and to all intermediate proxies and CDNs. In case of no password, or a weak/known password, this makes it easier for an intermediary to snoop on a VNC session. This is especially avoidable since it is not technically absolutely necessary.

I have made a [first pass](https://github.com/pigshell/noVNC/commit/ef284971dd138877130b96a685c11bd6375d57bb) at coding this up. Any config variable may be present either in the query string or hash fragment, with the hash fragment being preferred in case they are present in both.

Let me know if this approach looks OK and I will dot the i's and submit a pull request.

kanaka:
@ganeshv Yes, providing the option for noVNC integrators to avoid leaking host/port (or any other config variable) is fine and I support that.

However, I should point out it's not the host/port of the VNC server that is leaked, it's the host/port of the websocket/websockify server. Most integrators will be using the tokenfile plugin (or something similar) to switch between backend VNC targets and not connecting to anything without a token. As long as the tokens are opaque/strong and recreated for each connection, there really isn't much info that is leaked except the websocket address. Having the token in the hash will certainly help with certain attacks (e.g. replay), but hiding the host/port (which point to an open port on the Internet) should not be relied on since that can be discovered/predicted by other means (port scan, MitM).

[...]

PR #547, opened 2015-10-23
Add hash fragment as an optional method to supply config variables.

ganeshv:
Any config variable like host, port, password, token may be
specified either in the query string (like now), or in the URL hash
fragment. In case a given variable is present in both, the value in the
fragment takes precedence. Supplying variables in the fragment avoids
leaking them to the web server hosting the noVNC viewer HTML.

### DIFF
--- a/include/ui.js
+++ b/include/ui.js
@@ -97,7 +97,7 @@ var UI;
             UI.initSetting('path', 'websockify');
             UI.initSetting('repeaterID', '');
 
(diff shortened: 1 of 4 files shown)


DirectXMan12:
Related: #544

DirectXMan12:
LGTM :+1:

ibc:
Still mixing library and demo app guys?

PR #548, opened 2015-11-05
Localization support added for messages

LoekJanssen:
I created localization in German, Dutch and English for the messages in
the top messagebar and some logs.

### DIFF
--- a/include/locale/locale-de-DE.js
+++ b/include/locale/locale-de-DE.js
@@ -1,0 +1,20 @@
+Language = {
+    connectedUnencrypted:   'Verbunden mit (unverschlüsselt)',
+    connectedEncrypted:     'Verbunden mit (verschlüsselt)',
+    errorHostPort:          'Richten Sie Host und Port ein',
+    ready:                  'NoVNC bereit: Native WebSockets',
+    readyEmulation:         'NoVNC bereit: Emulierte WebSockets',
+    authentication:         'Authentifizierung',
+    errorFatal:             'Schwerwiegender Fehler. Die Verbindung kann nicht hergestellt werden.',
+    errorTimeout:           'Timeout. Verbindung unterbrochen',
+    errorConnected:         'Fehler beim Verbinden',
+    errorInitializing:      'Fehler beim Initialisieren',
+    startingVNC:            'Starte VNC Authentifizierung',
+    errorDisconnected:      'Verbindung zum Server getrennt',
+    errorConnection:        'Verbindung zum Server konnte nicht hergestellt werden',
+    errorProtocolVerion:    'Unvollständige Protokollversion',
+    errorServerVersion:     'Ungültige Serverversion',
+    errorSecurity:          'Sicherheitsfehler',
+    errorAuthFailure:       'Authentifizierungsfehler',
+    errorAuthAttempts:      'Zu viele Authentifizierungsversuche'
+};
--- a/include/locale/locale-en-GB.js
+++ b/include/locale/locale-en-GB.js
@@ -1,0 +1,20 @@
+Language = {
+    connectedUnencrypted:   'Connected to (unencrypted)',
+    connectedEncrypted:     'Connected to (encrypted)',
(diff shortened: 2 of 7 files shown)


DirectXMan12:
Localization support would be great!  However, could you please fix the failing unit tests so that they work with you patch?  Thanks!

LoekJanssen:
Nice to hear its a good addition. I'm setting up a test environment as i'm typing this. I'v got some problems getting the environment to work..

samhed:
Apologies for the slow response, it would be nice if you could squash at least the unit test-commits :)

DirectXMan12:
@samhed @CendioOssman how does this fit with #653?  Can you cherry-pick this into your changes (keeping commit attribution, etc)?

samhed:
Instead of having a key to lookup the values, could we instead use the full English string? Makes the code much easier to read plus that you don't have to look in a different file to see the actual string.

CendioOssman:
> Instead of having a key to lookup the values, could we instead use the full English string? Makes the code much easier to read plus that you don't have to look in a different file to see the actual string.

+1
This is how it is commonly done elsewhere, usually using gettext.

Also, note that there are strings in the HTML as well that would need translation.

kanaka:
+1 for localization. Also, +1 for using the full English strings as the lookup. One additional advantage of that is that if the key is used for the default translations that even the translation is missing for certain text elements (which is pretty much guaranteed when new texts are first added), then the user gets the full text (albeit untranslated) rather than just a internal token.

I would also request that the localization function is aliased to `_` in the modules that use it. This is pretty standard practice and avoids modifying the general appearance of the code with verbose function calls for every piece of UI text.

edfink:
+1 for interest. I'd really like to have these strings translated as well.

A quick look at the change has me wondering about string substitution instead of concatenation in a few places.

DirectXMan12:
> I would also request that the localization function is aliased to _ in the modules that use it. This is pretty standard practice and avoids modifying the general appearance of the code with verbose function calls for every piece of UI text.

That's probably not the best idea, unless we take pains to make sure it's scoped well, otherwise it'll conflict with underscore.js, etc.

kanaka:
@DirectXMan12 yes, sorry, I wasn't clear enough. I was intending to suggest a scoped variable i.e. in RFB object a local variable alias `_` is created for `Util.Localisation.get`.

DirectXMan12:
> I was intending to suggest a scoped variable i.e. in RFB object a local variable alias `_` is created for `Util.Localisation.get`.

ah, ok, that's fine :+1:

LoekJanssen:
I created a fork of the latest version of noVNC and merged the localization changes with that version because of the amount of conflicts and changes. It's an option to close this pull request, I will create a new one soon. (once all translations are finished)

samhed:
@LoekJanssen very nice to hear!! Any chance you could also have a look at using the full english string for lookup?

You don't have to create a new PR, you can just rebase and push --force this one.

CendioOssman:
Have you looked at Jed rather than inventing your own thing? It seems to be the popular alternative to gettext in the Javascript world.

LoekJanssen:
I've used full English strings for lookup.

Thanks for bringing Jed to my attention, I didn't know about it. So I haven't used Jed for noVNC, but I think the current version is straightforward and lightweight already. I will update the code soon so you can have a look.

LoekJanssen:
I accidentally removed the branch connected to this pull request...
I've got a new branch merged with the latest changes ready. Can I create a new pull request?

CendioOssman:
Have you tried creating a new branch with the same name? That might resolve things.

LoekJanssen:
@CendioOssman mine is just called noVNC. The same as before.

CendioOssman:
Oh, you mean you removed the entire repository, not just the branch?

LoekJanssen:
Ah yes, that's what I meant

LooksgoodNL:
I've created a new clean pull request. This one is now old and can be closed. New: #701

PR #558, opened 2015-12-02
Push image to the render queue when it is loaded

adambabik:
When running noVNC in Firefox or Safari, I noticed some glitches on the canvas from time to time. It turned out that the received image is drawn before it's loaded. Even though, `img.src` is a base64 string, it does not guarantee that the image is immediately available.

I added an `img.onload` callback which pushes the image to the queue instead of pushing it immediately.

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -1958,14 +1958,16 @@ var RFB;
                     // We have everything, render it
                     this._sock.rQskipBytes(1 + cl_header);  // shift off clt + compact length
                     var img = new Image();
-                    img.src = "data: image/" + cmode +
+                    img.onload = function (img, x, y) {
+                        this._display.renderQ_push({
+                            'type': 'img',
+                            'img': img,
+                            'x': x,
+                            'y': y
+                        });
+                    }.bind(this, img, this._FBU.x, this._FBU.y);
+                    img.src = "data:image/" + cmode +
                         RFB.extract_data_uri(this._sock.rQshiftBytes(cl_data));
-                    this._display.renderQ_push({
-                        'type': 'img',
-                        'img': img,
-                        'x': this._FBU.x,
-                        'y': this._FBU.y
-                    });
                     img = null;
                     break;
                 case "filter":


DirectXMan12:
So, the code in this PR could lead to an out-of-order render queue, which would induce graphical artifacts.

What's supposed to happen here is that the check in `_scan_renderQ` makes sure that an image is not rendered before it's loaded (see https://github.com/kanaka/noVNC/blob/master/include/display.js#L750-L758).  If that check is not working properly, it needs to be fixed there.  Do you have a consistent reproduction case?

dreame4:
Thanks for checking on that! I'll keep you posted when I have a chance to verify it, but I definitely agree that the fix should not be in the place I did it.

dreame4:
@DirectXMan12 I narrowed the problem a little bit. Here is an example how these glitches look like:
screen shot 2015-12-08 at 13 29 29

The problem is visible only in a slow network connection with high load, for instance on youtube.com.

From the logs, I found out that there are many frames accumulated in the `_renderQ` before they're drawn. My solution (imperfect in a general case, but working fine with images in base64) adds an image to the queue and draws it immediately as it's already loaded. 

However, I still don't know what is the root cause of this problem. The x and y coordinates are cached so they have proper values and the image is definitely valid and loaded...

dreame4:
I think the problem is that my VNC server sends both `jpeg` and `copy` frames. I assume that from time to time, a `copy` frame is handled before a `jpeg` frame and the glitches appear. To verify that, I turned off handling of `copy` frames and my problem disappeared. 

So, it looks like the root cause of the problem is that `jpeg` frames might be handled asynchronously while `copy` synchronously. Thus, the order of the received frames is not preserved.

DirectXMan12:
@dreame4 hmm... that may be a simple fix.  I'll take a peek.  Thanks for tracking it down!

dreame4:
@DirectXMan12 It's "COPYRECT".

samhed:
How does these changes go together with the ones in #668 ?

samhed:
This seems to be done in #668 on commit c483795ca32c61bb4a265308bcb140fc6971cb2b

Closing this one

PR #564, opened 2015-12-16
fix negotiation fail when empty sub-auth types received

avost:
When a TightVNC server configured to allow no-auth connection, it may sends (the recent TightVNC server actually sends) an empty 'sub-auth' capabilities list at tight security type sub-auth negotiation phase.  In this case noVNC viewer fails  on negotiation with ''Unsupported tiny auth scheme:" message.
Fix this to assume 'no-auth' authentication subtype selected when received an empty auth capabilities list.

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -795,6 +795,11 @@ var RFB;
             // second pass, do the sub-auth negotiation
             if (this._sock.rQwait("sub auth count", 4)) { return false; }
             var subAuthCount = this._sock.rQshift32();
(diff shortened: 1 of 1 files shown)


DirectXMan12:
For future reference, the RFB proto says this about TIGHT auth:

> If number-of-auth-types is non-zero, the client has to request an authentication type from the list with an authentication scheme request
> ...
> If number-of-auth-types is zero, the protocol the proceeds directly at security type None.

(https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#723tight-security-type), which matches the change in this PR.

PR #565, opened 2015-12-22
Resize Receive Queue to Fit Incoming Messages

### LINKED ISSUE #557: RangeError with Raw encoding at websock.js
I have a custom app that communicates via RFB protocol. It use noVNC to display UI. Using Hextile encoding is ok, switching to Raw periodically produce the following exceptions:

```
recv_message, caught exception: 
    name: RangeError
    message: Source is too large
RangeError: Source is too large
    at Uint8Array.set (native)
    at Object.Websock._decode_message (http://localhost:8080/widget/include/websock.js:346:22)
    at Object.Websock._recv_message (http://localhost:8080/widget/include/websock.js:352:22)
```
I'm not seeing errors in Raw encoding at my app side, but they can be there) However I think anyway noVNC should deal with this gracefully.
Could someone give me a clue why it happens?

DirectXMan12:
This PR causes the receive queue to be resized in order to fit incoming messages.  There is a hard cap of a size of 40 MiB, so that we can't consume too much memory by accident.

Fixes #557

### DIFF
--- a/include/websock.js
+++ b/include/websock.js
@@ -67,6 +67,11 @@ function Websock() {
 (function () {
     "use strict";
 
+    // this has performance issues in some versions Chromium, and
+    // doesn't gain a tremendous amount of performance increase in Firefox
+    // at the moment.  It may be valuable to turn it on in the future.
+    var ENABLE_COPYWITHIN = false;
+
     var typedArrayToString = (function () {
         // This is only for PhantomJS, which doesn't like apply-ing
         // with Typed Arrays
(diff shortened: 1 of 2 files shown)


DirectXMan12:
@kanaka @samhed look good to you?

kanaka:
@DirectXMan12 looks good. Should we add a test for this function?

DirectXMan12:
> Should we add a test for this function.

Yep, good catch.  Will do.

DirectXMan12:
@kanaka test added.  I'll merge shortly unless anyone has any objections.

kanaka:
@DirectXMan12 no objections here.

samhed:
LGTM!

Issue #559, opened 2015-12-04
Delays for non-JPEG mode caused by GC at regular intervals


samhed:
We have been looking at the performance of the new zlib code among other things when we noticed this new behavior.

I disabled JPEG by running the following in the web console:

    UI.rfb._encodings.splice(8,1)
    RFB.messages.clientEncodings(UI.rfb._sock, UI.rfb._encodings, true, true)

The garbage collector kicks in consistently at an interval of ~7 seconds and this causes a delay of about 50-150ms each time.

The old code only has these incidents sporadically. I have been trying to figure out why, but I cant find what is causing this.. Any ideas @DirectXMan12 ?

samhed:
The graph in Firefox's performance measurement tool illustrates this quite well.

![screenshot from 2015-12-04 16 39 16](https://cloud.githubusercontent.com/assets/4117162/11593843/d116f23e-9aa6-11e5-9f2f-518357cbee14.png)

Playing a video in the session is a good way to make the delays noticeable.

DirectXMan12:
hmm... I'll try and take a look next week.  This only happens when JPEG is disabled?  It's possible the server is sending one of the other encodings instead, and there's a bug in one of those.

samhed:
> This only happens when JPEG is disabled?

yes

DirectXMan12:
I think I've been able to reproduce what you're seeing (using TigerVNC) and I've identified a couple of potential improvements.  I still get GC pauses, but they're fewer and farther between.  I'll try and post a branch with the fixes soon so that you can test them out and see if things improve for you as well.

samhed:
that sounds great!

DirectXMan12:
Sorry about the delay -- I've posted a couple of changes which seem to reduce GC pauses for me.  They need some tweaking before being production-ready, but I wanted to see if they help you out at all: kanaka/novnc#566.  PTAL and let me know if they help at all.

samhed:
In fact I can't reproduce these problems which I had in the first place. According to the yum logs on the machine on which I found this, it had Firefox 38 at the time, which also seems to be correct judging from the look of the "performance"-tool in the screenshot. But with the old code I can't reproduce the issue with Firefox 37. 38, 40, 41, 42, 43 or 44. Firefox 39 crashed when I tried to measure the performance.

I don't have access anymore to the specific machine on which I found this problem originally. Perhaps different graphics cards or an update to some library which Firefox uses is the cause of this. I don't know.

[...]

PR #566, opened 2015-12-22
[WIP] Reduce GC Spikes in Firefox w/o JPEG mode

DirectXMan12:
See #559


DirectXMan12:
@samhed PTAL when you get a chance

samhed:
On it.

samhed:
I will continue with this on Monday.

samhed:
I have tested this today, I am not seeing any problems. But see #559

samhed:
I find it difficult to see if we want these changes even though the problem ( #559 ) isn't possible to reproduce anymore?

If we want them, we have to at least remove the comments from the code and fix the unit tests.

PR #570, opened 2016-01-23
Map Util.[Debug|Info|Warn|Error] to console.[debug|info|warn|error].

tghw:
This change will preserve file and line number reporting when Util.[Debug|Info|Warn|Error] is called, which should make development and debugging a lot easier.

See http://stackoverflow.com/questions/13815640/a-proper-wrapper-for-console-log-with-correct-line-number for an explanation of the binding.

### DIFF
--- a/include/util.js
+++ b/include/util.js
@@ -211,13 +211,13 @@ Util.init_logging = function (level) {
     /* jshint -W086 */
     switch (level) {
         case 'debug':
-            Util.Debug = function (msg) { console.log(msg); };
+            Util.Debug = console.debug.bind(window.console);
         case 'info':
-            Util.Info  = function (msg) { console.log(msg); };
+            Util.Info  = console.info.bind(window.console);
         case 'warn':
-            Util.Warn  = function (msg) { console.warn(msg); };
+            Util.Warn  = console.warn.bind(window.console);
         case 'error':
-            Util.Error = function (msg) { console.error(msg); };
+            Util.Error = console.error.bind(window.console);
         case 'none':
             break;
         default:


tghw:
Hmm... Looks like the checks failed because PhantomJS did not implement `Function.prototype.bind` until PhantomJS 2.0: https://github.com/ariya/phantomjs/issues/10522#issuecomment-72705658

DirectXMan12:
I've actually been playing around with this in #496.  I didn't get enough time to finish testing the potential performance impact of using `bind` (some sources suggest that bound functions have a performance impact).  If you could check some rough numbers, that would be great.

tghw:
I set up a jsPerf to test the performance of bound methods: http://jsperf.com/console-log-vs-console-log-bind

The results are very mixed.

On Chrome, the unbound version is about 10% faster than the bound version. (But only if a single argument is passed in, which is the way noVNV logging is set up. It's about equal if you just pass `arguments`.) With dev tools closed, it runs about 100,000 ops/sec. With them open, it's about 10,000 ops/sec.

On Firefox, the two are almost identical. It's about 30,000 ops/sec with the dev tools closed and about 6,000 ops/sec with them open.

On IE, when the Developer Tools are not open, `console.log` is just a no-op, leading to 5,000,000 ops / sec. I believe bound is faster here only because it doesn't have to do that extra function call to the no-op. But when you open the dev tools, it drops to 10,000-15,000. No real difference between bound and unbound.

Safari is similar, console.log seems to be a no-op unless the developer tools are open, leading to about 500,000 ops/sec for both. When the dev tools are open, it drops to about 125,000 ops/sec, with unbound being 5-10% faster.

Overall, it looks like binding is, at worst, about 10% slower, with the slowest browser tested able to do over 6,000 log messages per second with the dev tools open. Considering this, I don't think performance will be an issue.

DirectXMan12:
Alright, sounds good.  I'll merge this shortly once I figure out a good way to get PhantomJS to work.

kanaka:
On Linux the bound version has significantly worse performance with Chrome 48. 45,820 ops for unbound, 36,369 for bound (over 20% slower). That's pretty significant slow down.

DirectXMan12:
> On Linux the bound version has significantly worse performance with Chrome 48: 45,820 ops for unbound, 36,369 for bound (over 20% slower). Firefox 39 on Linux is about the same.

@kanaka I tested in vnc_playback a bit myself (just a bit) and didn't see as much of a slowdown as you're describing, but I wanted to do some more checking. 

> I'm not sure that's worth it for accurate line numbers

I'd say that much a of a slowdown isn't worth it, but a smaller slowdown might be (accurate error reporting would be *really* nice).

kanaka:
@DirectXMan12 yeah, the microbenchmark is just doing the logging and nothing else. If vnc_playback shows very small performance difference, then it's probably worth it. There isn't actually that many logging calls in the fast path now that I think about it (I had a bunch that I left commented out most of the time).

Another option that would allow us to have a small script that could rewrite rfb.js and display.js to uncomment/comment the fast-path debug logging (for those times when we want extremely verbose logging, but don't want to pay the cost during normal runtime). Something like:

```javascript
    // Util.Debug("FBU details: ..."); // FASTPATH DEBUG, see util/toggle_debug.sh
```

The `util/toggle_debug.sh` could then math on "FASTPATH DEBUG" and remove/add the comment at the beginning of the line. Just a thought.

CendioOssman:
So I tested this with Firefox 50 and Chrome 55 on Fedora 25. I could not see a meaningful performance difference between the two versions. The variation between test runs was larger than any variations between the two versions.

This is the test code used:

```JavaScript
document.body.innerHTML = "Running...";

//var logger = function (msg) { console.log(msg); };
var logger = console.info.bind(window.console);

var start = window.performance.now();
var count = 0;
while (start + 5000 > window.performance.now()) {
  logger('test');
  count++;
}

document.body.innerHTML = count / 5 + ' messages / second';
```

I tested the same thing using `vnc_playback.html`, modifying it with an extra `Util.init_logging('debug');`. Same result there, no meaningful difference in the numbers.

So I'm going to go ahead and merge this. :)

PR #581, opened 2016-02-17
Fix bug that copying image buffer to ImageData fails in IE10

pjknkda:
IE10 and older versions of IE does not support `Uint8ClampedArray`, so we need to copy buffer image manually via for-loop.

### DIFF
--- a/include/display.js
+++ b/include/display.js
@@ -15,12 +15,22 @@ var Display;
 (function () {
     "use strict";
 
-    var SUPPORTS_IMAGEDATA_CONSTRUCTOR = false;
+    var SUPPORTS_UINT8CLAMPEDARRAY = false;
     try {
-        new ImageData(new Uint8ClampedArray(1), 1, 1);
-        SUPPORTS_IMAGEDATA_CONSTRUCTOR = true;
+        new Uint8ClampedArray(1);
+        SUPPORTS_UINT8CLAMPEDARRAY = true;
     } catch (ex) {
         // ignore failure
+    }
+
+    var SUPPORTS_IMAGEDATA_CONSTRUCTOR = false;
+    if (SUPPORTS_UINT8CLAMPEDARRAY) {
+        try {
+            new ImageData(new Uint8ClampedArray(1), 1, 1);
+            SUPPORTS_IMAGEDATA_CONSTRUCTOR = true;
+        } catch (ex) {
+            // ignore failure
+        }
     }
 
     Display = function (defaults) {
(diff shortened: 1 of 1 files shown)


samhed:
I tested your code in IE 10 on a Windows 7 machine, can't find any obvious issues. However, in a way, we took a decision in #488 that we were OK with not supporting IE10 anymore. What do you think @kanaka , @DirectXMan12 and @astrand ?

DirectXMan12:
I'm not sure about this...  What will our policy be about other IE10 bugs?  Should we care about them?  I'm not totally against this fix, I just think we should figure out our policy so that we don't end up with partial IE10 support that clutters up the code but doesn't make noVNC work 100% with IE10 (also, I'm not overly concerned with Windows Server 2012 lacking IE11, since it's a server OS, and thus possibly the target of noVNC sessions, but probably not the source of the originating browser).

astrand:
We could decide on a "best effort" policy for IE10 and similar browsers, where we do not invest a lot of time in coding and testing ourselves, but if someone contributes an unintrusive patch and otherwise fine patch, it can be accepted. It is true that noVNC would probably not work 100% with IE10 most of the time, but it could still have some value for some users. After all, those users that wants IE10 can grab snapshots that are known to work with IE10. 

Wrt Windows Server 2012, it could be a target for noVNC in the case of terminal server deployments (Remote Desktop Services), as well as the case with an admin running noVNC based systems for administration purposes. One example of this could be connecting to KVM systems from a Windows 2012 console. 

Also, I would not be surprised if IE10 is still used in many embedded systems such as thin clients etc.

kanaka:
While the change is fairly small and unobtrusive, it still adds new code in the hot path. And this is the type of code that can occasionally have a negative affect on performance because it can interfere with the type inference engine in the JIT (the "img" variable now has multiple types that it can possibly be depending on the conditional). So, if we accept this, we need to test performance across multiple browsers to make sure there is no performance impact. My opinion is that we should not sacrifice any performance for IE10 support given the tiny markshare. However, if somebody on the team (which I think means @DirectXMan12 at this point) is willing to do performance testing for this type of request (or the submitter can give sufficient evidence of perf testing), then I'm fine with a "best effort" for browser versions with small market share. It also has a future maintenance and refactoring cost, so at the very least, this section of code should be clearly marked with a comment indicating why this was added and when it can go away.

I would probably give a tentative thumbs up to this patch assuming it doesn't have perf impact and the author adds an appropriate comment.

nierChen:
it is very useful,it just fixes my problem,thanks so much!

CendioOssman:
Another year and usage has fallen even further. Is this still worthwhile?

kanaka:
We could go with a compromise by adding this to a branch, mentioning it somewhere, but being clear that it's not officially supported. I agree that this is probably not worth supporting going forward in master.

samhed:
@kanaka +1

DirectXMan12:
I think having a separate branch is more work that is worth.  IMO, our official policy should probably be "use noVNC 0.6.2 if you want IE10 support".

kanaka:
@DirectXMan12 I'm fine with that. If somebody really wants to take this on they can fork and maintain this support themselves (and we could link to their fork if they so desired).

samhed:
I'm going to go ahead and close this one then.

PR #582, opened 2016-02-24
Make sure to copy arrays when using render queue

### LINKED ISSUE #571: Rendering artifacts in firefox
Setup:
- Browser: Firefox 43 (both @ Windows / Linux), current Chromium and Opera seem not to be affected
- TightVNC Server for Windows 2.7.10, same issue with UltraVNC Server

The origin of the issue seems to be commit abf2b09. I connected three different versions of the noVNC client simultaneously and moved the foreground window around. While a825582 behaves properly, a369a80<sup>1</sup> and 670dbdd<sup>2</sup> show rendering artifacts:

<sup>1</sup> *some commits after abf2b09 since while doing the test, abf2b09 lost connection shortly after being established.*
<sup>2</sup> *current head@master*

novnc-bug

DirectXMan12:
This commit ensures that input arrays are copied to new storage when
they are pushed onto the render queue.  This ensures that they are
not overwritten before they are eventually used.

Fixes #571

### DIFF
--- a/include/display.js
+++ b/include/display.js
@@ -471,9 +471,14 @@ var Display;
 
         blitImage: function (x, y, width, height, arr, offset, from_queue) {
             if (this._renderQ.length !== 0 && !from_queue) {
+                // NB(directxman12): it's technically more performant here to use preallocated arrays,
+                // but it's a lot of extra work for not a lot of payoff -- if we're using the render queue,
+                // this probably isn't getting called *nearly* as much
+                var new_arr = new Uint8Array(width * height * 4);
+                new_arr.set(new Uint8Array(arr.buffer, 0, new_arr.length));
                 this.renderQ_push({
                     'type': 'blit',
-                    'data': arr,
+                    'data': new_arr,
                     'x': x,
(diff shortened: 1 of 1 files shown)


DirectXMan12:
cc @kanaka @cavac

kanaka:
Minor typo in comment "but it but it's".

Probably a good idea to do a quick performance comparison just in case. Otherwise it LGTM.

DirectXMan12:
>  Probably a good idea to do a quick performance comparison just in case

@kanaka very little AFAICT -- basically, if you're using the render queue, you're probably not using the methods requiring arrays of image data as much (since the bulk of your frames are probably using TIGHT PNG or TIGHT JPEG).  If you're not using the render queue, there's no change, so no impact.

Additionally, there's not really much of an alternative -- one of the other functions is already doing this, and unless we want to add in a `malloc` implementation (which I've been trying to avoid for the most part), this is not going to go away.

DirectXMan12:
Fixed the typo.  Will merge as soon as we get Travis approval.

PR #585, opened 2016-03-11
Added support for RealVNC 5.3

larsolevard:


DirectXMan12:
Hi!  Thanks for this contribution!  Could you please add a bit more detail to the commit message and the PR message (if you have a link to where this is specified, that would be great as well?  Otherwise, the LGTM :+1:

larsolevard:
Added support for RealVNC 5.3: RFB max version is 3.8, but RealVNC 5.3 sends "005.000" as the version string since they've extended their implementation upon RFB v3.8. Since RealVNC 5.3 implementss RFB v3.8 the noVNC client can interpret the "005.000" version string as RFB v3.8. The patch is tested against a fresh install of RealVNC Server 5.3 running on Windows 7 x64.

larsolevard:
I amended the commit message to be the same as my previous comment and force pushed it

DirectXMan12:
Closing in favor of #661 (I just rebased and cleaned up the commit message).  Thanks!

PR #597, opened 2016-04-11
Handle missing leading slash in elem.pathname

### LINKED ISSUE #591: IE11 problem
There seems to be a strange issue on IE11 that I've uncovered.  In noVNC/include/webutil.js, The return statement for webUtil.injectParamIfMissing is "return elem.pathname.slice(1) + elem.search + elem.hash;"

In my use case, the path there is something like "/shell/etc..."  In IE11, the return value looks like "hell/etc...".  This results in an eventual 404 when trying to start the websocket connection, and a 1006 error presented to the user.  IE versions prior to 10 are known to leave off the leading slash.  When that slash is absent, the slice function is actually trimming the first letter of the path.  It's not clear why IE11 is misbehaving in this way, but I've reproduced it on sever machines in the last few days.

A hackish fix would be to change that return statement so it first checks for the leading slash:

if (elem.pathname.charAt(0) == "/") {
        return elem.pathname.slice(1) + elem.search + elem.hash;
    } else {
        return elem.pathname + elem.search + elem.hash;
    }

If that seems an acceptable fix I can send a pull request.
--Jonathan Bennett

oneru:
IE11 with compat mode turned off has been observed displaying this behavior.  This commit checks for and corrects this broken behavior.

Fixes #591

### DIFF
--- a/include/webutil.js
+++ b/include/webutil.js
@@ -282,5 +282,10 @@ WebUtil.injectParamIfMissing = function (path, param, value) {
         elem.search = "?" + query.join("&");
     }
 
-    return elem.pathname.slice(1) + elem.search + elem.hash;
+    // some browsers (e.g. IE11) may occasionally omit the leading slash
+    // in the elem.pathname string. Handle that case gracefully.
+    if (elem.pathname.charAt(0) == "/") {
+        return elem.pathname.slice(1) + elem.search + elem.hash;
+    } else {
+        return elem.pathname + elem.search + elem.hash;
 };


DirectXMan12:
LGTM :+1:

Nit: If you wouldn't mind, could you add a space between `//` and the comment text, and also change the comment to read something like "some browsers (e.g. IE11)" so it's easier to see which browsers we added this in for?

oneru:
I fully understand the nitpick.  I amended the commit, should be good now. Thanks

DirectXMan12:
Thanks!

Issue #599, opened 2016-04-13
The control bar can't be hidden


jp-bennett:
Yet another little issue, the fullscreen button doesn't truly fullscreen the vnc session.  It would be great if fullscreen from a desktop browser would hide the control bar and allow for true fullscreen.  Pushing the mouse all the way to the top-middle of the screen should make the control bar visible again, similar to how auto-hide works for the task bar on a desktop OS.

I hope to work on this myself.  A quick test in google Chrome's page inspector makes me think it should be doable.

samhed:
An auto-hide control bar would be desirable even outside of fullscreen

oneru:
Hiding the control bar is the easy part.  The challenge is how to handle making it visible again, and how to do that in a way that is intuitive to the user.  My thought was to make a "hot spot" in the top center of the screen that would make it visible again.  This would probably involve adding another callback, if not two, from rfb.js to ui.js.

Also, the hotspot idea doesn't work as well outside of fullscreen.  Mousing up and hitting a couple pixel tall hotspot is easy with the screen edge preventing overshoots, but would be nearly impossible otherwise.  Also, hotspot wouldn't work well on mobile.

One approach could be reserving a shortcut key to show the control bar, or to show a menu with control bar options.  Some native vnc clients use f8 for this purpose, I believe.

Another option would be to add a "leave fullscreen" or "Show control bar" button floating over the canvas at the top middle of the screen, not stealing from total canvas size, but still available to click/tap.

As I was playing with the idea last night, though, I found what might be a better option.  When going fullscreen, hide the control bar.  The browser (All of them that I tested, at least) displayed a notice that the tab went fullscreen, and to hit escape to exit (On mobile, you swipe down from the top of the screen and hit the back button to leave fullscreen).  So we hook the onresize function in ui.js, and if no longer fullscreen, make sure that the control bar is visible again.

I'll open a PR with my code, but it almost certainly isn't ready to pull in its current state.

samhed:
As I wrote in #601 I don't like the idea of never showing the control bar in fullscreen. But I still like the concept of having a way to hide the control bar.

Auto-hide combined with a movable floating button along the edge of the canvas is in my mind the best solution for showing the control bar.

oneru:
I agree with your comments, and hope to finish developing this feature so that it can be pulled.  I will pursue the idea of the floating button, I think it has potential to be the best solution, and could work on mobile as well.

samhed:
[...]

PR #601, opened 2016-04-13
Hides the control bar in fullscreen mode

oneru:
As discussed in #599, This patch is my implementation of hiding the status bar in fullscreen mode.

It might be better eventually to add a variable to track the status bar, and then add a function to toggle status bar visibility.


samhed:
This is not a good change in my mind. Without any intuitive way of showing the control bar again (such as a hot spot or a hotkey) I don't want to merge this.

Furthermore, the function 'updateFullscreenButton' is not a good place for this code.

PR #603, opened 2016-04-14
Added option for --ssl-only noVNC session

WardF:
Added an option. `--ssl-only`, to the `launch.sh` script.  This option is passed through to the underlying `websockify` executable.  The end result is a `noVNC` session which rejects `http` connections but will accept `https` connections, assuming the other preconditions for `https` have been met.

I apologize for the lack of a unit test to accompany this change, but I'm not a web developer and wasn't really sure how to go about adding one!

### DIFF
--- a/utils/launch.sh
+++ b/utils/launch.sh
@@ -5,7 +5,7 @@ usage() {
         echo "$*"
         echo
     fi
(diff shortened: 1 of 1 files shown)


WardF:
Fantastic, thanks!

PR #608, opened 2016-05-01
Enabling mouse use in chrome along touch.

zarmhast:
Tested in 50.0.2661.94 m (64-bit) on Windows. Loaded noVNC from master (0.5.1)
During the test I tried using the **mouse** to interact with various window elements of a remote machine: *success*.
Interact with various window elements using **touchscreen**: *success*.
My guess is that the following are not supported so these failed to even establish a VNC connection using:
- MS Edge
- Android Chrome
- Android Firefox
- MS IE 11
Hence touch and mouse tests failed on these by default with patch and without.

### DIFF
--- a/include/input.js
+++ b/include/input.js
@@ -332,19 +332,12 @@ var Keyboard, Mouse;
         grab: function () {
             var c = this._target;
 
-            if ('ontouchstart' in document.documentElement) {
-                Util.addEvent(c, 'touchstart', this._eventHandlers.mousedown);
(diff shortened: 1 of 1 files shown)


samhed:
> My guess is that the following are not supported so these failed to even establish a VNC connection using:
>
> MS Edge
> Android Chrome
> Android Firefox
> MS IE 11 Hence touch and mouse tests failed on these by default with patch and without.

All of these browsers should most definitely work fine with noVNC.

This solution doesn't work in Chrome or Firefox on Android or Safari or Chrome on iOS.

PR #611, opened 2016-05-08
Adding external event handler for WebSocket connection events

ayonghosh:
Adding external event handler for WebSocket connection events (open, close and error)


ayonghosh:
Does it make sense to merge this in?

DirectXMan12:
I apologize for the long review time.  I'm not sure what the motivation for this is.  If you're still interested in the PR, please reopen and provide a motivation.

ayonghosh:
Currently, noVNC does not offer a way for external handlers on WebSocket errors. The default behavior involves an error on the console but there is no way to communicate this to the user. This is a problem we faced at one of our projects in my current organization and that is the motivation for this PR.

Issue #510, opened 2015-07-15
Detecting touchscreen causes mouse not to work


RickGlimmer:
On a Surface Pro 3 which has a touchscreen it is not possible to use the mouse.
I think the code in input.js where a feature detect is done for touchstart only attaches the touch events and not the mouse.
It might work better when using  pointer events as they are now considered the best way of handling all kinds of pointer input devices.
I do not have a test environment otherwise i would have a go at it.

DirectXMan12:
See also #369 and #476.  Unfortunately, I don't have a Windows-based touch device to test this with.  If you do, or you know someone who does, we'd be glad to accept a PR

zarmhast:
Please see PR #608 . I have windows and the patch was tested together with the current master branch.
Both touchscreen and mouse worked without any issues in chrome. Please let me know if this causes issues with touchscreen browsers in Linux/OSX (would not expect it to).

Thank you.

samhed:
Your code does not work on Android and iOS.

zarmhast:
@samhed Hello, thank you for testing my changes. Since I do not own any iOS devices, I would really appreciate if you would elaborate about what exactly does not work?
After a few attempts noVNC in chrome on Android was finally able to establish a VNC connection and I was able to interact with windows on a remote machine by simply touching the screen wherever I needed to.

samhed:
> After a few attempts noVNC in chrome on Android was finally able to establish a VNC connection and I was able to interact with windows on a remote machine by simply touching the screen wherever I needed to.

The mousedown event isn't triggered when you start touching the screen, it's sent together with mouseup when you lift your finger. Yes, clicks go through, but it's impossible to do drag-and-drop or mouse selection rectangles for example. It's also impossible to use panning.

zarmhast:
> Yes, clicks go through, but it's impossible to do drag-and-drop or mouse selection rectangles for example. It's also impossible to use panning.
@samhed thank you, that is helpful. Now I have a better idea of what needs to be fixed.

zarmhast:
This time I know that the issue is fixed with pull request #619. @samhed, thank you for your input, it helped.

samhed:
Fixed by 4b20f2364c3924e89ccedf7eda56af322e5e9e1e

PR #613, opened 2016-05-13
Allow touchmode-interface for Microsoft Surface tablets

samhed:
We should use navigator.maxTouchPoints since 'ontouchstart' isn't enough to detect touch on MS Surface devices. This had the effect that the touch controls in the UI didn't show on these devices.

### DIFF
--- a/include/ui.js
+++ b/include/ui.js
@@ -53,7 +53,10 @@ var UI;
 
         // Render default UI and initialize settings menu
         start: function(callback) {
-            UI.isTouchDevice = 'ontouchstart' in document.documentElement;
+            UI.isTouchDevice = (('ontouchstart' in window)
+                                // required for MS Surface
+                                || (navigator.maxTouchPoints > 0)
+                                || (navigator.msMaxTouchPoints > 0));
 
             // Stylesheet selection dropdown
             var sheet = WebUtil.selectStylesheet();


zarmhast:
@samhed,
This bug is not unique to Microsoft Surface, it is unique to MS IE and Edge on touch capable devices.
The solution contained in this pull request is not complete and will yild the same problem that you stumbled when testing my pull request several weeks ago. (see #510)
My suggestion is to add
`Util.IsTouch=function(){ /*your solution here*/};`
then utilize this method in both ui.js and input.js for touch capability detection.

samhed:
> @samhed,
This solution though will display the touch related buttons in the toolbar it will not enable proper touch interactivity in MS Edge.

Yeah, I'm quite aware that this would only solve part of the problem. The fact still stands that the UI-detection for touch is incomplete. I believe that we should display the touch controls in the UI for any touch device, even if the device has access to an ordinary mouse and keyboard.

> I've tested master branch on a touch capable laptop with Windows 10 and MS Edge. It all worked once I changed the default setting to allow touch events, then the buttons appeared on their own.

> Navigate to about:flags in MS Edge in order to change the default "Always off" to anything else. Once that is done master works with touch events without a problem.

Expecting the users to change settings is not a good "solution" in my mind.

> The only issue on master right now is that once that is done, then mouse events are disabled. One of the fixes for that can be #619 (which, this time, was properly tested on touch devices, including an android tablet for clicking and dragging).

In order to properly support noVNC on MS Surface tablets for instance, we would need both the changes from this PR and the changes from your #619

zarmhast:
> Expecting the users to change settings is not a good "solution" in my mind.
Agree, but without that option change MS Edge pretends to not have a touchscreen. Meaning that even though the touch interface will be displayed, it will have no effect since only "mouse" messages will be coming through.
I have taken out the whole condition for touch events locally and put in a console.log statement in the mousebuttonhandle. With the option at its default setting, even though touchstart was registered, only MouseEvents where received.
So in the end it is not a problem with the code as much as with the browser. Btw, there is the same thing with Firefox on desktops (both mobile ie and firefox work fine with touch).
Maybe include a warning: we've detected that your device has a touchscreen but its support is disabled for this browser, click here for instructions how to enable touch support.
Or something like that.
What do you think?

zarmhast:
Actually it is even worse than I thought: with touchscreen support disabled, attempting to double tap engages some weird zoom mode that does not display as zoom in the menu, but still makes elements bigger and messes with the coordinate translation in the Mouse object.

In case you do not have a windows touch device we could set up a call and I'll share my screen with you to show you what is going on, if you'd like. Having more experience with noVNC, you are more likely to understand what is going on.

samhed:
You are confusing me here.. Didn't you say that #619 is a fix for sending touch events correctly, even on MS Edge?

#619 only concerns mouse events sent to the server. This PR only concerns user interface. The goal should be to fix both these parts. 

The functions that are specific to touch devices in the UI are:
* Forced disabled local cursor setting
* Forced clipping mode setting
* Mouse selector button
* Panning button
* Virtual keyboard button
* Extra key buttons (Ctrl, Alt, Tab, Esc)

zarmhast:
>You are confusing me here.. Didn't you say that #619 is a fix for sending touch events correctly, even on MS Edge?

I might've claimed something like that in one of my edited messages, but that was before i read up more about touch events, browser support and have actually tried on an edge browser. My tests were more concerned with testing on a variety of devices than a variety of browsers and hence did not notice that edge does not process touches. (I don't have Firefox installed on my touch device, btw.)

>#619 only concerns mouse events sent to the server. 

PR #619 is concerned with mouse events being captured, proceeded on client side without interfering with touch event handling. So that dual input devices can use touch and mouse if touch is supported by both the browser and the device. In other words: fix mouse control without screwing touch.

@samhed, sorry for confusing you, hope that this reply will introduce more clarity. Also, please let me know if my awkward attempts to be helpful are not helping.

samhed:
Seeing as this PR will still be useful alongside #619 and that this discussion mostly regards #619 we should move the discussion to that PR.

samhed:
Closing in favor for #650

samhed:
> Actually it is even worse than I thought: with touchscreen support disabled, attempting to double tap engages some weird zoom mode that does not display as zoom in the menu, but still makes elements bigger and messes with the coordinate translation in the Mouse object.

> In case you do not have a windows touch device we could set up a call and I'll share my screen with you to show you what is going on, if you'd like. Having more experience with noVNC, you are more likely to understand what is going on.

> P.S.: a few double taps later the canvas area disappeared, this is what I saw in the dom explorer:
<canvas width="-676" height="462" id="noVNC_canvas" style="width: 144px; height: 462px; cursor: none;">Canvas not supported.</canvas>

Fixed in d6cb04a

PR #619, opened 2016-06-05
Fix for touchscreen devices that also have a mouse.

zarmhast:
Change was tested on:
- touchscreen laptop (touchscreen and mouse for dragging objects, left/right clicking)
- android tablet (was able to drag objects, right click, pan around).

This fix relies on the fact that all touch events once captured are preventing from bubbling hence eliminating the danger of having onmousedown fire right after ontouchstart.

This should solve issue #510 completely.

### DIFF
--- a/include/input.js
+++ b/include/input.js
@@ -338,14 +338,13 @@ var Keyboard, Mouse;
                 Util.addEvent(window, 'touchend', this._eventHandlers.mouseup);
                 Util.addEvent(c, 'touchend', this._eventHandlers.mouseup);
                 Util.addEvent(c, 'touchmove', this._eventHandlers.mousemove);
-            } else {
-                Util.addEvent(c, 'mousedown', this._eventHandlers.mousedown);
-                Util.addEvent(window, 'mouseup', this._eventHandlers.mouseup);
-                Util.addEvent(c, 'mouseup', this._eventHandlers.mouseup);
-                Util.addEvent(c, 'mousemove', this._eventHandlers.mousemove);
-                Util.addEvent(c, (Util.Engine.gecko) ? 'DOMMouseScroll' : 'mousewheel',
-                              this._eventHandlers.mousewheel);
             }
+            Util.addEvent(c, 'mousedown', this._eventHandlers.mousedown);
+            Util.addEvent(window, 'mouseup', this._eventHandlers.mouseup);
+            Util.addEvent(c, 'mouseup', this._eventHandlers.mouseup);
+            Util.addEvent(c, 'mousemove', this._eventHandlers.mousemove);
+            Util.addEvent(c, (Util.Engine.gecko) ? 'DOMMouseScroll' : 'mousewheel',
+                          this._eventHandlers.mousewheel);
 
             /* Work around right and middle click browser behaviors */
             Util.addEvent(document, 'click', this._eventHandlers.mousedisable);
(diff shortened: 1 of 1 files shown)


samhed:
That sounds great! I'll see if I can find some time to try it on an iOS device.

samhed:
From #613 :

> both mobile ie and firefox work fine with touch

conflicts

> I don't have Firefox installed on my touch device, btw.

What have you actually tested?

* Windows touch laptop: Edge, IE, Firefox, Chrome?
* Android tablet: Firefox, Chrome?

Is MS Edge on a touch device the only browser you have tested and found problems with? Seeing as IE is extremely similar to Edge, I am surprised that IE doesn't have the same behavior..

All websites that have touch-specific input interactions will have the same problem that we have, which in turn means that the user will probably want to change the default setting for touch input.

zarmhast:
I don't have firefox on my windows "computer" with touch, but I did test it in firefox on my android tablet.

> What have you actually tested?

- Windows 10 touch laptop (testing mouse and touch input)
  + Chrome
- Windows non-touch computer (testing mouse input)
  + Chrome
  + Edge
  + IE 11
- Android 5 phone & 6 tablet (testing touch input)
  + Chrome
  + Firefox

After our conversation also tested Edge and IE 11 on a touch laptop, but as I have already mentioned these do not have touch support enabled so they failed the touch portion of the test. And according to feature support charts, Firefox on desktop also does not support touch by default.

> Is MS Edge on a touch device the only browser you have tested and found problems with? Seeing as IE is extremely similar to Edge, I am surprised that IE doesn't have the same behavior..

I was unable to find a way to enable IE to support touch. Good thing that modern touch enabled laptops and surfaces are mostly, I think, on Windows 10 with Edge in which the user can enable touch support if needed.

> All websites that have touch-specific input interactions will have the same problem that we have, which in turn means that the user will probably want to change the default setting for touch input.

noVNC is the first "site" I see that actually depends on precise touch input. Also somehow Edge has pen support, wonder how that works. Not to mention the fact that maps, images, etc. are scrollable meaning that the browser can interpret swipes without having to let the website know that it was actually touched. In case of noVNC, somehow I doubt that making the canvas scrollable would help us solve the Edge issue on surfaces, but it just might be worth a try (an expensive try, time-wise).

A slightly less expensive try might be implementing some sort of a touch-pad area. It could be made scrollable without affecting the canvas and simply after the scroll is complete reset it back to center like the mouse pointer in some 3D games. Btw, the touch-pad was requested in some issue.

Just trying to give ideas as constructive as I can come up with.

zmedico:
I have tested https://github.com/zarmhast/noVNC/commit/a6b9d0f7423dfbe0058704e55adda125e17eb696 on a Asus Chromebook Flip C100PA, with chrome 51.0.2704.106, and I can confirm that it makes the mouse work (and the touchscreen continues to work too). For completeness, I also tested the same patch on an ASUS Chromebook C201 (which does not have a touchscreen), with chrome 51.0.2704.106, and the mouse continues to work correctly.

garveen:
On my surfacebook, only mouseXXX events are fired while touchXXX didn't if there is a mouse click. Both touchpad and a mouse.

samhed:
>  On my surfacebook, only mouseXXX events are fired while touchXXX didn't if there is a mouse click. Both touchpad and a mouse.

with the changes purposed in this PR you mean?

garveen:
@samhed I mean what would happened if there is a mouse click on a windows which has touch support.

More clearly, noVNC is **unusable** on my surfacebook **without** this PR, because noVNC only listen to touch events, and a mouse click only fire mouse events.

samhed:
> More clearly, noVNC is unusable on my surfacebook without this PR, because noVNC only listen to touch events, and a mouse click only fire mouse events.

Does it work well for you too if you apply these changes?

samhed:
I will test this patch along with the tests for #650 and merge afterwards.

samhed:
Merged in 795719ec3d519d39b8f9ed7d0e352ff71144cc17

PR #622, opened 2016-06-08
Make noVNC_screen always visible in vnc_auto.html.

### LINKED ISSUE #621: display:none on <div id="noVNC_screen"> in vnc_auto.html
In the latest version of noVNC, the page vnc_auto.html is completely grey, because the main noVNC `<div id="noVNC_screen">` element has `display: none` (from base.css:125).

This makes noVNC unusable.

jankeromnes:
Fixes #621.

### DIFF
--- a/vnc_auto.html
+++ b/vnc_auto.html
@@ -46,7 +46,7 @@
 </head>
 
 <body style="margin: 0px;">
-    <div id="noVNC_screen">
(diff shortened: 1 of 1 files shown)


samhed:
Good catch, however in commit 553864e the name of noVNC_screen was changed. Fixed in 3daa86c

PR #623, opened 2016-06-09
Continuous Updates

samhed:
Continuous Updates and Fence
--------------------------------
Continuous updates in the context of VNC means that, instead of waiting for the client to request frame buffer updates, the server will keep sending updates continuously.

The two extensions called "Fence" and "Continuous Updates" are needed for this and are implemented in this PR, as described in the RFB protocol available here:

https://github.com/rfbproto/rfbproto

While the continuous update functionality itself doesn't strictly require fencing, without it, buffer limits on the network could easily be hit. This is where fencing comes in, it is a flow-control mechanism. The server will use the fencing messages to determine how often it will send updates.

If Continuous Updates isn't supported by the server, noVNC will fall back on the basic current behavior of sending update requests.

Expected performance improvement
--------------------------------
Currently, when in normal operation, noVNC only sends a frame buffer update request after it is finished processing the previous update. The current design means that it goes like this:

Example: 100 ms RTT, not accounting for decoding and drawing

time(s)	|	event
--------|-------------------
 0	| noVNC updates screen
 0	| noVNC sends request
 0.05	| Server receives request
 0.05	| Server sends update
 0.1	| noVNC receives update
 0.1	| noVNC updates screen

This means, in terms of fps, that you can at max reach a fps of 1 / latency. That would for example mean 10 fps when you have a round-trip latency of 100ms.

With the new design the fps can't be as easily predicted but as you imagine, it will instead be dependent on processing speed of the client and the sizes of the network buffers.

I have made a few screen recordings demonstrating the improvement, continuous updates are enabled in the browser on the right side, a simulated 100 ms round-trip latency is added for both connections:

http://webm.land/media/brWk.webm
http://webm.land/media/Xvor.webm
http://webm.land/media/RQqw.webm

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -67,7 +67,6 @@ var RFB;
         this._display = null;           // Display object
         this._keyboard = null;          // Keyboard input handler object
         this._mouse = null;             // Mouse input handler object
-        this._sendTimer = null;         // Send Queue check timer
         this._disconnTimer = null;      // disconnection timer
         this._msgTimer = null;          // queued handle_msg timer
 
@@ -268,8 +267,6 @@ var RFB;
             RFB.messages.keyEvent(this._sock, XK_Delete, 0);
             RFB.messages.keyEvent(this._sock, XK_Alt_L, 0);
             RFB.messages.keyEvent(this._sock, XK_Control_L, 0);
-
-            this._sock.flush();
         },
 
         xvpOp: function (ver, op) {
@@ -303,14 +300,11 @@ var RFB;
                 RFB.messages.keyEvent(this._sock, code, 1);
                 RFB.messages.keyEvent(this._sock, code, 0);
             }
-
-            this._sock.flush();
         },
 
         clipboardPasteFrom: function (text) {
             if (this._rfb_state !== 'normal') { return; }
             RFB.messages.clientCutText(this._sock, text);
-            this._sock.flush();
         },
 
         // Requests a change of remote desktop size. This message is an extension
@@ -386,11 +380,6 @@ var RFB;
         },
 
         _cleanupSocket: function (state) {
-            if (this._sendTimer) {
-                clearInterval(this._sendTimer);
-                this._sendTimer = null;
-            }
-
             if (this._msgTimer) {
                 clearInterval(this._msgTimer);
                 this._msgTimer = null;
@@ -564,7 +553,6 @@ var RFB;
         _handleKeyPress: function (keysym, down) {
             if (this._view_only) { return; } // View only, skip keyboard, events
             RFB.messages.keyEvent(this._sock, keysym, down);
-            this._sock.flush();
         },
 
         _handleMouseButton: function (x, y, down, bmask) {
@@ -669,10 +657,6 @@ var RFB;
             if (this._rfb_version > this._rfb_max_version) {
                 this._rfb_version = this._rfb_max_version;
             }
-
-            // Send updates either at a rate of 1 update per 50ms, or
-            // whatever slower rate the network can handle
-            this._sendTimer = setInterval(this._sock.flush.bind(this._sock), 50);
 
             var cversion = "00" + parseInt(this._rfb_version, 10) +
                            ".00" + ((this._rfb_version * 10) % 10);
@@ -992,7 +976,6 @@ var RFB;
 
             this._timing.fbu_rt_start = (new Date()).getTime();
             this._timing.pixels = 0;
-            this._sock.flush();
 
             if (this._encrypt) {
                 this._updateState('normal', 'Connected (encrypted) to: ' + this._fb_name);
@@ -1095,7 +1078,6 @@ var RFB;
                     var ret = this._framebufferUpdate();
                     if (ret) {
                         RFB.messages.fbUpdateRequests(this._sock, this._display.getCleanDirtyReset(), this._fb_width, this._fb_height);
-                        this._sock.flush();
                     }
                     return ret;
 
@@ -1285,6 +1267,7 @@ var RFB;
             buff[offset + 7] = keysym;
 
             sock._sQlen += 8;
+            sock.flush();
         },
 
         pointerEvent: function (sock, x, y, mask) {
@@ -1302,6 +1285,7 @@ var RFB;
             buff[offset + 5] = y;
 
             sock._sQlen += 6;
+            sock.flush();
         },
 
         // TODO(directxman12): make this unicode compatible?
@@ -1327,6 +1311,7 @@ var RFB;
             }
 
             sock._sQlen += 8 + n;
+            sock.flush();
         },
 
         setDesktopSize: function (sock, width, height, id, flags) {
@@ -1362,6 +1347,7 @@ var RFB;
             buff[offset + 23] = flags;
 
             sock._sQlen += 24;
+            sock.flush();
         },
 
         pixelFormat: function (sock, bpp, depth, true_color) {
@@ -1397,6 +1383,7 @@ var RFB;
             buff[offset + 19] = 0;   // padding
 
             sock._sQlen += 20;
+            sock.flush();
         },
 
         clientEncodings: function (sock, encodings, local_cursor, true_color) {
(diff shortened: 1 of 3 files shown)


samhed:
The following commit for Websockify is related to this PR. Especially with ContinuousUpdates, the delays introduced by Nagle would be a bottle-neck on high latency connections.
https://github.com/kanaka/websockify/commit/f23780eb42ec9a37753556734d1185c1742a3ae7

kanaka:
@samhed if someone asks me what the ideal PR looks like, this is the one I'm going to point at :-). Well documented PR with rationale and test results, commits are split into logical chunks with well formatted and informative commit messages, and full tests included. Very nice.

I think the only minor concern I have is that this might cause bandwidth usage to increase in both directions (both the noVNC changes and also the disabling of Nagle's algorithm in websockify). Any chance you could re-run one of your tests you video captured and compare total wire bandwidth before and after. It's needs to be on the wire bandwidth because the change would be due to both increased packet overhead and also payload (more frequent server to client updates). If the increase is significant, we might want to provide a knob (query param and UI selector) to disable continuous updates. The increased number of client to server messages probably aren't that consequential so we probably don't need a knob for that (although characterizing the bandwidth increase is probably worthwhile still). Also, I wouldn't bother with a switch for websockify. I think Nagle's should have been disable long ago given that websockify is an interactive proxy, so that's the probably just the right thing to do all the time.

Anyways, apart from my minor concern about bandwidth I reviewed all the changes and they look good. As a historical note, keeping bandwidth low was one of my original priorities with noVNC. Browsers were just starting to implement WebSocket support at the time so I didn't want to push them too hard and I didn't want IaaS providers to get hit with huge unexpected bandwidth costs. Those concerns are much less relevant now. However, since use of noVNC seems to be moving more into mobile devices and given many mobile ISP have bandwidth caps/charges (at least here in the U.S.), it's probably prudent of us to characterize how much bandwidth increase we might cause with this change and give users an option if it's a significant change.

Nice work!

samhed:
Thank you for reviewing the changes so quickly!

Yes, it will increase bandwidth usage for high latency connections. Low latency connections are mostly unaffected. I understand your point regarding a bandwidth limiting option, but users do still have the same problem even with the current master and there are better ways of limiting the bandwidth usage than not using continuous updates.

I have made some bandwidth measurements before and after for different latency numbers:

---

**Latency close to zero, master**
2 no latency disabled continuousupdates bandwidth

**Latency close to zero, continuousupdates**
1 no latency enabled continuousupdates bandwidth

---

**50 ms latency, master**
3 50 latency disabled continuousupdates bandwidth

**50 ms latency, continuousupdates**
4 50 latency enabled continuousupdates bandwidth

---

**100 ms latency, master**
6 100 latency disabled continuousupdates bandwidth

**100 ms latency, continuousupdates**
5 100 latency enabled continuousupdates bandwidth

---

I don't think that latency should limit update rates, bandwidth and processing speed should be the limiting factors. This is one of the main reasons for why I believe that we should, if the server supports it, always use continuous updates.

kanaka:
@samhed those are very informative graphs, thanks. What tool(s) did you use to measure/graph the data? Is the blue line the client to server bandwidth? Any idea what causes the spike in the second graph?

So if I'm reading it right, it looks like 50ms latency approximately doubles the server to client bandwidth and 100ms latency approximately triples it. I suspect somewhere between 50ms and 100ms will be pretty typical for IaaS type deployments of noVNC.

The changes are clearly a big improvement to the noVNC user experience and I definitely want them in master and used by default. I just want to make sure that projects/organizations that integrate noVNC are not caught unaware by a sudden spike in bandwidth carriage costs. So at the very least, prior to the next version release, we'll want to make loud announcements in the appropriate forums. 

You mentioned better ways of managing bandwidth. I would be interested in your thoughts on that.

@DirectXMan12 I'm fine with merging these changes once you've had a chance to review and approve. Regarding the bandwidth, you probably have the best pulse on noVNC's largest user-base (openstack). Do you have any thoughts on how this change will be regarded and how to best publicize it there?

samhed:
> What tool(s) did you use to measure/graph the data? Is the blue line the client to server bandwidth?Any idea what causes the spike in the second graph?

I used Gnome's system-monitor on the server, so nothing fancy :) The red line is received data and the blue line is sent. In these tests I played the same 40 second youtube video in the remote desktop every time. The spike in the first graph isn't noVNC related.

> You mentioned better ways of managing bandwidth. I would be interested in your thoughts on that.

One thing that we could use that I believe is supported by most servers is options for higher JPEG compression as well as lower color levels.

samhed:
@DirectXMan12 you think you will have time to look at this anytime soon?

EquisTango:
Hi, this is of great interest to me. But... which VNC servers support continuous updates? I can't find updated information on the web. I'm talking about VNC servers for Windows, Linux or MacOS.

CendioOssman:
TightVNC was the original implementation, but it wasn't terribly smart about it and I don't know if it's still around in current versions of TightVNC.

I'd say have a look at TigerVNC and TurboVNC. Those are the ones I know have implemented this. Neither of them have an OS X server though.

samhed:
I would like to get this merged. Unless someone has any objections I will merge in the upcoming days.

include/rfb.js
```diff
@@ -1131,6 +1138,19 @@ var RFB;
...
+                case 150: // EndOfContinuousUpdates
+                    var first = !(this._supportsContinuousUpdates);
+                    this._supportsContinuousUpdates = true;
+                    this._enabledContinuousUpdates = false;
+                    if (first) {
+                        this._enabledContinuousUpdates = true;
+                        this._updateContinuousUpdates();
+                        Util.Info("Enabling continuous updates.");
+                    } else {
+                        // FIXME: may need to send a framebufferupdaterequest here
```
DirectXMan12: Yeah, probably, although we should only receive this message if we ask to stop the requests, right?
Still, just in case, it's probably not a bad idea.

DirectXMan12:
Sorry it took me so long to take a look at this.  I have to agree with @kanaka -- this PR was really nice to review.  Just one small inline comment.  Also, it a follow-up PR, we may want to expose an option to enable or disable continuous updates at run time.

DirectXMan12:
Otherwise, LGTM :+1:

dkelson:
Does the VNC server in qemu support continuous updates?

include/rfb.js
samhed: Correct, but it's a small change, I'll add it.

samhed:
> Also, it a follow-up PR, we may want to expose an option to enable or disable continuous updates at run time.

Ok, but I won't think I will have time for that in the near future.

> Does the VNC server in qemu support continuous updates?

@dkelson it doesn't look like it from looking at the list of encodings here: http://git.qemu.org/?p=qemu.git;a=blob;f=ui/vnc.h

DirectXMan12:
> Ok, but I won't think I will have time for that in the near future.

Understood, just wanted to mention it 😉

include/rfb.js
```diff
@@ -1108,6 +1144,23 @@ var RFB;
...
+                    var first = !(this._supportsContinuousUpdates);
+                    this._supportsContinuousUpdates = true;
+                    this._enabledContinuousUpdates = false;
+                    if (first) {
+                        this._enabledContinuousUpdates = true;
+                        this._updateContinuousUpdates();
+                        Util.Info("Enabling continuous updates.");
+                    } else {
+                        // FIXME: We need to send a framebufferupdaterequest here
+                        // if we add support for turning off continuous updates
```
samhed: We would have to find a way to know if we have an outstanding request or not. Clarified the FIXME-comment instead.

samhed:
What the fuck is github doing.. how can the default behavior be to squash into one gigantic commit!?

samhed:
Properly merged in commit bf568e56f04eb708e50c69b5abce2dcbf598a63e

DirectXMan12:
> how can the default behavior be to squash into one gigantic commit!?

That shouldn't be the default unless someone changed something in the repo settings, IIRC.  Weird.

PR #629, opened 2016-07-24
Make launch.sh OSX compatible and handle spaces in file paths

kernel-sanders:
Make the launch.sh script compatible with OSX (it uses the FreeBSD readlink without the -f option) and handle spaces in the base path to noVNC.

### DIFF
--- a/utils/launch.sh
+++ b/utils/launch.sh
@@ -26,8 +26,7 @@ usage() {
 }
 
 NAME="$(basename $0)"
-REAL_NAME="$(readlink -f $0)"
-HERE="$(cd "$(dirname "$REAL_NAME")" && pwd)"
+HERE="$( cd "$( dirname "$( dirname "$0" )" )" && pwd )"
 PORT="6080"
 VNC_DEST="localhost:5900"
 CERT=""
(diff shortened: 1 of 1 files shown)


utils/launch.sh
```diff
@@ -26,8 +26,7 @@ usage() {
 }
 
 NAME="$(basename $0)"
-REAL_NAME="$(readlink -f $0)"
-HERE="$(cd "$(dirname "$REAL_NAME")" && pwd)"
+HERE="$( cd "$( dirname "$( dirname "$0" )" )" && pwd )"
```
DirectXMan12: This gives a different result on my Linux box.  There are two issues here, I think.  The first is that you have too many layers of `dirname`.  The second is that you don't account for `launch.sh` itself being a symlink (which is part of what this original code dealt with).

Suppose we have `/usr/lib/novnc.sh --> /opt/vnc/novnc/utils/launch.sh`.  With the old method, we get
```
REAL_NAME="$(readlink -f $0)" == $(readlink -f /usr/lib/novnc.sh) == /opt/vnc/novnc/utils/launch.sh
HERE="$(cd "$(dirname "$REAL_NAME")" && pwd) == $(cd $(dirname /opt/vnc/novnc/utils/launch.sh) && pwd) == $(cd /opt/vnc/novnc/utils/ && pwd) == /opt/vnc/novnc/utils
```

(note that the `cd && pwd` is to prevent issues with `.`)

However, with your new version, we get

```
HERE="$( cd "$( dirname "$( dirname "$0" )" )" && pwd )" == $(cd $(dirname $(dirname /usr/lib/novnc.sh)) && pwd)) == $(cd $(dirname /usr/lib) && pwd) == $(cd /usr/ && pwd) == /usr
```

kernel-sanders: Ah, you're right. Maybe it would be easiest to do an OS check, and only use `readlink` on non-BSD based systems?

samhed:
What's the status on this?

CendioOssman:
No response.

Issue #615, opened 2016-05-19
Unsupported tight subencoding received, filter: 183


veselins:
Hi Kanaka,

Very nice plugin thanks a lot to you and to all contributes to project. However, I faced one problem with noVNC opened in Mozzila FireFox. For some reason after while when connection was established, exception has been thrown causes the noVNC to block and freez the screen. The VNC server behind continues to work just fine, and if I restart connection to same server instance, I can continue from point when exception were thrown.

Here is more information about the error message:
```
recv_message, caught exception: 
    name: Error
    message: Unsupported tight subencoding received, filter: 183
RFB.encodingHandlers.display_tight@http://<ip address>:6081/include/rfb.js:1976:31
RFB.encodingHandlers.TIGHT@http://<ip address>:6081/include/rfb.js:1991:37
RFB.prototype._framebufferUpdate@http://<ip address>:6081/include/rfb.js:1170:27
RFB.prototype._normal_msg@http://<ip address>:6081/include/rfb.js:1089:31
RFB.prototype._handle_message@http://<ip address>:6081/include/rfb.js:554:25
Websock.prototype._recv_message@http://<ip address>:6081/include/websock.js:400:21
 util.js:220:43
Util.init_logging/Util.Error() util.js:220
Websock.prototype._recv_message()

```

However, the interesting part, is that when I accessed remote machine via Chrome, there is no a problem. The noVNC just works fine. Maybe it is the way how Chrome handles that exception.

Any suggestion.

Best
Veselin

samhed:
It is probably related to the [TODO](https://github.com/kanaka/noVNC/blob/master/include/rfb.js#L2014
) just above the exception.

Do you run into the same problem if you change it to:

    Util.Error("Unsupported tight subencoding received, filter: " + filterId);

veselins:
Yes, I do still receive the problem. Now even is worst. I am disconnected:
'Disconnected: unsupported encoding 403238093'

Even I had tried to comment the whole line. And it's not working. Again I got unsupported encoding.

samhed:
Apologies, forgot that Util.Error doesn't return false.. you have to use _fail():

    this._fail("Unsupported tight subencoding received, filter: " + filterId);

veselins:
Thanks,
Seems that this resolve the problem or at least there is one workaround.

On FireFox with above fix, it works for none TrueColor connections. But if TrueColor is checked (from noVNC settings menu) I am receiving new type of error and connection is disconnected:
[...]

PR #632, opened 2016-07-31
Fixed RGB data buffer size (issue 615)

patrakov:
No test - sorry! However, you can probably create one from debug.5.js.zip in the ticket.


DirectXMan12:
LGTM :+1:

DirectXMan12:
Fixed BLIT RGB data buffer size

The RGB buffer size was incorrectly marked as `pixels * 4` instead of the correct `pixels * 3`

Fixes #615

PR #635, opened 2016-08-06
Add support to commonJS/npm modules

halan:
Hi,

I extracted all globals and introduce commonJS module standard (http://wiki.commonjs.org/wiki/CommonJS). Who want keep using packed version, can use directly the compiled files (ui.js, rfb.js and playback.js). All globals on these versions can be configured through webpack configs (https://github.com/halan/noVNC/blob/master/webpack.config.js#L14-L15).

Now we can use any npm lib as external dependency and build it for the web bundle, and we also can use npm for include noVNC as a dependency in any other project!

For generate a bundle just type `npm run build`.

Tell me if need some adjust on it or if this broke something. For me, works fine!
 
Following the changes:

  * Introduce commonJS standard
  * Remove all globals except `$D` (and... WebUtil and RFB on web bundle)
  * Add webpack to compile web versions (`ui.js`, `playback.js`, `rfb.js`)
  * Extract pako to external dependency
  * Add test coverage (with sourcemap)


kanaka:
@halan there is already a fairly long discussion about this https://github.com/kanaka/noVNC/issues/442 and some work towards making noVNC more amenable to packaging:https://github.com/kanaka/noVNC/pull/496

However, one thing we are trying to maintain (at least for the time being), is that a user and/or developer can checkout the noVNC repository and immediately begin using it and or developing without the requirement to use node/npm/build tools or the need to pull down additional dependencies. That goal is definitely in tension with the desire to make noVNC into an en vogue node/npm/webpack/browserify style javascript module. I'm not against that style of development/packaging (and use it all the time), but for the time being, the priority for noVNC is on low overhead usage of the project. That doesn't mean that the noVNC core can't be packaged well into a npm module, but the current direction that @DirectXMan12 is working towards in the PR above is that the npm/node building/packaging will be an artifact built from the web runnable code base, rather than the other way around. That will satisfy both goals at least to some degree. @DirectXMan12 is pretty busy at the moment, but hopefully soon he'll find some time to finalize and merge #496

halan:
Nice. I just needed it to manage noVNC as npm package. I keep three files that can run without any dependency at root dir: `ui.js, playback.js, rfb.js`. Some projects keep it into `/dist/`.

Well, if you need some help with anything (maybe with https://github.com/kanaka/noVNC/pull/496), I can do it. If this structure conflicts with the original proposal, I will keep my version to my own use and will try to rebase the updates.

DirectXMan12:
thanks for the PR.  I've been meaning to rebase and merge #496, I've just been super busy with other work.  Please let me know if the structure of #496 does not work for you.  Hopefully soon I'll get a chance to rebase and merge it, and then we should be able to build an official NPM package.

halan:
Great! I will try.

DirectXMan12:
#496 just merged, closing this.

PR #640, opened 2016-08-16
Update launch.sh

nunojusto:
Just a correction of port in use test algoritm.
This way we will not have problems when using port X and having some other service using zyX or any *X port


kanaka:
Merged. Thanks.

nunojusto:
yw

PR #645, opened 2016-08-30
View-only mode doesn't need keyboard or mouse

### LINKED ISSUE #644: Grabbing keyboard and mouse in "View Only" mode
Currently if noVNC "View Only" mode, the keyboard and mouse is still grabbed. This seems to deter functionality of the keyboard without providing any additional benefit. Can this be modified so that the keyboard and mouse is not grabbed in "View Only" mode?

samhed:
Grabbing the keyboard and mouse is unnecessary in view_only mode.
In fact its unnecessary to create the objects at all. Fixes #644.

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -184,13 +184,15 @@ var RFB;
             throw exc;
         }
 
-        this._keyboard = new Keyboard({target: this._focusContainer,
-                                       onKeyPress: this._handleKeyPress.bind(this)});
+        if (!this._view_only) {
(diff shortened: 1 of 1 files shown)


kanaka:
I think given this change it's safe to remove the view_only checks in the _handleKey* and _handleMouse* methods.

samhed:
If we stop creating the keyboard and mouse objects we instead need checks all over the code to see if the objects exist. See #649 instead.

PR #648, opened 2016-09-01
Remove password state

### LINKED ISSUE #646: vnc_auto.html crash
vnc_auto.html doens't work after commit 9104d3b since it seems to depend on RFB._init_msg() to not return anything. I think it's bad practice to have functions be inconsistent with return values. Any ideas on the thought behind this @kanaka or @DirectXMan12 ?

samhed:
We already have a callback mechanism for this, so let's use that. Merging this would fix #646.

### DIFF
--- a/include/rfb.js
+++ b/include/rfb.js
@@ -265,7 +265,6 @@ var RFB;
 
         sendPassword: function (passwd) {
             this._rfb_password = passwd;
-            this._rfb_state = 'Authentication';
             setTimeout(this._init_msg.bind(this), 0);
         },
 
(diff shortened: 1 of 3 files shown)


kanaka:
One of my design goals was to keep the core logic as a state machine type system as much as possible. The implementation is a bit messy and hard to follow, but the general principle is sound. In other words, by looking at the single this._rfb_state value, you can tell exactly what state the core protocol logic is in. This also allows integrators of noVNC to attach to a single callback (onUpdateState) and receive every relevant protocol state change (e.g. if somebody is currently listening for the 'password' state they will be broken). It's true that password is a bit of a different class than most of the other states, but protocol-wise it is distinct from 'Authentication' because with 'password' the server is waiting on the client and with 'Authentication' the client is waiting on the server. My vote is to solve the problem with a different approach that still keeps the state machine model intact.

samhed:
I think it's important to distinguish visual states and protocol states in the code. It feels confusing to have one variable for both types of states. I don't see why the interface should be concerned with the protocol state at all?

I agree that this change in it self is probably not good, but I am convincing myself as I go along, see: #651

PR #649, opened 2016-09-01
Fix view only

samhed:
This PR fixes a number of issues with view-only:

* Keyboard and mouse were grabbed even in view-only (#644)
* A view-only session could change the desktop size
* The UI modified the keyboard and the mouse in view-only

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1096,8 +1096,11 @@ var UI;
                 } else if (resizeMode === 'scale' || resizeMode === 'downscale') {
                     var downscaleOnly = resizeMode === 'downscale';
                     var scaleRatio = display.autoscale(screen.w, screen.h, downscaleOnly);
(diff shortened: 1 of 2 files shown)


kanaka:
I haven't tested, but I've reviewed the changes and they LGTM.

samhed:
I have tested that it works with vnc.html and vnc_auto.html. I would love if @bravegnu could check if it solves his usecase as well.

DirectXMan12:
Needs rebase to fit with new file structure.

samhed:
> Does it make sense to also hide the keyboard-related and mouse-related buttons in view-only mode? That could be a follow-up PR

Absolutely! In fact, I can add that after I rebase

samhed:
Rebased and added hiding of input related UI elements.

PR #650, opened 2016-09-01
New way of detecting touch

samhed:
This is necessary to detect touch emulation in chrome. Unless anyone knows of any problems with this, I will merge soon.

### DIFF
--- a/tests/input.html
+++ b/tests/input.html
@@ -119,7 +119,7 @@
             mouse.grab();
             message("Display initialized");
 
-            if ('ontouchstart' in document.documentElement) {
+            if (Util.isTouchDevice) {
                 message("Touch device detected");
                 document.getElementById('button-selection').style.display = "inline";
                 document.getElementById('button1').onclick = function(){ selectButton(1) };
--- a/app/ui.js
+++ b/app/ui.js
@@ -51,7 +51,6 @@ var UI;
         controlbarMouseDownOffsetY: 0,
         keyboardVisible: false,
 
-        isTouchDevice: false,
         isSafari: false,
         rememberedClipSetting: null,
         lastKeyboardinput: null,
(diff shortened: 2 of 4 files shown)


tests/input.html
```diff
@@ -119,7 +119,7 @@
             mouse.grab();
             message("Display initialized");
 
-            if ('ontouchstart' in document.documentElement) {
+            if (document.ontouchstart' !== undefined) {
                 message("Touch device detected");
```
kanaka: Extraneous single-quote maybe?

kanaka:
Are you confident this will work on other (older) touch devices too?

tests/input.html
samhed: whoops

samhed:
Well, there is actually no perfect solution here. Every way of detecting touch will have false positives and true negatives. As far I've seen this is the same as our existing detection with the exception of the chrome development touch emulation.

DirectXMan12:
Needs rebase to match with new file structure.

samhed:
Rebased and combined with the changes proposed by #613 

I will manually test the following:
- [ ] Fedora 24, Chrome
- [ ] Fedora 24, Firefox
- [ ] Android 7.0, Chrome
- [ ] Android 7.0, Firefox
- [ ] OS X 11, Chrome
- [ ] OS X 11, Firefox
- [ ] OS X 11, Safari
- [ ] iOS 10, Chrome
- [ ] iOS 10, Firefox
- [ ] iOS 10, Safari
- [ ] Windows 10, Chrome
- [ ] Windows 10, Firefox
- [ ] Windows 10, IE 11
- [ ] Windows 10, Edge
- [ ] Windows 10 touch device, Chrome
- [ ] Windows 10 touch device, Firefox
- [ ] Windows 10 touch device, IE 11
- [ ] Windows 10 touch device, Edge
- [ ] Chromebook

I feel that if these platforms don't show any false positives or any true negatives, I'm satisfied. If you feel that something important is left out from these tests, feel free to test it yourself.

samhed:
In addition to the tests above I also tested

- [x] Android 4.3, Chrome
- [x] Android 4.3, Firefox
- [x] Android 6.0, Chrome
- [x] Android 6.0, Firefox
- [x] iOS 9, Chrome
- [x] iOS 9, Firefox
- [x] iOS 9, Safari

All works fine EXCEPT for Firefox on a Microsoft Surface tablet. It has a touch screen but I can't, after much research, find any way to detect it. I believe we will have to live with this for now. The problem is mitigated by the fact that the Microsoft Surface tablets have a UI button in the start-bar for bringing up the virtual keyboard.

As a summary, this new way of detecting touch does not seem to add any false positives and adds proper detection in the following cases:

* Chromebook
* Windows 10 touch device, IE 11
* Windows 10 touch device, Edge

PR #651, opened 2016-09-02
Separate visual state from protocol state

samhed:
The user interface should not have to be concerned about the protocol state. This change allows for cleaner state logic and removes special corner cases.

Note that this change breaks dependencies for the RFB module since it no longer sends out updates regarding protocol state via onUpdateState. But this is partly by design to avoid spamming the user-interface with irrelevant state updates.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -369,7 +369,7 @@ var UI;
                         // zero means no timeout
                         UI.showStatus(msg, 'error', 0);
                         break;
-                    case 'normal':
+                    case 'connected':
                         /* falls through */
                     case 'disconnected':
                     case 'loaded':
@@ -386,7 +386,7 @@ var UI;
 
         // Disable/enable controls depending on connection state
         updateVisualState: function() {
-            var connected = UI.rfb && UI.rfb_state === 'normal';
+            var connected = UI.rfb && UI.rfb_state === 'connected';
 
             //Util.Debug(">> updateVisualState");
             document.getElementById('noVNC_setting_encrypt').disabled = connected;
@@ -1062,7 +1062,7 @@ var UI;
 
             var screen = UI.screenSize();
 
-            if (screen && UI.rfb_state === 'normal' && UI.rfb.get_display()) {
+            if (screen && UI.rfb_state === 'connected' && UI.rfb.get_display()) {
 
                 var display = UI.rfb.get_display();
                 var resizeMode = UI.getSetting('resize');
@@ -1187,7 +1187,7 @@ var UI;
         // Handle special cases where clipping is forced on/off or locked
         enableDisableViewClip: function() {
             var resizeSetting = document.getElementById('noVNC_setting_resize');
-            var connected = UI.rfb && UI.rfb_state === 'normal';
+            var connected = UI.rfb && UI.rfb_state === 'connected';
 
             if (UI.isSafari) {
                 // Safari auto-hides the scrollbars which makes them
@@ -1243,7 +1243,7 @@ var UI;
         updateViewDrag: function() {
             var clipping = false;
 
-            if (UI.rfb_state !== 'normal') return;
+            if (UI.rfb_state !== 'connected') return;
 
             // Check if viewport drag is possible. It is only possible
             // if the remote display is clipping the client display.
--- a/app/ui.js
+++ b/app/ui.js
@@ -470,8 +470,8 @@ var UI;
                 time = 1500;
             }
 
-            // A specified time of zero means no timeout
-            if (time != 0) {
+            // Error messages do not timeout
+            if (status_type !== 'error') {
                 UI.statusTimeout = window.setTimeout(UI.hideStatus, time);
             }
         },
--- a/app/ui.js
+++ b/app/ui.js
@@ -339,6 +339,7 @@ var UI;
             try {
                 UI.rfb = new RFB({'target': document.getElementById('noVNC_canvas'),
                                   'onUpdateState': UI.updateState,
+                                  'onPasswordRequired': UI.passwordRequired,
                                   'onXvpInit': UI.updateXvpButton,
                                   'onClipboard': UI.clipboardReceive,
                                   'onFBUComplete': UI.initialResize,
@@ -372,15 +373,6 @@ var UI;
                     case 'disconnected':
                     case 'loaded':
                         UI.showStatus(msg, 'normal');
-                        break;
-                    case 'password':
-                        document.getElementById('noVNC_password_dlg')
-                            .classList.add('noVNC_open');
-                        setTimeout(function () {
-                                document.getElementById(('noVNC_password_input').focus());
-                            }, 100);
-
-                        UI.showStatus(msg, 'warn');
                         break;
                     default:
                         UI.showStatus(msg, 'warn');
@@ -977,6 +969,27 @@ var UI;
             // Don't display the connection settings until we're actually disconnected
         },
 
+/* ------^-------
+ *  /CONNECTION
+ * ==============
+ *   PASSWORD
+ * ------v------*/
+
+        passwordRequired: function(rfb, msg) {
+
+            document.getElementById('noVNC_password_dlg')
+                .classList.add('noVNC_open');
+
+            setTimeout(function () {
+                    document.getElementById('noVNC_password_input').focus();
+                }, 100);
+
+            if (typeof msg === 'undefined') {
+                msg = "Password is required";
+            }
+            UI.updateState(null, "warning", null, msg);
+        },
+
         setPassword: function() {
             UI.rfb.sendPassword(document.getElementById('noVNC_password_input').value);
             document.getElementById('noVNC_password_dlg')
@@ -985,7 +998,7 @@ var UI;
         },
 
 /* ------^-------
- *  /CONNECTION
+ *  /PASSWORD
  * ==============
  *   FULLSCREEN
  * ------v------*/
--- a/app/ui.js
+++ b/app/ui.js
@@ -929,8 +929,6 @@ var UI;
         },
 
         connect: function() {
-            UI.closeAllPanels();
-
             var host = document.getElementById('noVNC_setting_host').value;
             var port = document.getElementById('noVNC_setting_port').value;
             var password = document.getElementById('noVNC_setting_password').value;
@@ -948,6 +946,8 @@ var UI;
             }
 
             if (!UI.initRFB()) return;
+
+            UI.closeAllPanels();
 
             UI.rfb.set_encrypt(UI.getSetting('encrypt'));
             UI.rfb.set_true_color(UI.getSetting('true_color'));
--- a/app/ui.js
+++ b/app/ui.js
@@ -36,7 +36,8 @@ var UI;
 
     UI = {
 
-        rfb_state: 'loaded',
+        connected: false,
+        desktopName: "",
 
         resizeTimeout: null,
         statusTimeout: null,
@@ -340,15 +341,16 @@ var UI;
                 UI.rfb = new RFB({'target': document.getElementById('noVNC_canvas'),
                                   'onNotification': UI.notification,
                                   'onUpdateState': UI.updateState,
+                                  'onDisconnected': UI.disconnectFinished,
                                   'onPasswordRequired': UI.passwordRequired,
                                   'onXvpInit': UI.updateXvpButton,
                                   'onClipboard': UI.clipboardReceive,
                                   'onFBUComplete': UI.initialResize,
                                   'onFBResize': UI.updateViewDrag,
-                                  'onDesktopName': UI.updateDocumentTitle});
+                                  'onDesktopName': UI.updateDesktopName});
                 return true;
             } catch (exc) {
-                UI.updateState(null, 'fatal', null, 'Unable to create RFB client -- ' + exc);
+                UI.showStatus('Unable to create RFB client -- ' + exc, 'error');
                 return false;
             }
         },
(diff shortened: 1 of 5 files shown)


samhed:
@kanaka in the view of your comment in #648, what do you think about this change?

samhed:
Due to #646 I think we need to decide here soon. I think this PR is the best way forward.

kanaka:
@samhed sorry for slow reply (sickness in the family).

The commit to do proper filtering of auth state LGTM. The number of silly hacks I did in the code continues to drop :-)

I definitely would not drop the comment that lists the initialization states since it is really useful to have that listed in a single place.

I agree that the state management code could definitely use some cleanup and I'm glad you are attacking it. I'm a bit hesitant to push back too much because the fact that you are taking ownership of this outweighs any feedback I might have. But that said, here is my feedback ... 

I'm still not a big fan of removing `password` state. It's a unique state in the RFB protocol flow, even if it  doesn't happen to be named in the RFB "spec". The RFB protocol state machine is unfortunately not explicitly enumerated in the docs, so it has to be somewhat inferred. "Authentication" is really two separate protocol states. The main "Authentication" state that means the client is waiting on the server whereas "password" is the authentication state where the server is waiting on the client. "password" isn't necessarily tied to noVNC UI state either. For example, consider a couple of scenarios where you have some VNC server without passwords, and other more secure ones with passwords. In one scenario the VNC server password is rotated every 5 minutes and the page that noVNC is integrated into is already authenticated/logged in and can make a secure HTTP request to get the current server password which is used as the password value without doing any UI interaction. Or you could imagine an integration of noVNC where the "password" state is used to trigger an API call to a MFA service which would prompt the users phone for confirmation, and a security token would be returned to noVNC which would then use that for the password value (again, no noVNC UI prompt/update). Note that the two above examples are a bit unrealistic (because you would probably know to get/send the password with the initial request), but they are thought experiments to show that "password" is protocol state, not just UI state. It's a protocol state that has to be integrated into a large page/system in order to respond to, but it's still a separate state in the protocol.

I'm a bit leery of splitting the state into different state machines since they are not completely separate. You usually want to limit the number of interacting state machines if you can (by adding more granular states). But there isn't really a _right_ answer here, that's just the pattern I've generally seen recommended.

Finally, it's important to note that this area of protocol and page state management has historically been a big source of issues and complexity. The lack of real RFB specs and documentation makes this inherently a "discovery by trial and error" sort of problem. I originally had a more adhoc callback system  that grew into an awful unmanageable beast that I could no longer comprehend. So I moved to the central state machine model in order to deal with the complexity. The current state is still not elegant, however, it is SO much better than it used to be and is challenging but possible to reason about. I also apologize that I didn't go further in the refactoring and arrive at something more elegant and easy to follow. Changing this area of the code will almost certainly result in issues/bugs for certain VNC server and configurations even if it passes all automated and manual testing.

But circling back to my original point. If you are willing to take on this area of the code and deal with issues that arise as a result, then please consider my above feedback as just some points to keep in mind in the future. That said, since the scope of the changes has grown beyond just the 'password' state refactoring, I would like to get feedback from @DirectXMan12 on the code and whether the changes cause any issue for OpenStack integration since I suspect some spillover changes may be needed there.

include/rfb.js
```diff
@@ -1120,7 +1120,7 @@ var RFB;
 
             switch (xvp_msg) {
                 case 0:  // XVP_FAIL
-                    this._updateState(this._rfb_state, "Operation Failed");
+                    Util.Error("Operation Failed");
```
DirectXMan12: Should there be a way to message the UI that this occurred?  It seems like this might be useful information for the user to know (and you can't assume that logging communicates anything to the user).

include/rfb.js
```diff
@@ -706,8 +706,17 @@ var RFB;
                 var types = this._sock.rQshiftBytes(num_types);
                 Util.Debug("Server security types: " + types);
                 for (var i = 0; i < types.length; i++) {
-                    if (types[i] > this._rfb_auth_scheme && (types[i] <= 16 || types[i] == 22)) {
-                        this._rfb_auth_scheme = types[i];
+                    switch (types[i]) {
```
DirectXMan12: heh :+1:

include/rfb.js
```diff
@@ -507,23 +502,25 @@ var RFB;
...
+                this._rfb_state = 'failed';
                 this._onUpdateState(this, state, oldstate);
             } else {
                 this._onUpdateState(this, state, oldstate, statusMsg);
             }
+
+            // Make sure we transition to disconnected
+            if (state === 'failed') {
+                this._updateState('disconnected');
+            }
```
DirectXMan12: shouldn't we wait until actually receiving the "disconnected" event from the WebSocket (it's late, so I may be missing something here)?

include/rfb.js
```diff
@@ -432,7 +432,21 @@ var RFB;
          *   ClientInitialization - not triggered by server message
          *   ServerInitialization (to normal)
          */
+        _stateQueue: new Array(),
```
DirectXMan12: Doesn't this need to be per-instance?

include/rfb.js
```diff
@@ -432,7 +432,21 @@ var RFB;
...
         _updateState: function (state, statusMsg) {
+            /* Recursion protection */
```
DirectXMan12: in which states can we recurse like this?  It would be nice to have a comment describing this.

include/rfb.js
```diff
@@ -432,7 +432,21 @@ var RFB;
...
+            this._stateQueue.push(new Array(state, statusMsg));
```
DirectXMan12: I'd prefer an object with descriptive property names here over a tuple-like array, I think.

include/rfb.js
samhed: Good point, will add that

include/rfb.js
samhed: You mean that I should move this to the constructor?

include/rfb.js
samhed: +1

samhed:
I have updated the code a bit. The fact that we had a number of substates of 'Disconnected' (loaded, failed, fatal) was the cause for a lot of bad workarounds. In the previous iteration, I simply tried to work around this by adding recursion protection to the state machine etc. Now however, I have arrived to a cleaner solution where we don't need these workarounds nor any recursion protection.

The RFB connection state machine looks simple enough now:
statescropped

The init states (ProtocolVersion, Security, Authentication, SecurityResult, ClientInitialization, ServerInitialization) are all substates of the 'Connecting'-state.

I still need to rebase the code after the updates in master.

> Perhaps we should add in an alternate callback for [the init states] so that people can still see the transition there?

Could do, but I would argue that these states are internal for the RFB module and other modules which depend on these are badly written.

include/rfb.js
samhed: fixed

include/rfb.js
samhed: I have now removed the need for this protection

samhed:
Rebased and done! PTAL @kanaka and @DirectXMan12

samhed:
Found two more small issues that are fixed now.

include/rfb.js
DirectXMan12: yeah, otherwise, if you have two instances of the RFB object, they'll be using the same queue.

core/rfb.js
```diff
@@ -245,10 +245,10 @@
     });
 
     this._init_vars();
+    this._cleanupSocket();
```
DirectXMan12: We shouldn't have an open socket, so there should be no need to close it here (see comment below).  The other related stuff (clearing the screen, etc) should probably be in it's own function (at the very least, the name of `_cleanupSocket` is confusing here, IMO).

core/rfb.js
```diff
@@ -482,60 +456,66 @@
...
-                        Util.Error("Invalid transition from 'disconnected' or 'failed' to 'connected'");
+                    if (oldstate === 'disconnected') {
+                        Util.Error("Invalid transition from 'disconnected' to 'connected'");
                     }
                     break;
 
                 case 'connecting':
+                    // A previous connect may asynchronously cause a connection
+                    // so make sure we are still closed.
+                    this._cleanupSocket();
```
DirectXMan12: Since we do one connection per RFB object, there should be no open connection here, I think.

DirectXMan12: Also, IMO, once we disconnect, we should go into a permanent disconnected state, so that the user doesn't accidentally try to reconnect.

core/rfb.js
```diff
@@ -482,60 +456,66 @@
...
+                    Util.Error("Error while connected: " + msg);
+                    break;
+                case 'connecting':
+                    Util.Error("Error while connecting: " + msg);
+                    break;
+                default:
+                    Util.Error("RFB error: " + msg);
+                    break;
+            }
+            this._updateConnectionState('disconnected', msg);
```
DirectXMan12: I like this, but I think this should actually probably be "disconnecting" (because we haven't yet closed the WebSocket).  I think the old code (which jumped from "failed" to "disconnected") was either unclear or "incorrect" (in how it labeled states).

app/ui.js
```diff
@@ -270,99 +272,98 @@ var UI;
...
-                case 'fatal':
-                    klass = "noVNC_status_error";
+                case 'connecting':
+                    UI.setStatus("Connecting");
                     break;
                 case 'connected':
-                    klass = "noVNC_status_normal";
+                    UI.connected = true;
+                    if (rfb && rfb.get_encrypt()) {
+                        UI.setStatus("Connected (encrypted) to " +
```
DirectXMan12: :+1: this is good -- we should keep text like this in the UI.

core/rfb.js
```diff
@@ -420,134 +428,125 @@
...
             return false;
         },
 
+        /*
+         * Send a notification to the UI. Accepted optional options are:
+         *    { level: 'warn'|'error' }
```
DirectXMan12: what's the purpose of this being an object?  Why not just an optional "level" argument?

DirectXMan12: (more specifically, I think level is important enough to be it's own argument here).

include/rfb.js
samhed: This code is removed in the latest iteration

core/rfb.js
samhed: You mean permanent for each RFB object? So that you can't change from 'disconnected' to 'connecting'? In that case we can make the starting state empty

core/rfb.js
samhed: fixed

core/rfb.js
samhed: fixed

core/rfb.js
samhed: agreed, fixed

core/rfb.js
samhed: You are right, fixed now

samhed:
Ok, all done now! Found a few more errors which I fixed. My changes have full test coverage now as well.

core/rfb.js
ossman: There is one issue remaining here with this approach, and that's the fact that the RFB object leaves stray event listeners around. These need to be cleaned up on disconnect so that all references are properly dropped. All the ones left over originate in devices.js.

samhed: It seems like the browsers have some issues in displaying active event listeners properly in their development tools. All listeners that are added on the Mouse and Keyboard objects (devices.js) are properly removed on cleanup. Events are only added in the grab functions and are all removed in the ungrab functions (which are called on cleanup).

    $ git grep -n -e "addEventListen" -e "removeEventListen" core/input/devices.js
    core/input/devices.js:125:            c.addEventListener('keydown',     this._eventHandlers.keydown);
    core/input/devices.js:126:            c.addEventListener('keyup', this._eventHandlers.keyup);
    core/input/devices.js:127:            c.addEventListener('keypress', this._eventHandlers.keypress);
    core/input/devices.js:130:            window.addEventListener('blur', this._eventHandlers.blur);
    core/input/devices.js:139:            c.removeEventListener('keydown', this._eventHandlers.keydown);
    core/input/devices.js:140:            c.removeEventListener('keyup', this._eventHandlers.keyup);
    core/input/devices.js:141:            c.removeEventListener('keypress', this._eventHandlers.keypress);
    core/input/devices.js:142:            window.removeEventListener('blur', this._eventHandlers.blur);
    core/input/devices.js:350:                c.addEventListener('touchstart', this._eventHandlers.mousedown);
    core/input/devices.js:351:                window.addEventListener('touchend', this._eventHandlers.mouseup);
    core/input/devices.js:352:                c.addEventListener('touchend', this._eventHandlers.mouseup);
    core/input/devices.js:353:                c.addEventListener('touchmove', this._eventHandlers.mousemove);
    core/input/devices.js:355:                c.addEventListener('mousedown', this._eventHandlers.mousedown);
    core/input/devices.js:356:                window.addEventListener('mouseup', this._eventHandlers.mouseup);
    core/input/devices.js:357:                c.addEventListener('mouseup', this._eventHandlers.mouseup);
    core/input/devices.js:358:                c.addEventListener('mousemove', this._eventHandlers.mousemove);
    core/input/devices.js:359:                c.addEventListener((Util.Engine.gecko) ? 'DOMMouseScroll' : 'mousewheel',
    core/input/devices.js:364:            document.addEventListener('click', this._eventHandlers.mousedisable);
    core/input/devices.js:365:            document.body.addEventListener('contextmenu', this._eventHandlers.mousedisable);
    core/input/devices.js:372:                c.removeEventListener('touchstart', this._eventHandlers.mousedown);
    core/input/devices.js:373:                window.removeEventListener('touchend', this._eventHandlers.mouseup);
    core/input/devices.js:374:                c.removeEventListener('touchend', this._eventHandlers.mouseup);
    core/input/devices.js:375:                c.removeEventListener('touchmove', this._eventHandlers.mousemove);
    core/input/devices.js:377:                c.removeEventListener('mousedown', this._eventHandlers.mousedown);
    core/input/devices.js:378:                window.removeEventListener('mouseup', this._eventHandlers.mouseup);
    core/input/devices.js:379:                c.removeEventListener('mouseup', this._eventHandlers.mouseup);
    core/input/devices.js:380:                c.removeEventListener('mousemove', this._eventHandlers.mousemove);
    core/input/devices.js:381:                c.removeEventListener((Util.Engine.gecko) ? 'DOMMouseScroll' : 'mousewheel',
    core/input/devices.js:386:            document.removeEventListener('click', this._eventHandlers.mousedisable);
    core/input/devices.js:387:            document.body.removeEventListener('contextmenu', this._eventHandlers.mousedisable);

samhed:
Rebased once again. I also found 3 small new semi-unrelated problems which I added fixes for. Last chance for comments before I merge this!

core/util.js
```diff
@@ -329,7 +329,7 @@ Util.Features = {xpath: !!(document.evaluate), air: !!(window.runtime), query: !
         'presto': detectPresto(),
         'trident': detectTrident(),
         'webkit': detectInitialWebkit(),
-        'gecko': detectGecko(),
+        'gecko': detectGecko()
```
DirectXMan12: this is more a matter of personal taste, but the trailing comma makes diffs that add addition options (as well as blame logs) better, since adding a new line only shows the new line being added, instead of also showing the old line having a comma added as well.

DirectXMan12:
I think tihs LGTM.

PR #653, opened 2016-09-05
Change control bar to a hiding toolbar

CendioOssman:
Me and @samhed have been reworking the control bar to a hiding toolbar instead. The main purpose is to free up valuable screen real estate, but we've also put some time in to making the thing a bit more perty.

The large number of commits is because the JavaScript and the HTML/CSS were a bit too interconnected to do large interface changes like this. So the first two thirds of the commits are various cleanups that we did along the way.

Tested on Chrome, IE, Safari and Firefox on Windows, Linux, iOS and Android.

Before:
screenshot from 2016-09-05 09-41-32

After:
screenshot from 2016-09-05 09-42-05

Transparent after being idle:
screenshot from 2016-09-05 09-41-00

Folded away:
screenshot from 2016-09-05 09-41-11


samhed:
Fixes #599

CendioOssman:
Updated the patch set:

Removed the Shift and AltGr additions as they don't work like the other buttons and are most likely confusing.

Added a border for the status popup as otherwise it can be difficult to see if the background happens to match the popup.

CendioOssman:
Updated the patches again based on internal feedback and testing. Some visual tweaks, and handling for some corner cases. Mostly it is the display of the extra keys and status that has changed:

screenshot from 2016-09-14 17-49-30

screenshot from 2016-09-14 17-50-32

kanaka:
@CendioOssman @samhed this is looking great!

Have you had any chance to test this on mobile devices and see how well it works there? Especially in terms of the hidden away button being easy to push without getting in the way and also the interaction with mouse panning mode.

samhed:
Yes we have, some more thourogh testing is undergoing however. Personally I feel good about this now, I don't believe we will do further changes.

The handle gets a larger hitbox on touch devices. It feels good on both iOS and Android, even on devices with small screens. What do you mean regarding panning?

kanaka:
By panning I meant viewport dragging (the toggle for moving the viewport when it's smaller than the server display). It seems like it that mode you might not want to hide the control bar or maybe just have a longer timeout for hiding or something. Don't really know, just curious if you tested it in that mode and if it still works well.

samhed:
The control bar does only automatically hide 2 seconds after connecting. In all other situations, if you don't use the control bar for a few seconds it won't hide, it will just become a little bit transparent. If you want to fully hide it you have to press the handle again.

We also tested panning and it seems to work great!

kanaka:
Great! It would be good to get @DirectXMan12 to sign off since it's a pretty large change. It doesn't look like much impact to vnc_auto.html so I think it shouldn't have much impact on OpenStack, but it would be good for him to give it the once over.

DirectXMan12:
I'll try to take a peek this afternoon.  You might have to resolve a couple merge conflicts due to #496

DirectXMan12:
Ok, so I'm going to start out with some fluffy comments.  Don't take any of this comments in this particular comment *too* seriously 😛 

1. This is awesome. 🎉 
2. Seriously, this looks *really* nice.  Loop back to comment 1 a few times to get the full effect.
3. I'm going to be *super* nitpicky here and state that I much prefer a deploy animation like `0.5s cubic-bezier(.68,-0.55,.27,1.55)` for deploying the subpanels instead of a plain `ease-in-out` (it has a nice little bit of bounciness to it, but that was just from a quick playing around in the Firefox dev tools).  The thing that lead me to play around a bit there was that the ease-in felt a bit slow, but playing with the curves made it feel faster w/o changing the time.
4. The label next to "Style" and "Logging" depresses when I click on the corresponding dropdown.  I think it's caused by setting `border-top` under when `:active` in `base.css`.
5. There's no contextual hint (hover-hint), that the buttons/drop-downs are drop-downs, and w/o the little arrow, there's little to indicate that it's a dropdown or that it's pressable (notice, for instance, the GitHub comment button becomes a slight bit darker on hover).

I'll probably have some more concrete code comments as I go along.

include/util.js
```diff
@@ -482,7 +482,7 @@ Util.addEvent = function (obj, evType, fn) {
         obj.addEventListener(evType, fn, false);
         return true;
     } else {
-        throw new Error("Handler could not be attached");
+        throw new Error("Handler for \"" + evType + "\" on object \"" + obj + "\" could not be attached");
```
DirectXMan12: Nit: instead of escaping the double quotes, it's probably more readable to just use single quotes as the outer quotes.

DirectXMan12: Actually, this entire method is gone in master now, so no need to fix it.

include/ui.js
```diff
@@ -205,47 +205,47 @@ var UI;
...
-            $D("noVNC_clipboard_text").onblur = UI.displayFocus;
-            $D("noVNC_clipboard_text").onchange = UI.clipboardSend;
-            $D("noVNC_clipboard_clear_button").onclick = UI.clipboardClear;
-
-            $D("noVNC_settings_apply").onclick = UI.settingsApply;
-
-            $D("noVNC_connect_button").onclick = UI.connect;
-
-            $D("noVNC_setting_resize").onchange = UI.enableDisableViewClip;
+            Util.addEvent($D("noVNC_view_drag_button"), 'click', UI.toggleViewDrag);
```
DirectXMan12: just a heads up: this is all just going to need to be in the form of `document.getElementById("xyz").addEventListener('event', some_func)` now.

DirectXMan12: Otherwise, yeah, good idea, just make sure we don't try to unset it elsewhere.

include/util.js
```diff
@@ -508,6 +512,82 @@ Util.stopEvent = function (e) {
...
+    Util._captureElem.dispatchEvent(newEv);
+    Util._captureRecursion = false;
+
+    // Implicitly release the capture on button release
+    if ((e.type === "mouseup") || (e.type === "touchend")) {
+        Util.releaseCapture();
+    }
+}
+
+Util.setCapture = function (elem) {
```
DirectXMan12: this should be in WebUtil, since it's only used in the UI.

CendioOssman:
> 1. This is awesome. :tada:
> 2. Seriously, this looks really nice. Loop back to comment 1 a few times to get the full effect.

Thanks. We're rather happy with it ourselves. :)

> 3. I'm going to be super nitpicky here and state that I much prefer a deploy animation like 0.5s cubic-bezier(.68,-0.55,.27,1.55)...

I think this is one of those subjective things. I feel that spline is a bit too aggressive and you don't really get the same feel that the object slows down before landing. So unless this is a hard requirement I'm sticking with the default spline. ;)

(These things are also super-easy to override for your own site with the cleaned up style sheet)

> 4. The label next to "Style" and "Logging" depresses when I click on the corresponding dropdown. I think it's caused by setting border-top when :active in base.css.

It's caused by the extra top margin, which was necessary to get the thing to move downwards rather than up. I think I've found an alternative now though that should work.

> 5. There's no contextual hint (hover-hint), that the buttons/drop-downs are drop-downs, and w/o the little arrow, there's little to indicate that it's a dropdown or that it's pressable (notice, for instance, the GitHub comment button becomes a slight bit darker on hover).

We had hover hints to begin with, but they were annoying on touch where the hover never turned off again, so we nuked them. If you really want then we can add something back, but I'm hoping the slightly 3D look is sufficient hint.

The arrow seems to be impossible to get on there, given how magical input elements are. :/

include/ui.js
```diff
@@ -171,18 +171,6 @@ var UI;
                 UI.updateViewDrag();
                 UI.setBarPosition();
             } );
-
-            // While connected we want to display a confirmation dialogue
-            // if the user tries to leave the page
-            Util.addEvent(window, 'beforeunload', function (e) {
-                if (UI.rfb && UI.rfb_state === 'normal') {
```
CendioOssman: See here: https://developer.mozilla.org/en/docs/Web/Events/beforeunload
In short. Firefox doesn't show the message. Some mobile browsers doesn't show it all. The HTML standard seems to be moving towards not allowing any blocking of leaving a page.

CendioOssman:
> Overall, pretty good. However, I think a number of things need to be squashed together -- the CSS cleanups, for example, when possible,

I'm not seeing any clear candidates. I don't think that commits which just move things around should be mixed with commits that actually change things. It's very difficult to read the diffs if you do that.

> or the setCapture polyfill, which should be in the same commit where setCapture is used.

That I can do.

include/ui.js
```diff
@@ -84,8 +80,8 @@ var UI;
             UI.addSettingsHandlers();
 
             // Show the connect panel on first load unless autoconnecting
-            if (!autoconnect && !UI.connSettingsOpen)
-                UI.toggleConnectPanel();
+            if (!autoconnect)
```
CendioOssman: Given a choice I prefer to leave them out, but leaving them in is the dominant style here. I'll fix it.

PS. Is there a style guide somewhere for noVNC?

include/util.js
CendioOssman: Sure, I'll fix.

CendioOssman:
This should be fully updated now.

include/ui.js
DirectXMan12: > PS. Is there a style guide somewhere for noVNC?

Not officially.  It's mostly personal preferences and jshint.

DirectXMan12:
> I think this is one of those subjective things. I feel that spline is a bit too aggressive and you don't really get the same feel that the object slows down before landing. So unless this is a hard requirement I'm sticking with the default spline. ;)

That's fine.  I was being super-nitpicky, and it's largely a personal preference (alternatively, I wouldn't mind a smidge of a speed-up on the timing, maybe from 0.5s to 0.4s, but feel free to completely ignore that, since that's also almost entirely personal preference :-)).

> It's caused by the extra top margin, which was necessary to get the thing to move downwards rather than up. I think I've found an alternative now though that should work.

Well, on the up side it looks good.  On down side, I can no longer use the dropdowns (Firefox Nightly) -- the actual drop down appears and then closes immediately when I click the button.  It seems any kind of movement (position, transform, etc) causes the dropdown to close.  Not sure of a good solution yet.

Also, is there a reason you're not using `box-shadow` to simulate a shadow (instead of `margin-bottom`)?

> We had hover hints to begin with, but they were annoying on touch where the hover never turned off again, so we nuked them. If you really want then we can add something back, but I'm hoping the slightly 3D look is sufficient hint.

I would definitely prefer it.  You should be able to get away with something like
`:root:not(.noVNC_touch) input[type=button]:hover, :root:not(.noVNC_touch) select:hover` as the rule (I just tested it on my phone and, it seems to work fine) to prevent the sticky effect on touch devices.  In the future, we'll be able to use `hover` media query (once it gets decent support -- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover).

DirectXMan12:
> Not sure of a good solution yet.

The solution may be to just go to a "convex-concave" style button, like

```css
input[type=button], select {
  /* Disable default rendering */
  -webkit-appearance: none;
  -moz-appearance: none;
  background: none;

  border: none;
  background: linear-gradient(to top, rgb(240, 240, 240), rgb(255, 255, 255));
  border: 1px solid rgb(192, 192, 192);
  border-radius: 5px;
  padding: 2px;
  margin: 2px;
}

While I'm running things by you, assuming the touch detection mentioned above, what about

```css
.noVNC_button:hover {
  background: rgba(0, 0, 0, 0.2);
  border-color: rgba(0, 0, 0, 0.4);
}
```
as well?
input[type=button]:hover, select:hover {
  background: linear-gradient(to top, rgb(230, 230, 230), rgb(255, 255, 255));
}

input[type=button]:active, select:active {
  background: linear-gradient(to top, rgb(255, 255, 255), rgb(240, 240, 240));
}
```

CendioOssman:
We've found some issue on iOS that needs attention as well. We'll update the patches shortly. :)

CendioOssman:
> Well, on the up side it looks good. On down side, I can no longer use the dropdowns (Firefox Nightly) -- the actual drop down appears and then closes immediately when I click the button. It seems any kind of movement (position, transform, etc) causes the dropdown to close. Not sure of a good solution yet.

Dammit. :/

Fortunately I figured out why it was jumping around. It was aligning the baseline of the text in the input element with the baseline outside of it. Got a proper fix now and back to the original approach.

> I would definitely prefer it. You should be able to get away with something like
> :root:not(.noVNC_touch) input[type=button]:hover, :root:not(.noVNC_touch) select:hover as the rule (I just tested it on my phone and, it seems to work fine) to prevent the sticky effect on touch devices. In the future, we'll be able to use hover media query (once it gets decent support -- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/hover).

Good idea. I came up with the .noVNC_touch stuff much later, so it didn't occur to me to use it for the hovering. :)

CendioOssman:
Updated an ready for re-review. :)

CendioOssman:
People got confused by the merge of the keyboard and extra keys buttons, so we dropped that change for now. We'll have to revisit it at a later point.

CendioOssman:
Good to merge now?

samhed:
Unless someone has any further comments, I will merge this tomorrow.

DirectXMan12:
@kanaka can we get a tweet about this when it merges?

Also, we should open a conversation about the copyright lines in a separate issue, probably.

kanaka:
@DirectXMan12 I was on vacation until today. I just tweeted this: https://twitter.com/noVNC/status/784172373146611712 Thanks for the hard work everyone!

PR #654, opened 2016-09-06
Add Meta Key mapping to xtscancodes.js

asadpanda:
OSLeft/OSRight have been renamed to MetaLeft/MetaRight respectively.

This change will allow the Super/Windows key to be used under the qemu rfb extension under Chrome and updated browsers.

https://github.com/w3c/uievents-code/issues/7
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code


samhed:
Do we need to keep the old names for some sort of compatibility?

asadpanda:
Yes, I would leave it for compatibility. https://bugzilla.mozilla.org/show_bug.cgi?id=1264150 is still open for firefox, even though they have updated their docs, it still sends the code "OSLeft" or "OSRight". That's actually how i discovered the bug in the first place, it worked fine under firefox.

samhed:
Ok, could you add a comment to OSLeft and OSRight which explains the situation and I'll merge this

PR #661, opened 2016-09-16
Detect RealVNC's VNC 5.0 Protocol Version

DirectXMan12:
The RFB protocol specifies a max version of 3.8, but RealVNC likes to
increment their version when they add new features.  RealVNC 5.3 sends
"005.000" as the version string, since they've extended their
implementation upon RFB v3.8.  We can detect that, and just tell them we
only speak 3.8, instead of barfing on "005.000" as an invalid version.

Closes #585


DirectXMan12:
Thanks to @larsolevard for the original PR (I just cleaned it up a bit and rebased).

PR #662, opened 2016-09-17
Add external event handler for WebSocket connection events

ayonghosh:
Currently, noVNC does not offer a way for external handlers on WebSocket errors. The default behavior involves an error on the console but there is no way to communicate this to the user. This is a problem we faced at one of our projects in my current organization and that is the motivation for this PR.


DirectXMan12:
Ah, ok.  That makes a lot of sense.  I'm still a bit torn as to whether or not errors like that should be actual exceptions.  @samhed @kanaka thoughts? @ayonghosh would that also work for your use case?

Anyway, if not, this PR looks pretty reasonable to me.

DirectXMan12:
It just needs a rebase, since `include/rfb.js` became `core/rfb.js`.

ayonghosh:
@DirectXMan12 Created a new pull request after rebase: #663.

DirectXMan12:
For future reference, you don't need to create a new pull request -- you can just squash and force-push.

PR #663, opened 2016-09-17
Add external event handler for WebSocket connection events

ayonghosh:
Creating a new pull request after rebasing for #662

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -165,7 +165,8 @@
         'onFBUComplete': function () { },       // onFBUComplete(rfb, fbu): RFB FBU received and processed
         'onFBResize': function () { },          // onFBResize(rfb, width, height): frame buffer resized
         'onDesktopName': function () { },       // onDesktopName(rfb, name): desktop name received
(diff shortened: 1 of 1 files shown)


DirectXMan12:
From #662: I'm still a bit torn as to whether or not errors like that should be actual exceptions. @samhed @kanaka thoughts? @ayonghosh would that also work for your use case?

Anyway, if not, this PR looks pretty reasonable to me.

ayonghosh:
@DirectXMan12 that would work.

samhed:
+1 for exceptions on websocket errors

JeffLeFevre:
+1 on exceptions for websocket errors

samhed:
To clarify, I do not think we should have a callback for WebSocket connection events. In my mind, the websocket handling should not leak to the user interface.

However, in #651 I have added a onDisconnected callback which allows you to specify a reason. We could perhaps use _fail(msg) here. Is there any situation where we wouldn't want to disconnect on WebSocket errors?

PR #665, opened 2016-09-19
Support RFB bell

CendioOssman:
The protocol handling was already there. What was missing was
hooking it up to a sound file in the user interface.


DirectXMan12:
😃 

This should probably have an option to enable or disable it (audio bells can get *really* annoying sometimes).  Also, to bikeshed a bit, we could also have a visual bell that wobbles a bit or sometime.

kanaka:
I have no objection to adding bell support. I do agree with @DirectXMan12 that it should be configurable and that a visual option would be nice (maybe off entirely, visual, audible, both). Also, I'm wondering if this could be done programmatically with some minimal Web Audio synthesis. That would lower the bell latency and reduce the initial page load size/file count. Might not be worth the effort, but worth considering since most modern browsers have had Web Audio support for a while now.

CendioOssman:
> This should probably have an option to enable or disable it (audio bells can get really annoying sometimes).

Possibly. But the browser's already got handling for muting pages. And on the next level the OS can usually mute applications, and finally you can mute the entire machine if you don't want noise. I'm not sure it's worth the clutter to add an option for this.

(for reference, I don't think any other VNC client has an option to disable this)

kanaka:
You have obviously never been working in the same room with a baby sleeping in a crib next to you before and had your computer make an unexpected noise and scrambled trying to find what application made the sound while simultaneously searching for the tiny mute button among the function keys, and then forgetting you're on a Mac and the operation of the Fn key is reversed and then once you figure that out you accidentally press the mute button once instead of twice. Not from personal experience of course ;-) Anyways, all that to say, sometimes you want redundancy. 

The client situation is somewhat mixed. RealVNC and ssvnc both have the ability in the client to disable the bell. Tight doesn't seem to, but people complain about it. x11vnc has an easy way to disable it from the server side.

Anyways, doesn't need to be a UI option necessarily, if cluttering the UI is the concern. I think having an RFB option would suffice (with standard ability to specify via URL). Also, that way, when you get 10 openstack administrators file issues because their OpenStack UI is suddenly making sounds at them (because another admin upgraded Horizon dashboard) and waking their 3 month old child, you'll have an easy answer ;-)

Anyways, that's just my 2 cents. I don't feel strongly about it (kids are older now), so if you feel strongly that it should not be added, that's fine with me too.

samhed:
@kanaka that made me smile :)

CendioOssman:
Updated with a query string setting.

DirectXMan12:
I'd prefer a UI option long-term, but I think query string is fine for now.  LGTM!

Issue #666, opened 2016-09-20
client gets broken keycodes


jrosenboom:
See https://bugs.launchpad.net/horizon/+bug/1622684 for a detailed description, result is messages like

```
[ 41.750245] atkbd serio0: Unknown key released (translated set 2,
code 0x0 on isa0060/serio0).
[ 41.750591] atkbd serio0: Use 'setkeycodes 00 <keycode>' to make it known.
```

being shown on the console. Reverting to 4e0c36dda708628836dc6f5d68fc40d05c7716d9 solves the issue.

samhed:
@devicenull seems to have the same issue in #596 

I'm quoting @danielhb here:

> If you are using the qemu keyboard option when starting the guest, remove it

SilviaFc:
What if I'm using devstack to install openstack? How to revert the version?
Thanks

devicenull:
I haven't actually had time to go and gather debug logs for you.  I can tell you we do not have a keyboard/keymap option present on the qemu command line

danielhb:
I will go ahead and try installing Openstack Horizon to see if I can reproduce it here. Meanwhile any other information (operational system, browser version) is welcome. This info is missing from the original launchpad bug.

devicenull:
If it helps, I can provide access (https URL, user, pw) to a websockify session that's having this issue.  I can also provide SSH access to the VM itself.  I cannot provide access to the host though.

danielhb:
On 09/20/2016 01:18 PM, devicenull wrote:
>
> If it helps, I can provide access (https URL, user, pw) to a 
> websockify session that's having this issue. I can also provide SSH 
> access to the VM itself. I cannot provide access to the host though.
>

Feel free to send me the credentials for this websockify session that is 
experiencing the issue. I'll have a look as soon as I am able. It will 
be faster than installing Horizon and hopefully will be enough.

devicenull:
I sent you an email with login details

danielhb:
My first findings are that the remote QEMU VNC server is expecting a field (keysym) that the code isn't supplying. I can see that because the guest is receiving the numpad keys with NumLock ON as expected, the only cases where the keysym is sent. However, we can't make the code send the keysym every time -  this is exactly the problem we're trying to avoid with the extension.

[...]

PR #667, opened 2016-09-21
Add allowQEMUExtKeyEvent option, which allows the qemu keyboard exten…

devicenull:
Add allowQEMUExtKeyEvent option, which allows the qemu keyboard extensions to be disabled.  This is required for compatability with instances that are still running with a '-k' option on the command line.

Setting this to false would fix the issues reported in #666

### DIFF
--- a/vnc_auto.html
+++ b/vnc_auto.html
@@ -233,10 +233,16 @@
                                'local_cursor': WebUtil.getConfigVar('cursor', true),
                                'shared':       WebUtil.getConfigVar('shared', true),
                                'view_only':    WebUtil.getConfigVar('view_only', false),
+                               // If you have qemu guests that are still running with the '-k en-us' (or any other keymap), you'll need
+                               // to set this to false, or each keypress will result in messages like:
+                               // Unknown key released (translated set 2, code 0x0 on isa0060/serio0).
+                               'allowQEMUExtKeyEvent': true,
+
                                'onUpdateState':  updateState,
                                'onXvpInit':    xvpInit,
                                'onPasswordRequired':  passwordRequired,
-                               'onFBUComplete': FBUComplete});
+                               'onFBUComplete': FBUComplete,
+                });
             } catch (exc) {
                 updateState(null, 'fatal', null, 'Unable to create RFB client -- ' + exc);
                 return; // don't continue trying to connect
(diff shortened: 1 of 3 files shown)


vnc_auto.html
```diff
@@ -236,7 +236,11 @@
                                'onUpdateState':  updateState,
                                'onXvpInit':    xvpInit,
                                'onPasswordRequired':  passwordRequired,
-                               'onFBUComplete': FBUComplete});
+                               'onFBUComplete': FBUComplete,
+                               // If you have qemu guests that are still running with the '-k en-us' (or any other keymap), you'll need
+                               // to set this to false, or each keypress will result in messages like:
+                               // Unknown key released (translated set 2, code 0x0 on isa0060/serio0).
+                               'allowQEMUExtKeyEvent': true});
```
samhed: Could you add this for ui.js as well?

devicenull:
Updated

berrange:
This kind of hack should *NOT* be needed. The Extended keyboard event message contains both the scancode and keysym. The keysym sent should be exactly the same as the keysym sent with the non-extended keyboard event message. If QEMU has a keymap set, it will then ignore the scancode and only use the keysym.

If this is not working, then it suggests that the keysym being included in the extended keyboard event message is somehow different from the keysym being sent in the non-extended keyboard event message.

This bit of code in core/rfb.js looks pretty suspect

        _handleKeyPress: function (keyevent) {
            if (this._view_only) { return; } // View only, skip keyboard, events

            var down = (keyevent.type == 'keydown');
            if (this._qemuExtKeyEventSupported) {
                var scancode = XtScancode[keyevent.code];
                if (scancode) {
                    var keysym = keyevent.keysym;
                    RFB.messages.QEMUExtendedKeyEvent(this._sock, keysym, down, scancode);
                } else {
                    Util.Error('Unable to find a xt scancode for code = ' + keyevent.code);
                }
            } else {
                keysym = keyevent.keysym.keysym;
                RFB.messages.keyEvent(this._sock, keysym, down);
            }
        },


Notice that in the extended event scenario it is sending "keyevent.keysym" but in the non-extended scenario it is sending "keyevent.keysym.keysym". Surely those should both be keyevent.keysym.keysym

DirectXMan12:
@berrange this would be a short-term fix until we can integrate the two bits of handling logic (as per https://github.com/kanaka/noVNC/issues/666#issuecomment-249230647).  If we don't get that long-term fix before 0.7.0, we should merge this (with a UI option).

samhed:
@devicenull could you add a UI option for this?

CendioOssman:
No longer needed as the bug in the keyboard code should now be fixed.

PR #668, opened 2016-09-22
Improve rendering/processing delays

CendioOssman:
This resolves some rendering delays that we were seeing. Three issues resolved:

 - setTimeout(foo, 0) does not trigger right away. The spec mandates >4 ms, and some browsers have more. This delay is a practical problem for some small messages (like Fence).

 - requestAnimationFrame() triggers at 17 ms or worse. There is a proper event for waiting on an image to load.

 - Firefox has excessive delays in rendering images[1]. To avoid this and similar issues to create a massive rendering queue, limit the queue to one update.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1304391

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -746,10 +746,16 @@
             this._renderQ.push(action);
             if (this._renderQ.length === 1) {
                 // If this can be rendered immediately it will be, otherwise
-                // the scanner will start polling the queue (every
-                // requestAnimationFrame interval)
+                // the scanner will wait for the relevant event
                 this._scan_renderQ();
             }
+        },
+
+        _resume_renderQ: function() {
+            // "this" is the object that is ready, not the
+            // display object
+            this.removeEventListener('load', this._noVNC_display._resume_renderQ);
+            this._noVNC_display._scan_renderQ();
         },
 
         _scan_renderQ: function () {
@@ -776,6 +782,8 @@
                         if (a.img.complete) {
                             this.drawImage(a.img, a.x, a.y);
                         } else {
+                            a.img._noVNC_display = this;
+                            a.img.addEventListener('load', this._resume_renderQ);
                             // We need to wait for this image to 'load'
                             // to keep things in-order
                             ready = false;
@@ -786,10 +794,6 @@
                 if (ready) {
                     this._renderQ.shift();
                 }
-            }
-
-            if (this._renderQ.length > 0) {
-                requestAnimationFrame(this._scan_renderQ.bind(this));
             }
         },
     };
--- a/core/display.js
+++ b/core/display.js
@@ -365,7 +365,7 @@
 
         fillRect: function (x, y, width, height, color, from_queue) {
             if (this._renderQ.length !== 0 && !from_queue) {
-                this.renderQ_push({
+                this._renderQ_push({
                     'type': 'fill',
                     'x': x,
                     'y': y,
@@ -381,7 +381,7 @@
 
         copyImage: function (old_x, old_y, new_x, new_y, w, h, from_queue) {
             if (this._renderQ.length !== 0 && !from_queue) {
-                this.renderQ_push({
+                this._renderQ_push({
                     'type': 'copy',
                     'old_x': old_x,
                     'old_y': old_y,
@@ -398,6 +398,17 @@
 
                 this._drawCtx.drawImage(this._target, x1, y1, w, h, x2, y2, w, h);
             }
+        },
+
+        imageRect: function(x, y, mime, arr) {
+            var img = new Image();
+            img.src = "data: " + mime + ";base64," + Base64.encode(arr);
+            this._renderQ_push({
+                'type': 'img',
+                'img': img,
+                'x': x,
+                'y': y
+            });
         },
 
         // start updating a tile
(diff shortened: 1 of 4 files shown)


CendioOssman:
Comments? Good to merge?

core/rfb.js
```diff
@@ -548,19 +542,7 @@
...
-                            Util.Debug("More data to process, creating timer");
-                            this._msgTimer = setTimeout(function () {
-                                this._msgTimer = null;
-                                this._handle_message();
-                            }.bind(this), 0);
-                        } else {
-                            Util.Debug("More data to process, existing timer");
-                        }
-                    }
+                    while (this._normal_msg() && this._sock.rQlen() > 0);
```
DirectXMan12: Hmm... I like this in principle, but I'm worried about getting a large number of messages and then causing the browser to block.  I'd like @kanaka's thoughts on this one, but I think I'm tentatively ok with it.

core/display.js
```diff
@@ -741,6 +742,16 @@
             this._drawCtx.putImageData(img, x - vx, y - vy);
         },
 
+        _renderQ_push: function (action) {
+            this._renderQ.push(action);
```
DirectXMan12: I wonder if it would be more efficient here to process immediately instead of pushing and then popping if the render queue is empty?  That might take a bit of rewriting.  Just a thought for later.

core/display.js
```diff
@@ -400,6 +400,17 @@
             }
         },
 
+        imageRect: function(x, y, mime, arr) {
+            var img = new Image();
```
DirectXMan12: :+1:

core/display.js
```diff
@@ -746,12 +746,18 @@
...
                 // If this can be rendered immediately it will be, otherwise
-                // the scanner will start polling the queue (every
-                // requestAnimationFrame interval)
+                // the scanner will wait for the relevant event
                 this._scan_renderQ();
             }
         },
 
+        _resume_renderQ: function() {
+            // "this" is the object that is ready, not the
```
DirectXMan12: is there a significant performance penalty to using bind here instead (there might have been at some point -- it's worth investigating)?  Using bind would probably make this code a bit clearer (not that the comment is bad -- it just looks weird).

tests/test.display.js
```diff
@@ -407,7 +402,7 @@ describe('Display/Canvas Helper', function () {
         });
 
         it('should wait until an image is loaded to attempt to draw it and the rest of the queue', function () {
-            var img = { complete: false };
+            var img = { complete: false, addEventListener: sinon.spy() }
```
DirectXMan12: do we want to test this with an actual image?  IIRC, sinon has ways to test asynchronous tests, and I'd rather test the actual code path here.

core/display.js
```diff
@@ -746,12 +746,18 @@
             this._renderQ.push(action);
             if (this._renderQ.length === 1) {
                 // If this can be rendered immediately it will be, otherwise
-                // the scanner will start polling the queue (every
-                // requestAnimationFrame interval)
+                // the scanner will wait for the relevant event
```
DirectXMan12: :+1:

core/rfb.js
```diff
@@ -1203,6 +1223,14 @@
                     now = (new Date()).getTime();
                     Util.Info("First FBU latency: " + (now - this._timing.fbu_rt_start));
                 }
+
+                // Make sure the previous frame is fully rendered first
+                // to avoid building up an excessive queue
+                if (this._display.pending()) {
+                    this._flushing = true;
```
DirectXMan12: you'd probably need to rewind here, since you've already consumed 3 bytes worth of information (unless I'm misreading something).

core/rfb.js
```diff
@@ -542,7 +544,17 @@
                     Util.Error("Got data while disconnected");
                     break;
                 case 'normal':
-                    while (this._normal_msg() && this._sock.rQlen() > 0);
+                    while (true) {
+                        if (this._flushing) {
+                            break;
+                        }
+                        if (!this._normal_msg()) {
```
DirectXMan12: so, here we're replacing a long render queue for a bigger receive queue.  That's fine *until* we hit the point at which we need to resize the receive queue, at which point we take a performance hit (especially if we have to resize the receive queue multiple times -- allocating large typed arrays is fairly costly).  It's mitigated by the fact that we never downsize our arrays, but I'm still concerned, and I'd like to see some numbers on this change from different VNC sessions, with different servers (at least qemu's VNC server and TigerVNC).

CendioOssman: > so, here we're replacing a long render queue for a bigger receive queue.

That is the point of the change. The server can detect this and back off. It cannot, however, detect how long our rendering queue is

core/rfb.js
CendioOssman: In theory, yes. But we're already processing as much of the FBU as we can in one go, and that's where most of the processing is done. So if it was a problem then we would already have seen it. IMO.

core/rfb.js
CendioOssman: Actually no, because we've already put those three bytes in other variables at this point.

But ideally I'd like to put this code higher, but there isn't a rewind function in the socket object AFAICT.

core/display.js
CendioOssman: Probably, but this was the existing behaviour so I kept it around.

tests/test.display.js
CendioOssman: That would be nice, but I'm not sure how to robustly make all browsers delay loading.

samhed:
I'll merge this next week unless someone objects.

samhed:
Rebased and merged.

One easy way to feel the improvements is to disable local cursor and try before and after. Especially noticeable on Firefox and IE/Edge.

PR #670, opened 2016-09-23
Add URL parameter for automatic reconnects

tyll:
Sometimes the connection just fails and users need to reconnecct. When
the URL parameter reconnect=true is set, noVNC reconnects automatically.
The delay can be set with the URL parameter reconnect_delay.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -56,6 +56,9 @@ var UI;
         ctrlOn: false,
         altOn: false,
 
+        reconnect: false,
+        reconnect_delay: 500,
+
         // Setup rfb object, load settings from browser storage, then call
         // UI.init to setup the UI/menus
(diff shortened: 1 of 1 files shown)


DirectXMan12:
Thanks for the PR!  Not sure how I feel about this being generally applicable to noVNC, but I'm welcome to arguments to convince me otherwise ;-)

It will need to be updated to fit with the latest UI changes, though (should be just a bit of refactoring)

@samhed thoughs?

samhed:
Yeah.. I agree with you here. It feels like the problem of the disconnections should be solved instead.

DirectXMan12:
Ok, I'm going to close this for the moment.  If @tyll comes up with a super-compelling argument, I'll reopen.  Thanks for the PR!

PR #687, opened 2016-10-14
Improve info and error icons

CendioOssman:
Give them a solid background to make them more easily identifiable
as icons, rather than text or some other more inline element.

A bit of bike shedding, but I think this is a worthwhile improvement. You now get easily recognisable shapes for the messages: round for info, triangle for warning, square for error.


samhed:
Looks good!

PR #689, opened 2016-10-15
Don't send Unicode as keysym

ossman:
We had a bug where we would send a Unicode codepoint as a keysym to the server. This of course ended up with a completely wrong symbol.

Also made sure we can transfer the entire Unicode range.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -14,6 +14,7 @@
 /* [module]
  * import Util from "../core/util";
  * import KeyTable from "../core/input/keysym";
+ * import keysyms from "./keysymdef";
  * import RFB from "../core/rfb";
  * import Display from "../core/display";
  * import WebUtil from "./webutil";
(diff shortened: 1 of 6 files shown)


samhed:
The changes to test.helper.js seem to cause at least the Travis-tests to fail, do they work for you?

samhed:
Do you have a specific example of something that was fixed by this?

ossman:
The tests were buggy and I fixed them as part of this. A result of this is that it requires ECMAscript 6. It works fine here, but your Travis instance doesn't seem to handle it. Perhaps the test needs to be changed to something simpler. 

As for what was fixed, a lot of characters on virtual keyboards were wrong. One easy example are smileys.

samhed:
@DirectXMan12 can we update Node JS on our travis?

DirectXMan12:
> can we update Node JS on our travis?

Yeah, that shouldn't be a problem.

samhed:
I don't seem to be able to access settings for travis

DirectXMan12:
@samhed you should be able to change the version of node.js via .travis.yml.  The actual issue here probably PhantomJS, though.  I'll have to figure out what version of PhantomJS the karma launcher is pulling.

PR #696, opened 2016-10-21
Support standard mouse wheel event

CendioOssman:
Mouse wheel event handling has now been standardised and has broad
support. Use this event rather than the older, non-standard ones.
At the same time fix up support for horisontal mouse wheel events.

Tested on Windows, Linux and macOS.

### DIFF
--- a/core/input/devices.js
+++ b/core/input/devices.js
@@ -291,18 +291,25 @@
 
             var evt = (e ? e : window.event);
             var pos = Util.getEventPosition(e, this._target, this._scale);
-            var wheelData = evt.detail ? evt.detail * -1 : evt.wheelDelta / 40;
-            var bmask;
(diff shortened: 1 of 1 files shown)


samhed:
nice!

CendioOssman:
caniuse doesn't include this, so I only know what I've tested unfortunately.

CendioOssman:
Addendum: MDN tracks support for this:

https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent

In short, no raised requirements by this.

DirectXMan12:
nice!  Fine by me to merge, then.

nagexiucai:
yew, raw mouse and keyboard events are supported now.
how to get those higher events during operation within vnc.
thus i can monitor user's action on frontend for deep learning.

PR #697, opened 2016-10-21
Better method to keep focus on virtual keyboard

CendioOssman:
This prevents the focus being lost rather than trying to restore it
at various places.

Tested on the major browsers on iOS and Android.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -49,7 +49,6 @@ var UI;
         controlbarDrag: false,
         controlbarMouseDownClientY: 0,
         controlbarMouseDownOffsetY: 0,
-        keyboardVisible: false,
 
         isSafari: false,
         rememberedClipSetting: null,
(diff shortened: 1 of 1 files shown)


samhed:
I have tested on Android 7.0.1 and iOS 10 and can't find any regressions there. However, when testing on a _**Microsoft Surface device**_ with Windows 10 I do find differences in behavior with regards to opening and closing the virtual keyboard on Edge..

---
**master on Edge:**
* Can never bring up virtual keyboard using noVNC's keyboard button
* Can never close virtual keyboard using noVNC's keyboard button

**this PR on Edge:**
* Can almost always bring up virtual keyboard using noVNC's keyboard button in Windows Tablet-mode
* Can't bring up virtual keyboard using noVNC's keyboard button when not in Windows Tablet-mode
* Can never close virtual keyboard using noVNC's keyboard button
* If the virtual keyboard was brought up using noVNC's keyboard button, and closed using the 'X'-button on the keyboard, any click in the VNC session will bring the keyboard back up.
* The only way to "properly" close the virtual keyboard is to close it using the 'X'-button on the keyboard and after that click somewhere outside any button in the control bar.
* It's easy to get noVNC's keyboard button out of sync.

---
Note that we aren't detecting touch in Firefox on this platform. And note that the behavior with this patch remains unchanged on IE and Chrome.

samhed:
In the current master, noVNC's keyboard button does nothing in Edge, but at least it is consistent. The behavior of Edge with these changes is really annoying. Do we have any possible explanation to why it is like this?

CendioOssman:
I'm note sure what Windows keyboard handling is up to. It seems to be a general issue though as I'm seeing random behaviour with other input fields as well, and in Chrome. I couldn't find any pattern or workaround. No hits on Google either. :/

Still, I think this PR makes things better on the whole. A standard Windows configuration has no keyboard button, so we should provide one. And I think most users will find the "X" to close the keyboard even if our closing is unreliable.

samhed:
Yeah.. Agreed. Unless @kanaka or @DirectXMan12 has any objections. I'll merge this next week.

Issue #518, opened 2015-08-20
Double buffering


samhed:
Currently we don't buffer the entire framebuffer in noVNC, the only 'buffer' except for the receive queue we have is the canvas itself, the same canvas we show the user. Additionally, the canvas only contains what is shown in the viewport. In the case of clipping there are parts of the screen which isn't stored client-side.

I would like to discuss the possibilities of a double buffering architecture in noVNC. One buffer which would contain the entire remote framebuffer and the other being the canvas which contains the data  shown in the viewport. It can be worth noting that most native VNC clients use double buffering.

I see the following advantages with double buffering:
- Avoid screen-tearing since we can draw the entire on-screen canvas at the same time through copying the contents from the buffer ( #482 )
- If the viewport is smaller than the remote frame buffer we can resize the viewport or use panning (if clipping) and see the new areas without having to wait for an update from the server.
- We can get more robust code and avoid bugs like #517. It can also make features like zoom easier to implement.

The disadvantages I can come to think of are:
- Rendering will be slightly slower.
- It will use a bit more memory on the client.
- In the case of a smaller viewport, it could use more bandwidth since we could get updates for things happening outside the viewport which we might or might not draw.

---

I'm thinking a second canvas which is off-screen. Canvas context operations like drawImage and getImageData can be used with another canvas as the source. This way we can use the same operations and the restructure doesn't have to be as comprehensive.

Now when @DirectXMan12 has merged the performance improvements from #488, do you think we have enough leeway to handle the hit to the rendering speed?

DirectXMan12:
> I'm thinking a second canvas which is off-screen. Canvas context operations like drawImage and getImageData can be used with another canvas as the source. This way we can use the same operations and the restructure doesn't have to be as comprehensive.

This sounds like the way to do it.  I wonder if there's any performance change due to offscreen or onscreen canvas use.  I think the speed of this really depends on how good the browsers are at redrawing due hiding/showing the canvas.  Can you whip up a prototype so that we can take a look?

samhed:
Will do, I have some other stuff I'm working on at the moment, but this shouldn't take too long so I hope to get to it soon enough.

samhed:
Merged in #699

Issue #517, opened 2015-08-20
Sometimes we don't get a framebuffer update after panning


samhed:
When clipping and panning we require new data from the server to draw the new parts of the screen. This seems to be working very inconsistently. The result is that you have blank parts of the screen until you do something else which will trigger the update.

It is a lot easier to reproduce on touch devices, where you don't have a mouse cursor. Disabling "Local cursor" can also make it easier to reproduce.

DirectXMan12:
Did you notice this before or after the performance code merge?  I would like to make sure that I didn't cause any issues.

samhed:
Before. I don't think your changes affected this. I have seen the issue for quite a while, just never got to reporting it.

samhed:
Yeah, just verified briefly, the behavior is still the same, no difference from your changes.

samhed:
Fixed now seeing as #699 is merged.

Issue #482, opened 2015-05-01
Tearing on high res vnc session


blauwepiet:
Hello,

I've been testing multiple web-based vnc clients and noVNC looks very promising. It is fast and can handle 1920x1080 resolution with a decent framerate. Although I've noticed some screen tearing both horizontal and vertical. It is more prominent in firefox compared to chrome. Is there a way to reduce this tearing?

I don't know how the frames are drawn but it looks like it is done in chunks or blocks. I think the tearing is caused by the fact that the last frame is still present and a block is already drawn on this old frame. So what I thought: Isn't it possible to increase the draw block size? In this case the block will encapsulate the whole moving part of the display and will only "cut" the none moving parts where tearing isn't visible.

These are just some wild thoughts, but maybe they can be of some use. 

To recap my main question is how can I reduce the screen tearing? Increase the draw speed somehow?

Thanks in advance!

DirectXMan12:
One of the major factors is what encoding the VNC server is using, and how big the chunks are -- noVNC tends to handle a few big chunks in png or jpeg better than lots of small chunks in RRE form, for instance.  See #431 , for instance (unfortunately, I haven't had the time to replace noVNC's zlib implementation, but, depending on your setup, that may make a difference).

blauwepiet:
The vnc server I am using is TurboVNC and this already uses tight encoding, which is the one with the big chunks. I've verified it by looking in the rfb object in the UI. Now I was wondering if you could change the compression level in the rfb object, because I couldn't find a compression level setting?

blauwepiet:
Nevermind, I found the compression setting, but it is still not what I want :/.. The jpeg quality trade off against the tearing. 

It's just when I use the vncviewer of Turbovnc itself it runs perfect but whenever I use it with noVNC there is tearing. But that might have something to do with an optimized protocol between the TurboVNC server and viewer, because that protocol is slightly different than the tight protocol.

surfsara-visualization:
> One of the major factors is what encoding the VNC server is using, 
> and how big the chunks are -- noVNC tends to handle a few big 
> chunks in png or jpeg better than lots of small chunks in RRE form, 
> for instance. 

[...]

PR #699, opened 2016-10-25
Display double buffering

CendioOssman:
Do all rendering to a hidden canvas and then copy over the finished frame to the visible canvas once everything is done. This simplifies things and solves some bugs as we can retain a copy of the entire
frame buffer.

Tested on Chrome and Firefox. Performance is roughly the same after these changes, so no gain there unfortunately. But it makes panning behave very nicely and it allows for more fun stuff with panning and zooming in the future.

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -26,9 +26,6 @@
     this._fb_width = 0;
     this._fb_height = 0;
 
-    // the visible "physical canvas" viewport
-    this._viewportLoc = { 'x': 0, 'y': 0, 'w': 0, 'h': 0 };
-
     this._prevDrawStyle = "";
     this._tile = null;
     this._tile16x16 = null;
@@ -60,6 +57,9 @@
     }
 
     this._targetCtx = this._target.getContext('2d');
+
+    // the visible canvas viewport (i.e. what actually gets seen)
+    this._viewportLoc = { 'x': 0, 'y': 0, 'w': this._target.width, 'h': this._target.height };
 
     // The hidden canvas, where we do the actual rendering
     this._backbuffer = document.createElement('canvas');
@@ -156,10 +156,19 @@
 
         viewportChangeSize: function(width, height) {
 
-            if (typeof(width) === "undefined" || typeof(height) === "undefined") {
+            if (!this._viewport ||
+                typeof(width) === "undefined" ||
+                typeof(height) === "undefined") {
 
                 Util.Debug("Setting viewport to full display region");
                 width = this._fb_width;
+                height = this._fb_height;
+            }
+
+            if (width > this._fb_width) {
+                width = this._fb_width;
+            }
+            if (height > this._fb_height) {
                 height = this._fb_height;
             }
 
@@ -169,18 +178,17 @@
                 vp.h = height;
 
                 var canvas = this._target;
-                if (canvas.width !== width || canvas.height !== height) {
-                    if (canvas.width !== width) {
-                        canvas.width = width;
-                        canvas.style.width = width + 'px';
-                    }
-                    if (canvas.height !== height) {
-                        canvas.height = height;
-                        canvas.style.height = height + 'px';
-                    }
-                    this._damage(vp.x, vp.y, vp.w, vp.h);
-                    this.flip();
-                }
+                canvas.width = width;
+                canvas.height = height;
+
+                // The position might need to be updated if we've grown
+                this.viewportChangePos(0, 0);
+
+                this._damage(vp.x, vp.y, vp.w, vp.h);
+                this.flip();
+
+                // Update the visible size of the target canvas
+                this._rescale(this._scale);
             }
         },
 
@@ -219,9 +227,11 @@
                 }
             }
 
-            this._rescale(this._scale);
-
-            this.viewportChangeSize();
+            // Readjust the viewport as it may be incorrectly sized
+            // and positioned
+            var vp = this._viewportLoc;
+            this.viewportChangeSize(vp.w, vp.h);
+            this.viewportChangePos(0, 0);
         },
 
         // Track what parts of the visible canvas that need updating
@@ -545,6 +555,14 @@
         // Overridden getters/setters
         set_scale: function (scale) {
             this._rescale(scale);
+        },
+
+        set_viewport: function (viewport) {
+            this._viewport = viewport;
+            // May need to readjust the viewport dimensions
+            var vp = this._viewportLoc;
+            this.viewportChangeSize(vp.w, vp.h);
+            this.viewportChangePos(0, 0);
         },
 
         get_width: function () {
--- a/core/display.js
+++ b/core/display.js
@@ -32,7 +32,6 @@
 
     // the visible "physical canvas" viewport
     this._viewportLoc = { 'x': 0, 'y': 0, 'w': 0, 'h': 0 };
-    this._cleanRect = { 'x1': 0, 'y1': 0, 'x2': -1, 'y2': -1 };
 
     this._prevDrawStyle = "";
     this._tile = null;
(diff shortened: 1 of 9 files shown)


samhed:
Yay!

This should fix at least #482 and #517, and see #518 for earlier discussions. It's great that it doesn't seem slower!

CendioOssman:
I need to have a look at the tests. Hopefully before the end of the week.

DirectXMan12:
Can't wait to review this!  You've been a powerhouse of pull requests recently :-).  I'd like to run some numbers, and at some point, I'd like to put together a suite of noVNC recordings from different servers to use for benchmarking (for example, Qemu's server has highly different performance characteristics than TigerVNC, since it sends data differently).

CendioOssman:
Have you had time to have a look?

CendioOssman:
I found a bunch of issues in the viewport handling that was easier to sort out with this in place, so I've added those commits as well.

samhed:
Thanks for cleaning up the viewport code as well! It looks great.

@DirectXMan12 will you have time to run your numbers? I think we should get this merged

DirectXMan12:
Sorry for the delay -- had a conference last week and was super-busy before that.  I'll try and take a look this week.

samhed:
As far as merging goes, I'm satisfied with the testing @CendioOssman has done here. Do you have any specific concerns @DirectXMan12? Otherwise I'll go ahead and get this merged :)

DirectXMan12:
I'd like to also take a run-through.  I'm on vacation this week (it's Thanksgiving week in the US), so I may or may not get to it this week.  If not, I'll try for early next week.

CendioOssman:
> The canvas isn't resized in this case. Easy ways to trigger the resize is to decrease the size of the browser window, or opening and closing the settings panel.

Found the issue. We weren't properly updating all UI state when the server resized the framebuffer. Should be fixed now.

tests/playback.js
```diff
@@ -117,6 +129,16 @@ queue_next_packet = function () {
 var bytes_processed = 0;
 
 do_packet = function () {
+    // Avoid having an excessive queue buildup
+    if (rfb._flushing) {
```
DirectXMan12: unless I'm misreading this, I don't think we should be doing anything special in our performance testing code.  If the normal behavior is insufficient, the performance code should exhibit the same behavior, not work around it.

core/util.js
```diff
@@ -241,6 +241,47 @@ Util.stopEvent = function (e) {
     e.preventDefault();
 };
 
+// Immediate polyfill
+Util.setImmediate = function (func) {
```
CendioOssman: I agree that something on a standards track would be more interesting. Unfortunately no such thing exists. :/

We were having similar issues in the generic code at one point, so it ended up in Util. Fortunately we solved those issues another way, so we could move this function if you insist.

tests/playback.js
CendioOssman: The performance test is not doing normal stuff. It is flooding the code with updates without any form of flow control. :)
This chunk is to have some semblance of rate control to get it a bit more closer to the behaviour a normal connection would see. Normally this code would be in the VNC server, so it makes sense that we put it in the performance test (which has to emulate the server after all).

DirectXMan12: except, out captures *already* have time information in them.  We *already* have an option to match exactly what the server does, which pulls the timing info from the recording and submits the frames as they came in.  We don't want the code here messing with that.

core/util.js
DirectXMan12: yeah, I'd prefer moving it to the tests directory.

tests/playback.js
CendioOssman: But that's just the playback mode, not the performance mode. The only metric we have is total time, which means we have to have this code or else we're only measuring the time it takes to move buffers around.

core/display.js
```diff
@@ -405,12 +297,21 @@
...
-                var y1 = old_y - this._viewportLoc.y;
-                var x2 = new_x - this._viewportLoc.x;
-                var y2 = new_y - this._viewportLoc.y;
+                // Due to this bug among others [1] we need to disable the image-smoothing to
+                // avoid getting a blur effect when copying data.
+                //
+                // 1. https://bugzilla.mozilla.org/show_bug.cgi?id=1194719
+                //
+                // We need to set these every time since all properties are reset
+                // when the the size is changed
```
CendioOssman: In theory, yes. However I couldn't see any issue so I left it out. It may be that the problem is very small and only noticeable when you keep copying the same data many times.

core/rfb.js
```diff
@@ -1198,11 +1198,9 @@
             switch (msg_type) {
                 case 0:  // FramebufferUpdate
                     var ret = this._framebufferUpdate();
-                    if (ret) {
-                        RFB.messages.fbUpdateRequests(this._sock,
-                                                      this._enabledContinuousUpdates,
-                                                      this._display.getCleanDirtyReset(),
-                                                      this._fb_width, this._fb_height);
+                    if (ret && !this._enabledContinuousUpdates) {
+                        RFB.messages.fbUpdateRequest(this._sock, 1, 0, 0,
```
CendioOssman: Indeed. But it was already this way so I figured I'd leave it. I can change it if you want. :)

core/rfb.js
```diff
@@ -1346,6 +1344,8 @@
                 if (!ret) { return ret; }  // need more data
             }
 
+            this._display.flip();
```
CendioOssman: It was easier this way. Otherwise we need triple buffering to avoid tearing. Something to experiment with in the future perhaps.

CendioOssman:
That should take care of the requested changes.

core/rfb.js
DirectXMan12: ack, alright

core/display.js
DirectXMan12: I'd add a comment in flip about it, then, just in case we have an issue in the future

tests/playback.js
CendioOssman: Precisely. And without this code we're failing to test that.

What happens instead is that we keep feeding the RFB code new buffers. But it is waiting for the browser, so all it can do is queue up things. So after a short while we're mostly testing the speed of buffer management as the buffers grow to unrealistic proportions.

A real server would either notice the lack of framebuffer update requests, or lack of fence responses, and hold off on sending more updates. So the performance code needs something similar.

core/display.js
CendioOssman: Sure

tests/playback.js
DirectXMan12: Yep, woke up in the middle of the night last night and realized what you were saying :-P.  I'd still prefer if we could only turn it on "performance" mode though, since I use recordings to debug issues in the realitime mode, and sometimes timing matters there.

Honestly, though, sometimes checking buffer management is what I want to do with the performance test, though.  Could we maybe just have it behind an option (I've used this code to fix GC-related performance issues in our buffer management, for instance)?

CendioOssman: > Yep, woke up in the middle of the night last night and realized what you were saying :-P. I'd still prefer if we could only turn it on "performance" mode though, since I use recordings to debug issues in the realitime mode, and sometimes timing matters there.
> 

I suppose it's possible. But I wouldn't have too much faith in the timing system. The browsers do not respect the timeout value given.

> Honestly, though, sometimes checking buffer management is what I want to do with the performance test. Could we maybe just have it behind an option (I've used this code to fix GC-related performance issues in our buffer management, for instance)?

That sounds like something that a more specific unit test is better suited for. This is a "whole system" after all. You never quite know what parts might be included.

DirectXMan12:
LGTM :+1:.  Anyone have any last comments?  If not, I'll merge on Monday.

samhed:
I'll go a head and merge :)

PR #701, opened 2016-10-27
Localization support added for messages

LoekJanssen:
I created localization in German, Dutch and English for the messages in
the top messagebar and some logs.

I've used full English strings for lookup.

This is the new version of PR #548, because I made a few mistakes.

### DIFF
--- a/app/locale/locale-de-DE.js
+++ b/app/locale/locale-de-DE.js
@@ -1,0 +1,13 @@
+Language = {
+	"Connected (encrypted) to ": "Verbunden mit (verschlüsselt) ",
+	"Connected (unencrypted) to ": "Verbunden mit (unverschlüsselt) ",
+	"Must set host and port": "Richten Sie Host und Port ein",
+	"Disconnect timeout": "Timeout beim trennen",
+	"Password is required": "Passwort ist erforderlich",
+	"Forcing clipping mode since scrollbars aren't supported by IE in fullscreen": "'Clipping-Modus' aktiviert, Scrollbalken in 'IE-Vollbildmodus' werden nicht unterstützt",
+	"Connecting": "Verbunden",
+	"Disconnecting": "Verbindung trennen",
+	"Disconnected": "Verbindung zum Server getrennt",
+	"Invalid state": "Falscher Status",
+	"Unable to create RFB client -- ": "RFB Client kann nicht erstellt werden -- ",
+};
--- a/app/ui.js
+++ b/app/ui.js
@@ -27,7 +27,7 @@ var UI;
     /* [begin skip-as-module] */
     // Load supporting scripts
     WebUtil.load_scripts(
-        {'core': ["base64.js", "websock.js", "des.js", "input/keysymdef.js",
+        {'core': [Util.Localisation.getLanguageFileLocation(), "base64.js", "websock.js", "des.js", "input/keysymdef.js",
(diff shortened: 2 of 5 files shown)


samhed:
The commits are very messy, could you clean them up a bit?

LoekJanssen:
@samhed I've done a rebase and put everything in one commit

samhed:
Thank you!

app/locale/locale-de-DE.js
```diff
@@ -0,0 +1,13 @@
+Language = {
+	"Connected (encrypted) to ": "Verbunden mit (verschlüsselt) ",
+	"Connected (unencrypted) to ": "Verbunden mit (unverschlüsselt) ",
+	"Must set host and port": "Richten Sie Host und Port ein",
+	"Disconnect timeout":	"Timeout beim trennen",
```
samhed: Nitpick: Remove the extra spaces after the ":" here please :)

samhed: Same thing for en-GB and nl-NL

app/ui.js
```diff
@@ -475,8 +475,8 @@ var UI;
                     statusElem.classList.add("noVNC_status_normal");
                     break;
             }
-
-            statusElem.innerHTML = text;
+
```
samhed: Remove this whitespace change please

samhed:
I didn't mean you should remove the spaces after ALL the colons :smiley: 

There were extra spaces after the colon on the "Disconnect timeout" lines!

LoekJanssen:
Ah haha, I saw that, but I thought you meant all. I will change it now :)

app/ui.js
```diff
@@ -476,7 +476,7 @@ var UI;
                     break;
             }
 
-            statusElem.innerHTML = text;
+            statusElem.innerHTML = Util.Localisation.get(text);
```
CendioOssman: This is generally not a good idea as it makes it very difficult to find what needs to be translated. Only put translations on literals.

app/ui.js
```diff
@@ -369,9 +369,9 @@ var UI;
                 case 'connected':
                     UI.connected = true;
                     if (rfb && rfb.get_encrypt()) {
-                        msg = "Connected (encrypted) to " + UI.desktopName;
+                        msg = Util.Localisation.get("Connected (encrypted) to ") + UI.desktopName;
```
CendioOssman: A good start perhaps, but this won't work long term. What if the order of words needs to be changed? A printf() style API is need to handle languages outside the Germanic family.

core/util.js
```diff
@@ -366,4 +366,85 @@ Util.Flash = (function () {
...
+        'nl':    'nl-NL',
+        'nl-NL': 'nl-NL',
+        'nl-BE': 'nl-NL',
+        'de':    'de-DE',
+        'de-DE': 'de-DE'
+    },
+
+    // Get language file location
+    getLanguageFileLocation: function () {
+        return '../app/locale/locale-'+Util.Localisation.getLanguageCode()+'.js';
```
CendioOssman: Util should only be used for core API things and should not reference the app side of things.

samhed: You mean that we can't expect everyone to have an 'app' or a 'app/locale' directory I guess.. Good point, didn't think of that!

app/ui.js
samhed: Fixed in fdb02f857083ab1015ccaaa8d0d1be2a978537a3

PR #708, opened 2016-11-05
rfb: Allow port to be optionally appended to URI

bkylerussell:
Supports server configurations that might prefer wss:// connections
on the default port, then proxies them through the web server to the
VNC server.

This proxy configuration is helpful on servers using self-signed
certificates.  Accessing the https://<host>/vnc_auto.html page and
adding an exception for that host is sufficient to also satisfy the
wss://<host>/ request, unlike requests to wss://<host>:<port>/ which
may require an extra exception.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -273,9 +273,9 @@
             this._rfb_password = (password !== undefined) ? password : "";
             this._rfb_path = (path !== undefined) ? path : "";
 
-            if (!this._rfb_host || !this._rfb_port) {
+            if (!this._rfb_host) {
                 return this._fail(
-                    _("Must set host and port"));
+                    _("Must set host"));
             }
 
             this._rfb_init_state = '';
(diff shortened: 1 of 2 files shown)


core/rfb.js
```diff
@@ -369,7 +369,11 @@
                 uri = this._encrypt ? 'wss' : 'ws';
             }
 
-            uri += '://' + this._rfb_host + ':' + this._rfb_port + '/' + this._rfb_path;
+            uri += '://' + this._rfb_host;
+            if(this._rfb_port)
+                uri += ':' + this._rfb_port;
```
DirectXMan12: please add braces -- we try not to use braceless multi-line if in noVNC

bkylerussell: Sorry, I can fix this up.  I had also planned on adjusting the test case so that the build checks passed.

bkylerussell:
Looks like part of the description in my merge request got clobbered.

In a self-signed certificate server configuration, this feature allows the admin to configure proxying the wss request on the web server to the actual vnc server port in order to minimize the number of certificate exceptions the end user has to deal with.

If, for example, I access the launch page at `https://ip/` (and accept a self-signed certificate that my server is configured with), and noVNC issues the wss request to `wss://ip/`, then the previous certificate exception for https is also sufficient for the wss request.  However, if noVNC issues the wss request to `wss://ip:5900/`, then the browser requires a separate certificate exception for `ip:5900` (for example, by navigating to `https://ip:5900/`).  It is inconvenient for the user to have to accept self-signed certificates at both `https://ip/` and `https://ip:5900/`.

bkylerussell:
Let me know if you still have concerns I haven't addressed.  Sorry for the delay in fixing this up.

CendioOssman:
Sorry for this falling between the cracks. Looks very nice. I'll have it merged.

PR #713, opened 2016-11-09
Better error handling

### LINKED ISSUE #678: Disconnect timeout when we can't find a VNC server
Since commit c2a4d3ef86f966580c661272d772a708f0da003d we are no longer properly handling these two cases:
- can't find a VNC server
- can't find a websocket proxy on the specified host and port combination

The proper error message is: "Failed to connect to server"

Instead we are now getting the disconnect timeout.

samhed:
Adds a catch fallback for uncaught errors as well as changing the way RFB._fail() works. This fixes #678 as well.

### DIFF
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -173,6 +173,43 @@ input[type=button]:active, select:active {
 }
 
 /* ----------------------------------------
+ * Fallback error
+ * ----------------------------------------
+ */
+
+#noVNC_fallback_error {
+  position: fixed;
+  z-index: 3;
+  left: 50%;
+  transform: translate(-50%, -50px);
+  transition: 0.5s ease-in-out;
+
+  visibility: hidden;
+  opacity: 0;
+
+  top: 60px;
+  padding: 15px;
+  width: auto;
+
+  text-align: center;
+  font-weight: bold;
+  word-wrap: break-word;
+  color: #fff;
+
+  border-radius: 10px;
+  box-shadow: 6px 6px 0px rgba(0, 0, 0, 0.5);
+  background: rgba(200,55,55,0.8);
+}
+#noVNC_fallback_error.noVNC_open {
+  transform: translate(-50%, 0);
+  visibility: visible;
+  opacity: 1;
+}
+#noVNC_fallback_errormsg {
+  font-weight: normal;
+}
+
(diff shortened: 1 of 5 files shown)


samhed:
I don't think anything in here is controversial really, if someone wants to take a look you should do so soon. Otherwise I'll merge next week.

app/styles/base.css
```diff
@@ -23,6 +23,43 @@ html {
...
+  padding: 15px;
+  width: auto;
+
+  text-align: center;
+  font-weight: bold;
+  word-wrap: break-word;
+  color: #fff;
+
+  -webkit-border-radius: 10px;
+  -moz-border-radius: 10px;
```
CendioOssman: Not necessary with our browser requirements.

app/styles/base.css
```diff
@@ -23,6 +23,43 @@ html {
   height:100%;
 }
 
+#fallback_error {
```
CendioOssman: Should have a `noVNC_` prefix.

app/ui.js
```diff
@@ -24,6 +24,21 @@ var UI;
 (function () {
     "use strict";
 
+    // Fallback for all uncought errors
+    window.onerror = function(msg, url, line) {
```
CendioOssman: Should we use `addEventListener()`?

core/rfb.js
```diff
@@ -231,13 +231,19 @@
                 this._updateConnectionState('disconnected');
                 break;
             case 'connecting':
-                this._fail('Failed to connect to server' + msg);
+                this._fail('Failed to connect to server', msg);
+                break;
+            case 'connected':
+                this._updateConnectionState('disconnecting');
+                this._updateConnectionState('disconnected');
```
CendioOssman: A comment on this sequence perhaps?

vnc.html
```diff
@@ -65,6 +65,12 @@
 </head>
 
 <body>
+
+    <div id="fallback_error">
+        noVNC encountered an error:<br>
```
CendioOssman: Probably cleaner to give this it's own div instead. It's weird mixing inline and block elements.

samhed:
@CendioOssman good points, thanks! Should all be fixed now

PR #715, opened 2016-11-14
Cleanup of settings (amongst other things)

CendioOssman:
This does various cleanups for things that have annoyed me for a while. :)
The primary focus is the settings and connect dialogs, which I felt were needlessly complex for most users.

Give it a try and see what you think.


samhed:
Really nice!

I also like the ideas of removing the extra stylesheets and the token setting.

samhed:
For some reason Chrome decides to do line-breaks in the settings with this new code:

nowrapafter

This is how it looks with master:

nowrapbefore

Should we add this to the css for the panels?

    white-space: nowrap;

samhed:
98400d7 seems to fix an issue where the 'connecting'-transition screen appeared in front of the password dialog.

CendioOssman:
> For some reason Chrome decides to do line-breaks in the settings with this new code:

It always did. It's just that the input field pushed out the width of the dialog so that we didn't see any line breaks with the current elements.

A general `nowrap` is most likely a bad idea.But it could make sense for labels. I'll have a look.

app/ui.js
```diff
@@ -180,7 +180,6 @@ var UI;
             UI.initSetting('view_only', false);
             UI.initSetting('path', 'websockify');
             UI.initSetting('repeaterID', '');
-            UI.initSetting('token', '');
```
DirectXMan12: Hmm... does this prevent token from working as a URL param?  If so, it should still be in there.  If we somehow broke passing a token as part of the URL at some point, then we need to fix that.

CendioOssman: Ah, right. I see now where it was used. However seeing that I have to object to its inclusion. Users who want a special websocket path should merely specify a different `path`. What makes this token so special?

samhed:
I see that you have made the "noVNC" text larger above the connect button in the middle. I think this looks nicer now!

When the browser is very narrow (for example, phones in portrait mode) it looks like this:
connectcleanup

Should perhaps be tweaked a bit? No strong feelings about this one though

samhed:
As far as the discussion regarding token goes, I don't understand why it was added in the first place (back in 2012: 4c75210)? Isn't this a Nova specific thing?

CendioOssman:
I've fixed the text and button so that it scales for narrow screens now.

samhed:
Looks good!

Only the token discussion remaining then imo.

DirectXMan12:
> As far as the discussion regarding token goes, I don't understand why it was added in the first place (back in 2012: 4c75210)? Isn't this a Nova specific thing?

Not really?  It's the only method of doing target selection when using websockify (which many people use noVNC with), so it's a convenience factor for users of websockify.  #537 and #536 are the relevant recent discussion (people seem to expect to be able to pass token in the path).

CendioOssman:
> It's the only method of doing target selection when using websockify (which many people use noVNC with), so it's a convenience factor for users of websockify.

As I mentioned, you can just modify the `path` instead.

> #537 and #536 are the relevant recent discussion (people seem to expect to be able to pass token in the path).

None of these mention what's so special about `token` unfortunately. This is a very slippery road. `token` is in no way standardised. Some other gateways might want `auth`, or `username`, or `password`, or `cookie`, etc. Having an ever increasing list of these things isn't really sensible.

app/ui.js
patrakov: We currently have deployed noVNC on a bastion host, with a local change that makes host and port hidden inputs, and renames the label of the token. Tokens map 1:1 to available VNC ports of internal hosts (via websockify configuration file that lists them all), so that we can run one websockify for all of them.

While I agree that this setting is confusing to users, and should not be exposed by default, keeping the token as a hidden input (so that we can turn it into a regular input and change the label locally) would help us.

CendioOssman: Can't you just change `path` in that case? That's what we do for ThinLinc and it has worked out fine.

CendioOssman:
Can we get this moving forward? What needs to be done to resolve the situation?

samhed:
Good to merge in my opinion

app/ui.js
DirectXMan12: We should have a way to add in URL-only parameters.  We should discuss more in a follow-up PR.  For now, this is fine.

PR #718, opened 2016-11-16
Use gettext .po files, and translate everything

ossman:
Partial translations are worse than no translations IMO. So this sorts out having the HTML translated as well.

It also switches to using gettext files and tools for translations. This makes it much easier to work with translations, and also allows us to have external parties do translations in a much more straight forward way.


samhed:
Really nice!

Is there any reason to why you are keeping the en_GB "translation"?

samhed:
I get the following when loading vnc.html while having my browser set to using English (United States):

> Failed to load resource: the server responded with a status of 404 (File not found) http://localhost:6080/app/locale/en.js

DirectXMan12:
> Is there any reason to why you are keeping the en_GB "translation"?

So that proper British citizens can use the Queen's English, and not have to settle for using the improper form spoken in uppity former colonies! 😜 

Actually, I think we should just have a translation for every language, including the "default" one.

ossman:
All comments should be fixed now.

core/util.js
```diff
@@ -440,7 +440,81 @@ Util.Localisation = {
...
+                if (isAnyOf(elem.getAttribute("translate"), ["", "yes"])) {
+                    enabled = true;
+                } else if (isAnyOf(elem.getAttribute("translate"), ["no"])) {
+                    enabled = false;
+                }
+            }
+
+            if (enabled) {
+                if (elem.hasAttribute("abbr") &&
+                    elem.tagName === "th") {
```
CendioOssman: Damn. Missed one tag that needs to be uppercase. I need to fix this before we merge.

samhed:
I see that you have written a wiki page about the translations as well, nice!

https://github.com/kanaka/noVNC/wiki/Translations

samhed:
Note that the en_GB translation was removed.

PR #720, opened 2016-11-24
Update Greek translation

kosmasgiannis:
I hope it is OK.


po/el.po
```diff
@@ -55,3 +55,242 @@ msgstr ""
 #: ../core/rfb.js:556
 msgid "Disconnect timeout"
 msgstr "Παρέλευση χρονικού ορίου αποσύνδεσης"
+
+#: ../vnc.html:70
+msgid "noVNC encountered an error:"
+msgstr "το noVNC αντιμετώπισε ένα σφάλμα"
```
CendioOssman: Shouldn't there be a `:` here?

po/el.po
```diff
@@ -55,3 +55,242 @@ msgstr ""
...
+msgstr "Καταγραφή:"
+
+#: ../vnc.html:234
+msgid "Apply"
+msgstr "Εφαρμογή"
+
+#: ../vnc.html:241 ../vnc.html:271
+#, fuzzy
+msgid "Connect"
+msgstr "Συνδέεται..."
```
CendioOssman: `fuzzy` means that it found something similar, but not a perfect match. These needs to be adjusted and the `fuzzy` comment removed.

po/noVNC.pot
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: noVNC 0.6.1\n"
 "Report-Msgid-Bugs-To: novnc@googlegroups.com\n"
-"POT-Creation-Date: 2016-11-15 19:32+0100\n"
+"POT-Creation-Date: 2016-11-24 08:53+0200\n"
```
CendioOssman: As you can see you didn't need to update the `.pot` file for this. Please remove this from the commit to avoid polluting the history.

kosmasgiannis:
Hi,

Thanks for your suggestions.
I fixed the problems and it should be ok now.

CendioOssman:
Great. Looks good to me. 👍

PR #725, opened 2016-11-26
Allow control bar to be moved to the right side

CendioOssman:
In case the remote session has something on the left side that makes the control bar's placement there inconvenient.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -598,6 +598,8 @@ var UI;
 
             e.preventDefault();
             e.stopPropagation();
+            UI.keepControlbar();
+            UI.activateControlbar();
         },
 
         // Move the handle but don't allow any position outside the bounds
@@ -654,6 +656,8 @@ var UI;
                 UI.toggleControlbar();
                 e.preventDefault();
                 e.stopPropagation();
+                UI.keepControlbar();
+                UI.activateControlbar();
             }
             UI.controlbarGrabbed = false;
         },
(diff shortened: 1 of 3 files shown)


app/ui.js
```diff
@@ -576,11 +576,35 @@ var UI;
...
+            UI.controlbarDrag = true;
+        },
+
         dragControlbarHandle: function (e) {
             if (!UI.controlbarGrabbed) return;
 
             var ptr = Util.getPointerEvent(e);
 
+            var anchor = document.getElementById('noVNC_control_bar_anchor');
+            if (ptr.clientX < (window.innerWidth * 0.1)) {
```
samhed: I'm thinking that we should probably try to use a physical width here in the future. However, not a big deal

CendioOssman:
> The only drawback is that I feel that most noVNC users won't know that this feature exists. Is there any way we can display a graphical hint that in some way shows that the control bar can be moved to the other side? I'm absolutely not an expert, but some sort of shadow or arrow on the other side perhaps?

No idea right now. We'd want something that's clear and obvious, but at the same time not annoy users who already know they can do this.

app/ui.js
CendioOssman: Maybe. The advantage of a percentage is there is no risk of overlap. Seems to work well in practice right now, but we can always tweak it in the future.

samhed:
> No idea right now. We'd want something that's clear and obvious, but at the same time not annoy users who already know they can do this.

Yeah.. I don't see it as a reason not to merge

PR #727, opened 2016-11-30
Fix travis tests

CendioOssman:
Tweak the list of environment, and make the new language tests more tolerant of shitty browsers.


.travis.yml
```diff
@@ -8,12 +8,10 @@ node_js:
...
-  - TEST_BROWSER_NAME=chrome TEST_BROWSER_OS='Windows 7,Linux'
-  - TEST_BROWSER_NAME=firefox TEST_BROWSER_OS='Windows 7,Linux' TEST_BROWSER_VERSION='30,26'
-  - TEST_BROWSER_NAME='internet explorer' TEST_BROWSER_OS='Windows 7' TEST_BROWSER_VERSION=10
-  - TEST_BROWSER_NAME='internet explorer' TEST_BROWSER_OS='Windows 8.1' TEST_BROWSER_VERSION=11
-  - TEST_BROWSER_NAME=safari TEST_BROWSER_OS='OS X 10.8' TEST_BROWSER_VERSION=6
-  - TEST_BROWSER_NAME=safari TEST_BROWSER_OS='OS X 10.9' TEST_BROWSER_VERSION=7
+  - TEST_BROWSER_NAME=chrome TEST_BROWSER_OS='Windows 10,Linux,OS X 10.11'
+  - TEST_BROWSER_NAME=firefox TEST_BROWSER_OS='Windows 10,Linux,OS X 10.11'
+  - TEST_BROWSER_NAME='internet explorer' TEST_BROWSER_OS='Windows 10'
+  - TEST_BROWSER_NAME=safari TEST_BROWSER_OS='OS X 10.11'
```
samhed: I wan't to give @DirectXMan12 a chance to comment on this. But I'm fine with only testing the latest versions of browsers seeing as the update rate has always been very high barring IE. When it comes to operating system versions I've never experienced that it makes much difference when it comes to browsing.

tests/test.util.js
```diff
@@ -63,7 +63,19 @@ describe('Utils', function() {
...
+                this.skip();
+            }
+
             Object.defineProperty(window, "navigator", {value: {}});
+            if (window.navigator.languages !== undefined) {
+                // Object.defineProperty() doesn't work properly in old
+                // versions of Chrome
+                this.skip();
+            }
+
```
samhed: Were you able to reproduce the issue locally with Chrome version <= 48? As it seems like saucelab's linux machines are limited to Chrome 48 which isn't very old (~Jan 2016). After a quick look I can't find any chrome bug w.r.t this

Anyway, since you are actually checking the value of languages, we should be fine either way.

CendioOssman: I used Chromium 48 to test this issue. No notes about the issue in MDN either.

.travis.yml
CendioOssman: It's more of a sauce labs things. They only seem to be updating the latest versions properly.

samhed: Okay, that settles things then.

PR #732, opened 2016-12-10
Visually hint that you can drag the control bar to the other side

samhed:
It's not obvious, so I thought I would add a hint for it.

### DIFF
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -566,6 +566,29 @@ input[type=button]:active, select:active {
 }
 
 /* ----------------------------------------
+ * Silhouettes
+ * ----------------------------------------
+ */
+
+/* Outline that shows the form of a target element */
+.noVNC_silhouette {
+  /* Two colors to make it easier to see no matter the background */
+  box-shadow: 0 4px 5px white, 0 -4px 5px skyblue;
+  transition: visibility 0s linear 0.5s, opacity 0.5s ease-in-out;
+  position: fixed;
+  visibility: hidden;
+  opacity: 0;
+}
+.noVNC_silhouette.noVNC_active {
+  visibility: visible;
+  opacity: 1;
+  transition-delay: 0s;
+}
+.noVNC_silhouette.noVNC_flipped {
+  transform: rotate(180deg);
+}
+
+/* ----------------------------------------
  * Status Dialog
  * ----------------------------------------
  */
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -289,19 +289,33 @@ input[type=button]:active, select:active {
 
 #noVNC_control_bar_handle {
   position: absolute;
-  left: -15px;
+  left: 100%;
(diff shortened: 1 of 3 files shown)


samhed:
screenshot from 2016-12-10 21-37-53

How it looks while moving the control bar handle, while the control bar is open.

app/styles/base.css
```diff
@@ -289,20 +289,24 @@ input[type=button]:active, select:active {
...
+  border-radius: 0px 5px 5px 0px;
   background-color: rgb(83, 99, 122);
   background-image: url("../images/handle_bg.svg");
   background-repeat: no-repeat;
   background-position: right;
   box-shadow: 3px 3px 0px rgba(0, 0, 0, 0.5);
 }
+#noVNC_control_bar_handle:hover {
+  box-shadow: 0px 0px 7px 3px rgba(255, 255, 255, 0.6);
+}
```
CendioOssman: This removes the original shadow and just looks weird. Why this hover effect?

CendioOssman: I assume this is also what causes the outline to be asymmetrical? Any specific reason for that? I think it looks odd.

app/ui.js
```diff
@@ -608,15 +608,64 @@ var UI;
...
 
+        displayControlBarHint: function (show) {
+            var handleHint = WebUtil.getSilhouette("noVNC_control_bar_handle");
+            var controlBarHint = WebUtil.getSilhouette("noVNC_control_bar");
+
+            if (!show) {
+                if (handleHint === null || controlBarHint === null) return;
+
+                handleHint.style.display = "none";
+                controlBarHint.style.display = "none";
```
CendioOssman: We did a lot of work to get stuff like this out of the code an in to CSS. It would be a shame if we regress on that.

Any reason why you cannot put the look of the hint in CSS and just let JavaScript control when it is visible?

CendioOssman:
An animation to fade the hint in and out would be nice as well. :)

app/styles/base.css
samhed: The reason I added it in the first place was to emphasize that it could be interacted with.. But now as I think more about it, I don't mind removing this.

Another reason to remove it is how poorly hovers work on mobile.

samhed: I removed it.

app/ui.js
samhed: Fixed, in most cases. Some of the style manipulation has to be done in JavaScript since the silhouettes needs the dimensions and positions of the target elements.

samhed:
Good points!

I fixed the shadow (with a pseudo element), added a fade-in and out animation and moved as much of the style manipulation as possible to css.

I'll probably make a new PR with another alternative, a more simple hint. I'm personally not 100% sold on this one due to the complexity of the solution and the questionable value.

samhed:
I'm going to close this one, I don't like this approach.

Issue #737, opened 2016-12-28
Requests highest compression needlessly


patrakov:
By default, noVNC presents pseudo-encoding 0xffffff09 in its list of preferred encodings. This encoding corresponds to the highest compression level (9). However, libvncserver also interprets this compression level in its Tight encoder, and has an interesting comment:

https://github.com/LibVNC/libvncserver/blob/9c7efb7633ba62cd80c93e83284663f805bb3031/libvncserver/tight.c#L331

    /* CL 9 (which maps internally to CL 3) is included mainly for backward
       compatibility with TightVNC Compression Levels 5-9.  It should be used
       only in extremely low-bandwidth cases in which it can be shown to have a
       benefit.  For low-color workloads, it provides typically only 10-20%
       better compression than CL 2 with JPEG and CL 1 without JPEG, and it
       uses, on average, twice as much CPU time. */

See also the top of the file.

Besides the bad consequences for the CPU usage, it also hints libvncserver to use JPEG compression more often and thus sometimes blurs text. Red text in terminals on Ubuntu MATE 16.04 (with Vino) is affected. The screenshot below is with medium JPEG quality, but the same kind of blur is also reproducible with high JPEG quality, so it's best to avoid triggering this bug.

![image](https://cloud.githubusercontent.com/assets/2943163/21530771/edcded82-cd64-11e6-9082-d97d543c311c.png)

I suggest defaulting to ['compress_custom', -254 ] which is the highest compression level that doesn't lead to this kind of artifact.

kanaka:
Interesting, I hadn't seen information about that before. Sounds like a reasonable change. @patrakov Can you submit a simple pull request so that you get credit for the change/suggestion (if the rest of the team approves).

patrakov:
Yes - tomorrow. I should not submit pull requests after 1AM :)

kanaka:
@patrakov no hurry. How's the weather in Yekaterinburg? :-)

patrakov:
The weather is fine.

Further testing revealed that the blurry bug specifically affects Vino 3.8.1, which is still shipped by Ubuntu as the last version without hard dependency on the rest of GNOME. There is no blur when testing against x11vnc. Still, the "comment in libvncserver" rationale applies.

samhed:
Fixed in #738

hongweipeng:
@patrakov hi brother, where can set jpeg compression level? I want to use the `JPEG_quality_lo` .

PR #738, opened 2016-12-29
Changed the default compression level to 2

patrakov:
The old default was to ask for the maximum compression level. This
is against the recommendations in libvncserver/tight.c due to excessive
CPU load. It also causes Vino 3.8.1 (still shipped with Ubuntu 16.04
LTS) to prefer the blurry JPEG compression too much - e.g. red text on
the default background in MATE terminal becomes almost unreadable.

The new default is the recommended compression level for low-color
workloads, according to libvncserver source. Also, it is the maximum
compression level that doesn't trigger the Vino bug with red text in
most cases.

Fixes issue #737.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -60,7 +60,8 @@
         ['JPEG_quality_med',      -26 ],
         //['JPEG_quality_hi',     -23 ],
         //['compress_lo',        -255 ],
-        ['compress_hi',          -247 ],
+        ['compress_hi',          -254 ],
+        //['compress_max',       -247 ],
 
         ['DesktopSize',          -223 ],
         ['last_rect',            -224 ],


CendioOssman:
Remove it and leave it up to the server?

patrakov:
Interesting proposal, I have to test it

patrakov:
Tested. Unfortunately specifying no compression level (i.e. leaving it up to the server) leads to the same blurry image with Vino 3.8.1, i.e. does not fix the bug.

samhed:
@DirectXMan12 and @ossman what's your take on this one?

CendioOssman:
The naming is wrong, but otherwise I say go ahead and merge it. It gets noVNC in line with most other VNC implementations.

samhed:
The name of the encoding?

CendioOssman:
It's not really "high" compression anymore.

DirectXMan12:
merging (we can always fix the naming later).

PR #743, opened 2017-01-07
Fix rfb._cleanup()

samhed:
Regression caused by e549ae074fcea9febde32c0fa260a64c15cc1b8e. Things work fine with 84cd0e719ead6a1d99827d00794ef821593564fc. The function display.get_context() was removed but it is still called in rfb.js in _cleanup().

This results in bugs where you for example are be unable to write in text input fields after disconnecting.

@ossman removing this wasn't intended I guess?

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -444,15 +444,13 @@
         },
 
         _cleanup: function () {
-            if (this._display && this._display.get_context()) {
-                if (!this._view_only) { this._keyboard.ungrab(); }
-                if (!this._view_only) { this._mouse.ungrab(); }
-                this._display.defaultCursor();
-                if (Util.get_logging() !== 'debug') {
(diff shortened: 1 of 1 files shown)


CendioOssman:
Removing that function was very much intended. In fact, that was one of the main points of the commit. I didn't notice the call in `_cleanup()` though.

Looking at that code though, it's not obvious what it is up to. Are you familiar with it?

CendioOssman:
I'd say just nuke the `if`. `_display` is set in the constructor, so I don't see how `_cleanup` could ever be reached with `_display` not set and initialised.

samhed:
Agreed, that check isn't useful anymore. I checked the blame logs and it is simply a remnant of when the function worked differently and was called under other circumstances.

PR #747, opened 2017-01-11
Add reconnect option

### LINKED ISSUE #746: Add option for automatic reconnect
Hi,

thanks a lot for your work on noVNC!

I'm using this excellent tool to share my display with external people (e.g. clients), and work together with them (so noVNC is used for screen-sharing). Usually I'm sending a URL with configuration parameters for noVNC to the client, so they can open it and it connects automatically. Most of the time, this is just used for displaying my content, instead of having the client interact with the programs running on my machine (the VNC server is used in read-only mode, so no interaction via VNC is possible).

It regularly happens that people who have sub-optimal Internet connectivity (e.g. IPv4 via Dual-Stack lite) experience disconnects. That's really annoying because most of the time they need to find the mouse for the computer running the browser (and noVNC) and click on "Connect", and that means a break, which interrupts our work (imagine a phone conference with me on one side and several people standing in front of a TV set watching my desktop on the other side).

Since I have no control over the other party's network connection, I can't really debug or find out what's going on. Locally, I'm unable to reproduce the issue, but as far as I can see my setup is solid and stable.

I'd like to request adding an option to automatically reconnect in case of a disconnect.

I know that several people in the past suggested this (e.g. in #539, #540, #670) and it was always denied/closed, but I'd really like you to reconsider.

If you're willing to merge a PR (e.g. with code similar to #670) I'm willing to work on this. Please reconsider and let me know.

Thanks!

fd0:
Add an option to automatically reconnect after a delay when the
connection is closed.

Closes #746

### DIFF
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -771,8 +771,12 @@ input[type=button]:active, select:active {
   flex-direction: column;
 }
 :root.noVNC_connecting #noVNC_transition,
-:root.noVNC_disconnecting #noVNC_transition {
+:root.noVNC_disconnecting #noVNC_transition,
+:root.noVNC_reconnecting #noVNC_transition {
   display: flex;
+}
+:root:not(.noVNC_reconnecting) #noVNC_cancel_reconnect_button {
+  display: none;
 }
 #noVNC_transition_text {
   font-size: 1.5em;
--- a/app/ui.js
+++ b/app/ui.js
@@ -98,6 +98,10 @@ var UI;
         lastKeyboardinput: null,
         defaultKeyboardinputLen: 100,
 
+        inhibit_reconnect: true,
+        reconnect_callback: null,
+        reconnect_password: null,
+
         // Setup rfb object, load settings from browser storage, then call
         // UI.init to setup the UI/menus
         load: function(callback) {
@@ -204,6 +208,8 @@ var UI;
             UI.initSetting('view_only', false);
             UI.initSetting('path', 'websockify');
             UI.initSetting('repeaterID', '');
+            UI.initSetting('reconnect', false);
+            UI.initSetting('reconnect_delay', 5000);
         },
 
         setupWindowEvents: function() {
@@ -341,6 +347,8 @@ var UI;
                 .addEventListener('click', UI.disconnect);
             document.getElementById("noVNC_connect_button")
                 .addEventListener('click', UI.connect);
+            document.getElementById("noVNC_cancel_reconnect_button")
+                .addEventListener('click', UI.cancelReconnect);
 
             document.getElementById("noVNC_password_button")
                 .addEventListener('click', UI.setPassword);
@@ -413,6 +421,7 @@ var UI;
             document.documentElement.classList.remove("noVNC_connecting");
             document.documentElement.classList.remove("noVNC_connected");
             document.documentElement.classList.remove("noVNC_disconnecting");
+            document.documentElement.classList.remove("noVNC_reconnecting");
 
             switch (state) {
                 case 'connecting':
@@ -421,6 +430,7 @@ var UI;
                     break;
                 case 'connected':
                     UI.connected = true;
+                    UI.inhibit_reconnect = false;
                     document.documentElement.classList.add("noVNC_connected");
                     if (rfb && rfb.get_encrypt()) {
                         msg = _("Connected (encrypted) to ") + UI.desktopName;
@@ -467,6 +477,8 @@ var UI;
             document.getElementById('noVNC_setting_port').disabled = UI.connected;
             document.getElementById('noVNC_setting_path').disabled = UI.connected;
             document.getElementById('noVNC_setting_repeaterID').disabled = UI.connected;
+            document.getElementById('noVNC_setting_reconnect').disabled = UI.connected;
+            document.getElementById('noVNC_setting_reconnect_delay').disabled = UI.connected;
 
             if (UI.connected) {
                 UI.updateViewClip();
@@ -844,6 +856,8 @@ var UI;
             UI.saveSetting('path');
             UI.saveSetting('repeaterID');
             UI.saveSetting('logging');
+            UI.saveSetting('reconnect');
+            UI.saveSetting('reconnect_delay');
 
             // Settings with immediate (non-connected related) effect
             WebUtil.init_logging(UI.getSetting('logging'));
@@ -890,6 +904,8 @@ var UI;
             UI.updateSetting('path');
             UI.updateSetting('repeaterID');
             UI.updateSetting('logging');
+            UI.updateSetting('reconnect');
+            UI.updateSetting('reconnect_delay');
 
             document.getElementById('noVNC_settings')
                 .classList.add("noVNC_open");
@@ -1037,12 +1053,15 @@ var UI;
             }
         },
 
-        connect: function() {
+        connect: function(event, password) {
             var host = document.getElementById('noVNC_setting_host').value;
             var port = document.getElementById('noVNC_setting_port').value;
             var path = document.getElementById('noVNC_setting_path').value;
 
-            var password = WebUtil.getConfigVar('password');
+            if (typeof password === 'undefined') {
+                password = WebUtil.getConfigVar('password');
+            }
+
             if (password === null) {
                 password = undefined;
             }
@@ -1073,16 +1092,49 @@ var UI;
             UI.closeAllPanels();
             UI.rfb.disconnect();
 
+            // Disable automatic reconnecting
+            UI.inhibit_reconnect = true;
+
             // Restore the callback used for initial resize
             UI.rfb.set_onFBUComplete(UI.initialResize);
 
             // Don't display the connection settings until we're actually disconnected
         },
 
+        reconnect: function() {
+            UI.reconnect_callback = null;
+
+            // if reconnect has been disabled in the meantime, do nothing.
+            if (UI.inhibit_reconnect) {
+                return;
+            }
+
+            UI.connect(null, UI.reconnect_password);
+        },
+
         disconnectFinished: function (rfb, reason) {
             if (typeof reason !== 'undefined') {
                 UI.showStatus(reason, 'error');
+            } else if (UI.getSetting('reconnect', false) === true && !UI.inhibit_reconnect) {
+                document.getElementById("noVNC_transition_text").textContent = _("Reconnecting...");
+                document.documentElement.classList.add("noVNC_reconnecting");
+
+                var delay = parseInt(UI.getSetting('reconnect_delay'));
+                UI.reconnect_callback = setTimeout(UI.reconnect, delay);
+                return;
             }
+
+            UI.openControlbar();
+            UI.openConnectPanel();
+        },
+
+        cancelReconnect: function() {
+            if (UI.reconnect_callback !== null) {
+                clearTimeout(UI.reconnect_callback);
+                UI.reconnect_callback = null;
+            }
+
+            document.documentElement.classList.remove("noVNC_reconnecting");
             UI.openControlbar();
             UI.openConnectPanel();
         },
@@ -1110,7 +1162,9 @@ var UI;
         },
 
         setPassword: function() {
-            UI.rfb.sendPassword(document.getElementById('noVNC_password_input').value);
+            var password = document.getElementById('noVNC_password_input').value;
+            UI.rfb.sendPassword(password);
+            UI.reconnect_password = password;
             document.getElementById('noVNC_password_dlg')
                 .classList.remove('noVNC_open');
             return false;
--- a/vnc.html
+++ b/vnc.html
@@ -231,6 +231,14 @@
                                 </ul></div>
                             </li>
                             <li><hr></li>
+                            <li>
+                                <label><input id="noVNC_setting_reconnect" type="checkbox" /> Automatic Reconnect</label>
+                            </li>
+                            <li>
+                                <label for="noVNC_setting_reconnect_delay">Reconnect Delay (ms):</label>
+                                <input id="noVNC_setting_reconnect_delay" type="input" value="" />
+                            </li>
+                            <li><hr></li>
                             <!-- Logging selection dropdown -->
                             <li>
                                 <label>Logging:
@@ -289,6 +297,9 @@
     <!-- Transition Screens -->
     <div id="noVNC_transition">
         <div id="noVNC_transition_text"></div>
+        <div>
+        <input type="button" id="noVNC_cancel_reconnect_button" value="Cancel" class="noVNC_submit" />
+        </div>
         <div class="noVNC_spinner"></div>
     </div>
 


fd0:
This is the first version, do I need to mention the options somewhere, e.g. in the docs?

samhed:
I'm not sure I see the use case here? Why would we want this?

> do I need to mention the options somewhere, e.g. in the docs?

No, we don't have much documentation for the example application, mostly for the main modules.

samhed:
I now see that you explained your use case in #747

app/ui.js
```diff
@@ -436,6 +443,10 @@ var UI;
                     break;
                 case 'disconnected':
                     UI.showStatus(_("Disconnected"));
+
+                    if(UI.reconnect) {
```
samhed: pls add a space after "if" before the parenthesis

app/ui.js
```diff
@@ -436,6 +443,10 @@ var UI;
...
+                        setTimeout(UI.connect, UI.reconnect_delay);
```
samhed: What do you think about displaying "reconnecting" instead of "connecting" in the status message here?

fd0: Good idea, thanks.

fd0:
Thanks for the review, I've addressed the style issues.

I've also pushed a new commit that changes the behavior slightly to address your concerns:
 * Automatic reconnect is only enabled after a successful connection has been established once. In case of a wrong password, this will not connect again until the correct password has been supplied.
 * Manual disconnect will disable auto reconnect.

Let me know what you think :smile:

app/ui.js
```diff
@@ -98,6 +98,10 @@ var UI;
         lastKeyboardinput: null,
         defaultKeyboardinputLen: 100,
 
+        reconnect: false,
+        reconnect_delay: 5000,
```
samhed: You should be using our settings API instead of global variables for "reconnect" and "reconnect_delay"

fd0: Thanks for the hint, will do.

fd0:
Done, anything else? :)

samhed:
I would prefer if you could squash the first and last commit together, it doesn't make much sense history-wise to ever having used global variables. Thank you for your quick responses!

fd0:
Sure, here we go.

app/ui.js
```diff
@@ -98,6 +98,8 @@ var UI;
         lastKeyboardinput: null,
         defaultKeyboardinputLen: 100,
 
+        reconnect_enabled: false,
```
CendioOssman: This is easily confused with the primary reconnect setting. Another name would be nice.

app/ui.js
```diff
@@ -435,6 +440,16 @@ var UI;
                     document.documentElement.classList.add("noVNC_disconnecting");
                     break;
                 case 'disconnected':
+                    var reconnect = WebUtil.getConfigVar('reconnect');
+
+                    if ((reconnect === 'true' || reconnect == '1') && UI.reconnect_enabled) {
+                        UI.showStatus(_("Reconnecting"));
+
+                        var delay = parseInt(WebUtil.getConfigVar('reconnect_delay', UI.reconnect_delay));
+                        setTimeout(UI.connect, delay);
```
CendioOssman: A good start, but two issues:

 a) How do you abort the reconnect, if that wasn't what you wanted?
 b) We should block access to the rest of the UI whilst we're waiting for this timer or the user could easily get things in some odd state.

CendioOssman:
A UI setting could be added later as well.

app/ui.js
fd0: That's what I thought, but I didn't come up with a good name. Any suggestions? `last_connect_successful`?

app/ui.js
fd0: Hm, I'm not sure how to solve this. Do you have any suggestions?

samhed: For a) a dialogue with the option to cancel would probably be nice, and for b) we could use the transition screen that is used while connecting and disconnecting.

app/ui.js
samhed: Yeah, I find it tough to find a good name here as well. Your suggestion works fine for me

CendioOssman: We could reverse it and call it "inhibit_reconnect" or something like that.

fd0: Ok, that sounds good to me.

app/ui.js
fd0: Done, thanks for the suggestions.

fd0:
I've addressed the review comments. Should I squash the commits?

app/styles/base.css
```diff
@@ -771,9 +771,16 @@ input[type=button]:active, select:active {
...
+:root.noVNC_disconnecting #noVNC_transition,
+:root.noVNC_reconnecting #noVNC_transition {
   display: flex;
 }
+#noVNC_cancel_reconnect_button {
+  display: none;
+}
+:root.noVNC_reconnecting #noVNC_cancel_reconnect_button {
+  display: block;
+}
```
CendioOssman: Minor tip, here you can use the `:not` selector to avoid hard coding it's normal `display' state. There are some examples in that css file.

app/ui.js
```diff
@@ -383,6 +384,11 @@ var UI;
             window.addEventListener('msfullscreenchange', UI.updateFullscreenButton);
         },
 
+        addCancelReconnectHandlers: function() {
+            document.getElementById("noVNC_cancel_reconnect_button")
+                .addEventListener('click', UI.cancelReconnect);
+        },
+
```
CendioOssman: Perhaps we can group this with the connect and disconnect buttons?

app/ui.js
```diff
@@ -1099,14 +1096,49 @@ var UI;
             // Don't display the connection settings until we're actually disconnected
         },
 
+        reconnect: function() {
+            delete UI.reconnect_callback;
```
CendioOssman: We've used the style of setting it to `null` when not used. It should also be initialised in the constructor.

app/ui.js
```diff
@@ -1099,14 +1096,49 @@ var UI;
...
+                return;
+            }
+
+            UI.connect();
+        },
+
         disconnectFinished: function (rfb, reason) {
+            var reconnect = UI.getSetting('reconnect');
+
+            if ((reconnect === 'true' || reconnect == '1') && !UI.inhibit_reconnect) {
```
CendioOssman: The rest of the code just checks booleans directly without doing string comparisons, so it's probably best to do the same thing here.

app/ui.js
```diff
@@ -1099,14 +1096,49 @@ var UI;
...
+                document.getElementById("noVNC_transition_text").innerHTML = _("Reconnecting...");
```
CendioOssman: We just had a fix to avoid using `innerHTML`, so please switch to `textContent`. :)

app/ui.js
```diff
@@ -1099,14 +1096,49 @@ var UI;
...
+                document.documentElement.classList.add("noVNC_reconnecting");
+                UI.updateVisualState();
```
CendioOssman: Why is this call necessary?

app/ui.js
```diff
@@ -1099,14 +1096,49 @@ var UI;
...
+
+                var delay = parseInt(UI.getSetting('reconnect_delay'));
+                UI.reconnect_callback = setTimeout(UI.reconnect, delay);
+                return;
+            }
+
             if (typeof reason !== 'undefined') {
                 UI.showStatus(reason, 'error');
             }
```
CendioOssman: We probably still want to show errors. And perhaps not trigger the reconnect in those cases.

app/ui.js
```diff
@@ -1099,14 +1096,49 @@ var UI;
...
             UI.openControlbar();
             UI.openConnectPanel();
         },
 
+        cancelReconnect: function() {
+            UI.inhibit_reconnect = true;
```
CendioOssman: I don't think this is necessary here. Doesn't it get reset when you connect?

CendioOssman:
> I've addressed the review comments. Should I squash the commits?

Yes, but we can wait until there are no more comments. :)

app/styles/base.css
fd0: Ok

app/ui.js
fd0: done.

app/ui.js
fd0: fixed.

app/ui.js
fd0: Uhm, that's not the case, there's the autoconnect function which is checked here:
```javascript
if (autoconnect === 'true' || autoconnect == '1') {
    autoconnect = true;
```

Shall I also correct that code?

fd0: And then there's the check for the bell function:
```javascript
if (WebUtil.getConfigVar('bell', 'on') === 'on')
```

fd0: I've the code I added for the reconnect function.

app/ui.js
fd0: I'll do that. Please note that the other two transition text modifications also use `innerHTML`:
```javascript
case 'connecting':
    document.getElementById("noVNC_transition_text").innerHTML = _("Connecting...");
    document.documentElement.classList.add("noVNC_connecting");
    break;
[...]
case 'disconnecting':
    UI.connected = false;
    document.getElementById("noVNC_transition_text").innerHTML = _("Disconnecting...");
    document.documentElement.classList.add("noVNC_disconnecting");
    break;
```
I was unable to find the fix, is it maybe a good idea to also apply the fix in these two places?

app/ui.js
fd0: It apparently isn't, I've removed it.

app/ui.js
fd0: Good idea, I've moved the reconnect logic below the error check.

app/ui.js
fd0: I think you're right.

fd0:
Ok, next round of comments done :)

app/ui.js
samhed: The fix was very recent, please update your ui.js

fd0: Oh, indeed.

fd0:
I've rebased my commits on the current master branch, including the `innerHTML` fix.

samhed:
You should also run updateSetting for the new settings in UI.openSettingsPanel

app/ui.js
CendioOssman: Well this is embarrassing, that's my code and it's fairly recent. :)

But I had a close look, and the magic starts when you start using `UI.getSetting()` rather than `WebUtil.getConfigVar()`. The former checks if the setting is attached to a check box, and if so tries to convert things to a boolean.

CendioOssman:
I think it looks good now. Fix Samuel's last points and it just needs some testing.

fd0:
Oh, sorry, must've missed updating/saving the settings, pushed another commit.

samhed:
reconnectsetting

The setting works fine now, might want to move the setting a bit in the UI though. Or at least add a divider

fd0:
Divider is in :)

CendioOssman:
Seems to work well. However you have to reauthenticate. Is that really what we want for a reconnect?

fd0:
Hm, what's the solution for that? Adding code which caches the password for a reconnect, if reconnect is enabled?

What's missing for this PR to be merged?

CendioOssman:
> Hm, what's the solution for that? Adding code which caches the password for a reconnect, if reconnect is enabled?
> 

Probably. To some extent that is already in place as we don't clear out the input element.

> What's missing for this PR to be merged?

Nothing really. But I wanted to check the auth issue with you first, as it limits the usefulness if it only works for people with disabled authentication.

fd0:
I think you are right, reconnect should also work seamlessly with connections that require authentication. I'll add code for that shortly.

fd0:
I've added code for that. Please let me know if there is anything that needs to be done for this PR :)

CendioOssman:
Hmm.. You need to clear the password better somewhere. noVNC currently wedges itself if you enter the wrong password, as it keeps retrying the bad password.

fd0:
Hu? I don't understand, the reconnect should only be enabled once a successful connection (with the correct password) has been established. Did the password change in between connections?

CendioOssman:
You store the password right away, even if it is wrong. So I get an error that it failed to connect. But when I click Connect again I don't get a new prompt, so I can't fix my typo.

fd0:
Ah, thanks for the explanation!

fd0:
Fixed, thanks for pointing it out!

CendioOssman:
It's still not working properly for some reason. Does it work for you? I'm testing with Firefox on Linux.

I'm also cautious about having to add code here and there to clear the password. Seems very error prone. We should have a mechanism that works the other way, i.e. require opt-in.

How about this, extend `UI.connect()` to take an optional password argument. You then only pass that along from the timer, not from other places.

fd0:
Hm, it works for me (Linux, Firefox). Your suggestion sounds good, I'll do that.

fd0:
Great, this feels much better. @CendioOssman Could you please try again?

app/ui.js
```diff
@@ -1053,18 +1053,17 @@ var UI;
             }
         },
 
-        connect: function() {
+        connect: function(_, password) {
```
CendioOssman: What's with the `_` argument?

app/ui.js
```diff
@@ -1053,18 +1053,17 @@ var UI;
...
             var host = document.getElementById('noVNC_setting_host').value;
             var port = document.getElementById('noVNC_setting_port').value;
             var path = document.getElementById('noVNC_setting_path').value;
 
-            var password = WebUtil.getConfigVar('password');
+            if (typeof password === 'undefined') {
```
CendioOssman: `password === undefined` works just as well.

app/ui.js
samhed: why the _ parameter?

samhed: This results in an error as well, if host or port is empty, since the name of the translation function is _:

 Uncaught TypeError: _ is not a function
    at HTMLDivElement.connect (ui.js:1078)

fd0: When `connect()` is called via the connect button, the first parameter is the DOM element that was clicked on. We don't need that, so it is ignored.

app/ui.js
fd0: Yes, I was trying to match the style of the file. Shall I change that?

app/ui.js
fd0: @CendioOssman also asked that, see above ;)

fd0: Oh, that's a good point. I'll change the name of the parameter. The first parameter is used for the `MouseEvent` in an `OnClick` handler.

fd0: @samhed also notices that `_` is not a good parameter name, so I changed it.

fd0:
Sure, pushed. Anything else I need to change? :)

fd0:
Could you restart the Travis CI test please? It seems that one test timed out...

app/ui.js
CendioOssman: Nah, not a biggie.

CendioOssman:
Looks good. Thanks for working with us on this. :)

fd0:
Great, it was a pleasure working with you both :)

CraigZych:
Hi, The password caching method currently used doesn't work if the original connection was done thru url parameters and the autoconnect method. 
If using autoconnect it just requests a password on reconnect. I guess the password would have to be copied to the password field from the query string.

CendioOssman:
Please open an issue for that as this merge is already completed.

Craigzyc:
I actually realized it only doesnt work in the vnc lite page. It works fine in the full page so not really an issue, just not supported in lite.
Thanks

PR #614, opened 2016-05-14
Support for ATEN iKVM servers

kelleyk:
Implements support for the 0x56 and 0x57 formats used by the AST2100-family (AST2400, etc.) BMCs, among others.

Closes #385.

Issue #683, opened 2016-10-10
Fix up the ModulesAPI wiki Page


DirectXMan12:
https://github.com/kanaka/noVNC/wiki/ModulesAPI has outdated and incorrect information.  It should be updated to match the latest changes.

samhed:
Should be updated now. Feel free to take a look and fix any errors I have made.

Issue #675, opened 2016-09-30
We should publish on NPM


DirectXMan12:
As a follow-up to #496, we should set up Travis to automatically publish to NPM on a successful tagged version.

julien-f:
Any news on this?

nraynaud:
Hello y'all, 
as a temporary measure, I have published a scoped version of the package on NPM.
It's here: https://www.npmjs.com/package/@nraynaud/novnc

here is an example:
```javascript
import RFB from '@nraynaud/novnc/lib/rfb'
...
    const rfb = new RFB({
      encrypt: isSecure,
      target: this.refs.canvas,
      wsProtocols: [ 'binary' ],
      onClipboard: onClipboardChange && ((_, text) => {
        onClipboardChange(text)
      }),
      onUpdateState: this._onUpdateState
    })
    rfb.connect(url.hostname, url.port, null, path)
```

aftabnaveed:
For beginner like me it is very hard to use configure and use noVNC with existing projects. I recently started working on ReactJS/webpack project and all my modules are based on NPM. 
It is kind of hard for me to follow the integration instructions. My question is, is there a way to easily integrate noVNC into my existing project without using the browser-es-module-loader ?

lanwen:
As a workaround direct link to repo can be used:

`yarn add https://github.com/novnc/noVNC.git#master`

martinpitt:
Seems this became slightly more urgent, now that bower calls itself deprecated:
```
npm WARN deprecated bower@1.7.9: ..psst! While Bower is maintained, we recommend Yarn and Webpack for *new* front-end projects! Yarn's advantage is security and reliability, and Webpack's is support for both CommonJS and AMD projects. Currently there's no migration path but we hope you'll help us figure out one.
```
In cockpit we are [moving away from bower to npm+webpack](https://github.com/cockpit-project/cockpit/pull/6814), but for the time being we will use `"noVNC": "https://github.com/novnc/noVNC/archive/v0.6.2.tar.gz"` in `packages.json`. But that makes it much harder to stay on the current noVNC version.

Thanks for considering!

PR #756, opened 2017-01-17
Use RequireJS module system

CendioOssman:
Use a standard module system rather than a home grown one, and
RequireJS fits better with browsers than CommonJS.

This removes almost all head ache when it comes to splitting up files, handling dependencies and avoiding polluting namespaces.

More cleanup can probably be done, but I wanted some early feedback.

This became more relevant with #614 which wants to add many new files.


CendioOssman:
This PR isn't as huge as it first seems. View it using `git show -w` and you'll see it's mostly re-indentation.

samhed:
You can add ?w=1 at the end of the diff url to hide white space changes from the diff.

samhed:
I like this! @DirectXMan12 what's your opinion? I have a feeling this have been discussed, but I can't remember.

CendioOssman:
We're starting to look at cleaning up the keyboard code in order to solve the bug we're seeing, and this PR would be very helpful there.

So if no one objects in the next week then it'll probably get merged. :)

CendioOssman:
Might affect #675 and #683.

DirectXMan12:
So, this appears to remove support for browserify (CommonJS, something commonly asked for) and ES6 modules (which is where everything will end up once the module loader spec stabilizes and gets implemented), so my initial reaction is 👎 .  

Replacing the homegrown solution in `app/ui.js` and `vnc_auto.html` would be nice, and I'm not (at least initially) against RequireJS, but I don't think this PR is quite the right way to go, at first glance.

CendioOssman:
As far as I've understood things, browsers will support both CommonJS and RequireJS in the ES6 system, so that shouldn't be a problem.

And CommonJS code can import RequireJS modules via requirejs' CommonJS module. We could provide examples if people have problems sorting it out.

CendioOssman:
And if I've understood things correctly, browserify requires you to first process the files, right? With RequireJS people can just drop them in and use them right away.

CendioOssman:
And just to clarify, CommonJS can not be used in a browser. At all. So if we want to use that API then we will have to have a build stage in noVNC to convert things.

I can live with that if that's what it takes to get a proper module system in place, but I don't think it's the best solution.

DirectXMan12:
> As far as I've understood things, browsers will support both CommonJS and RequireJS in the ES6 system, so that shouldn't be a problem.

Not entirely sure what you're talking about here, but AFAIK, browsers will just support the ES6 module system -- no CommonJS or RequireJS (apart from the normal RequireJS functionality), and importing CommonJS and RequireJS modules from ES6 code will be difficult and/or impossible.

> And if I've understood things correctly, browserify requires you to first process the files, right? With RequireJS people can just drop them in and use them right away.

Yes, that's correct -- you write CommonJS style, and then bundle.  In our case, we actually just write ES6 modules in comment blocks (more or less -- there are some restrictions) and then a special series of browserify passes (which can be run standalone) uncomment the comments, which can then be used as-is (if your project already uses ES6 modules) or converted to CommonJS style for use with browserify.

The tricky part here is that if we go with RequireJS natively, we pretty much lose the ability to work with browserify and ES6-import-based (via Babel or what have you) projects, which was a commonly requested feature.  However, if we but RequireJS behind a Babelify (babelify == browserify + babel) pass (similar to how we get our CommonJS), we lose the benefit of our current system, which is that it "just works" out of the box without any compilation phase.

We *could* use something like this polyfill (https://github.com/ModuleLoader/browser-es-module-loader) to polyfill the module loader API, which would mean while developing the "ES6 imports to currently usable code" transformation would be done in-browser (meaning things would "just work"), but when we publish we'd want to publish the "compiled" version (whether we compile to CommonJS, RequireJS, etc).  This would mean just dropping the "comment" part of the equation and writing straight ES6 modules, and we'd basically just be able to drop the polyfill and compiling in the ✨ future ✨ .

CendioOssman:
> > As far as I've understood things, browsers will support both CommonJS and RequireJS in the ES6 system, so that shouldn't be a problem.
> 
> Not entirely sure what you're talking about here, but AFAIK, browsers will just support the ES6 module system -- no CommonJS or RequireJS (apart from the normal RequireJS functionality), and importing CommonJS and RequireJS modules from ES6 code will be difficult and/or impossible.
> 

I was positive I saw something about the browsers discussing adding shims to make the transition less painful. But I can't find anything now, so that must have been a wishful dream on my part...

> The tricky part here is that if we go with RequireJS natively, we pretty much lose the ability to work with browserify and ES6-import-based (via Babel or what have you) projects, which was a commonly requested feature. However, if we but RequireJS behind a Babelify (babelify == browserify + babel) pass (similar to how we get our CommonJS), we lose the benefit of our current system, which is that it "just works" out of the box without any compilation phase.

Right. I don't think we can make everyone happy. And in that case I think we should care more about people running noVNC in the browser than those running it in Node.

That said, using noVNC from Node after this PR is as simple as this:

```JavaScript
#!/usr/bin/env node

"use strict";

var requirejs = require("requirejs");
var rfb = requirejs("core/rfb");

var conn = rfb.RFB();
```

(Well almost. You still need to define a browser like environment with `window` and `document` first).

I assume this would work fine in Browserify? And Babel?

DirectXMan12:
> Right. I don't think we can make everyone happy. And in that case I think we should care more about people running noVNC in the browser than those running it in Node.

I think you're under a bit of a misconception here -- the CommonJS support isn't mainly for people who run under Node, it's for people who manage their projects with browserify (which is somewhat common these days), which uses CommonJS to manage dependencies, but bundles everything together in the end.  I'm not sure if browserify handles the case where things are defined as AMD -- I'll have to check, but I'd really prefer not to move away from being able to generate ES6 modules.  I need to play around a bit with some other solutions.

CendioOssman:
Ah. That's a much more sensible use case.

So I took a stab at it. Unfortunately browserify parses the source and looks for `require()`, so it cannot handle shims such as `requirejs()` or `amdefine()`. However there is a transform available for browserify called "deamdify". It seems to work well except for two issues:

 a) It cannot handle dynamically loaded modules (like our translations)
 b) It cannot handle multiple modules in the same file (like our mashed up inflator.js)

The second was easily fixed by simply not mashing everything together. The first will require some other loader.

I have a branch here that works fine for me:

https://github.com/CendioOssman/noVNC/tree/browserify

To generate the big `novnc.js` you run:

```
NODE_PATH=. ./node_modules/browserify/bin/cmd.js -t deamdify app/novnc -o novnc.js --debug
```

Have a look and see what you think.

kelleyk:
Is there anything that I can do to help this work land upstream?  It seems like it's become something of a dependency for #614.

DirectXMan12:
I need to take a look at @CendioOssman's code and take a look at my own idea.  I'll try to get it done in the next couple of days.

kelleyk:
Okay, great!

DirectXMan12:
@CendioOssman the ES6 alternative I was talking about: https://github.com/novnc/noVNC/tree/refactor/es6-module-loader.

By default, it runs using the es6-browser-module-loader polyfill (which translates to SystemJS on the fly in-browser), but `utils/use_require.js` can now be used to easily output AMD (RequireJS), CommonJS (Browserify), SystemJS (a thing related to the ES6 module loader spec), and UMD by using Babel (for instance, the tests are run through Babel first to convert to RequireJS format).

When the browsers move to have built-in support for the module loader (Soon <sup>:tm:</sup>), we won't have to do anything.  It also lets us reorganize things a bit to make the modules a bit more rollup friendly.

Since the browser module loader spec supports dynamic imports (i.e. our translations), that works fine with the SystemJS transformation (Babel parses the syntax fine in the other formats, but they don't implement the requisite handling of that syntax properly).

As for inflator.js, I've removed the pre-bundled copy in my branch (and converted pako to be natively es6 compatible).

CendioOssman:
I find rewriting things on the fly a bit scary, but as long as it works. :)
Debugging seems to be fine, showing the original source. However it takes a rather long time to load. And it's using a lot of CPU, so I assume it is Babel that is the cause of the delay. A bit of a shame, but I suppose we can handle it using a loading screen...

If you are happy with this then can you do a proper PR so we can comment on the details?

> Since the browser module loader spec supports dynamic imports (i.e. our translations), that works fine with the SystemJS transformation (Babel parses the syntax fine in the other formats, but they don't implement the requisite handling of that syntax properly).

It's just data so I think it's easier if we just convert it to JSON or XML instead and avoid the JavaScript loader.

CendioOssman:
> However it takes a rather long time to load. And it's using a lot of CPU, so I assume it is Babel that is the cause of the delay. A bit of a shame, but I suppose we can handle it using a loading screen...

Not easily apparently... It's keeping the browser so busy that it cannot update CSS animations properly.

DirectXMan12:
> If you are happy with this then can you do a proper PR so we can comment on the details?

Yep, will do.

> It's just data so I think it's easier if we just convert it to JSON or XML instead and avoid the JavaScript loader.

ack, I'll look in to doing that either way

> Not easily apparently... It's keeping the browser so busy that it cannot update CSS animations properly.

Strange, I didn't see that when I was testing.  I'll look into it.  At any rate, that's just for testing until the browsers natively support it.  We'll want to have "releases" of the app use SystemJS or RequireJS (I'd lean towards SystemJS since it's a bit more flexible, but whatever) until we have broad browser support for modules (initial support is on the way -- Firefox has it enabled for browser UI, and Chrome looks like the bits are mostly there, just not quite enabled yet).

CendioOssman:
Closing this in favour of #774.

PR #762, opened 2017-01-23
Save settings immediately

samhed:
Instead of saving settings when clicking on apply or toggling the settings panel, save them immediately after a change. This is more in line with how most modern interfaces work and saves us from having two different 'states' to keep track of in the UI code (cookie and HTML-elem value).

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1052,15 +1052,6 @@ var UI;
                 .classList.remove("noVNC_open");
         },
 
-        toggleConnectPanel: function() {
-            if (document.getElementById('noVNC_connect_dlg')
-                .classList.contains("noVNC_open")) {
-                UI.closeConnectPanel();
-            } else {
-                UI.openConnectPanel();
-            }
-        },
-
         connect: function(event, password) {
             var host = document.getElementById('noVNC_setting_host').value;
             var port = document.getElementById('noVNC_setting_port').value;
--- a/app/ui.js
+++ b/app/ui.js
@@ -146,7 +146,7 @@ var UI;
             UI.initFullscreen();
 
             // Setup event handlers
-            UI.setupWindowEvents();
+            UI.addResizeHandlers();
             UI.addControlbarHandlers();
             UI.addTouchSpecificHandlers();
             UI.addExtraKeysHandlers();
@@ -154,6 +154,8 @@ var UI;
             UI.addConnectionControlHandlers();
             UI.addClipboardHandlers();
             UI.addSettingsHandlers();
+            document.getElementById("noVNC_status")
+                .addEventListener('click', UI.hideStatus);
 
             UI.openControlbar();
 
@@ -235,13 +237,10 @@ var UI;
             UI.initSetting('reconnect_delay', 5000);
         },
 
-        setupWindowEvents: function() {
+        addResizeHandlers: function() {
             window.addEventListener('resize', UI.applyResizeMode);
             window.addEventListener('resize', UI.updateViewClip);
             window.addEventListener('resize', UI.updateViewDrag);
-
-            document.getElementById("noVNC_status")
-                .addEventListener('click', UI.hideStatus);
         },
 
         addControlbarHandlers: function() {
--- a/app/ui.js
+++ b/app/ui.js
@@ -191,7 +191,7 @@ var UI;
 
             // Settings with immediate effects
             UI.initSetting('logging', 'warn');
-            WebUtil.init_logging(UI.getSetting('logging'));
+            UI.updateLogging();
 
             // if port == 80 (or 443) then it won't be present and should be
             // set manually
@@ -868,7 +868,7 @@ var UI;
             UI.saveSetting('reconnect_delay');
 
             // Settings with immediate (non-connected related) effect
-            WebUtil.init_logging(UI.getSetting('logging'));
+            UI.updateLogging();
             UI.updateViewClip();
             UI.updateViewDrag();
             //Util.Debug("<< settingsApply");
@@ -1699,6 +1699,10 @@ var UI;
                 UI.rfb.get_keyboard().set_focused(true);
                 UI.rfb.get_mouse().set_focused(true);
             }
+        },
+
+        updateLogging: function() {
+            WebUtil.init_logging(UI.getSetting('logging'));
         },
 
         updateSessionSize: function(rfb, width, height) {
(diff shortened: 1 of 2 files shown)


samhed:
Another old annoying behavior that is removed with this change is the fact that toggling (by pressing the settings button again) the settings panel did save the settings while closing (by pressing some other toolbar button) didn't.

app/ui.js
```diff
@@ -367,14 +367,41 @@ var UI;
...
+            var settingElem = document.getElementById("noVNC_setting_" + name);
+            if (changeFunc === undefined) {
+                settingElem.addEventListener('change', function () {
+                    UI.saveSetting(name);
+                });
+            } else {
+                settingElem.addEventListener('change', changeFunc);
+            }
+        },
+
```
CendioOssman: If you wish to merge code paths a bit you could do this instead:

```JavaScript
if (changeFunc === undefined) {
    changeFunc = function () { UI.saveSettings(name); };
}
```

app/ui.js
```diff
@@ -367,14 +367,41 @@ var UI;
...
                 .addEventListener('click', UI.toggleSettingsPanel);
-            document.getElementById("noVNC_settings_apply")
-                .addEventListener('click', UI.settingsApply);
 
-            document.getElementById("noVNC_setting_resize")
-                .addEventListener('change', UI.enableDisableViewClip);
+            UI.addSettingChangeHandler('encrypt');
+            UI.addSettingChangeHandler('true_color');
+            if (Util.browserSupportsCursorURIs()) {
+                UI.addSettingChangeHandler('cursor');
```
CendioOssman: Why can't we add this unconditionally? Won't it be disabled where unsupported?

app/ui.js
```diff
@@ -139,9 +139,10 @@ var UI;
                 UI.toggleControlbarSide();
             }
 
-            // Setup and initialize event handlers
+            UI.initFullscreen();
```
CendioOssman: Why the move?

app/ui.js
```diff
@@ -265,6 +262,9 @@ var UI;
             // resize events aren't available for elements
             window.addEventListener('resize', UI.updateControlbarHandle);
 
+            document.getElementById("noVNC_status")
+                .addEventListener('click', UI.hideStatus);
```
CendioOssman: The status isn't really part of the control bar.

app/ui.js
```diff
@@ -502,7 +502,6 @@ var UI;
             }
 
             UI.enableDisableViewClip();
-            document.getElementById('noVNC_setting_resize').disabled = UI.connected;
```
CendioOssman: Isn't more code required? Don't we have to trigger an update of the canvas size and position?

samhed: I'm not sure what you mean? It works just fine

app/ui.js
samhed: It isn't really related to setting up event handlers, with the new separation it felt to have more in common with the init functions

app/ui.js
samhed: I like that, I'll change

app/ui.js
samhed: mostly just did the same thing as the old code here, but you are right, it's unnecessary.

app/ui.js
samhed: Yeah.. well it doesn't fit anywhere tbh. But I'll move it into start() I guess

kanaka:
I don't have time to do a code review, but I heartily approve of the general idea.

samhed:
Ok, thank you Joel.

I have rebased on the current master and updated the commits after your comments @CendioOssman, thanks for the feedback!

app/ui.js
CendioOssman: It doesn't here. Changing the setting doesn't do anything. I have to resize the window for things to take effect. Also, switching from "scaling" to "none" is just broken.

CendioOssman:
PS. Why don't you have your own fork for your own development? ;)

app/ui.js
```diff
@@ -375,14 +375,38 @@ var UI;
...
-            document.getElementById("noVNC_setting_resize")
-                .addEventListener('change', UI.enableDisableViewClip);
+            UI.addSettingChangeHandler('encrypt');
+            UI.addSettingChangeHandler('true_color');
+            UI.addSettingChangeHandler('cursor');
+            UI.addSettingChangeHandler('resize');
+            UI.addSettingChangeHandler('resize', UI.enableDisableViewClip);
+            UI.addSettingChangeHandler('clip');
+            UI.addSettingChangeHandler('clip', UI.updateViewClip);
+            UI.addSettingChangeHandler('clip', UI.updateViewDrag);
```
CendioOssman: Why are these needed at this point? These settings don't take effect until we are connected, and you can't change those settings once connected.

samhed: You are correct, the extra handlers for 'clip' are not needed. I was half way into a solution where we could change that setting while connected, but I never finished it.

app/ui.js
samhed: Yes, you are right.. I'm not sure why I thought that was acceptable tbh. I'll take a stab at fixing these things

samhed:
Changing the resize setting while connected should work as intended now. Thanks for testing!

> PS. Why don't you have your own fork for your own development? ;)

I do, I should probably use that more often :P

samhed:
Had forgotten to set the correct scaling for mouse coordinates, fixed now.

app/ui.js
```diff
@@ -375,14 +375,36 @@ var UI;
...
-            document.getElementById("noVNC_settings_apply")
-                .addEventListener('click', UI.settingsApply);
 
-            document.getElementById("noVNC_setting_resize")
-                .addEventListener('change', UI.enableDisableViewClip);
+            UI.addSettingChangeHandler('encrypt');
+            UI.addSettingChangeHandler('true_color');
+            UI.addSettingChangeHandler('cursor');
+            UI.addSettingChangeHandler('resize');
+            UI.addSettingChangeHandler('resize', UI.enableDisableViewClip);
```
CendioOssman: This should be removed as well, I assume?

samhed: We will actually need it I believe, since choosing any of the two scaling settings forces clipping to be disabled

CendioOssman: Alright. But it should still be in the patch that makes it possible to change when the session is active in that case.

samhed: Ah, I'll move it to the correct commit

samhed: Done

PR #764, opened 2017-01-24
Convert key code 219 to XK_bracketleft for passthrough of ctrl-[

matthewcmead:
Added a single conversion from key code 219 to XK_bracketleft so that ctrl-[
passes through noVNC into the server session.  This is useful for editing with
vi/vim as ctrl-[ acts as the ESC key but allows one to leave their fingers on
the home row.

### DIFF
--- a/core/input/util.js
+++ b/core/input/util.js
@@ -229,6 +229,7 @@ var KeyboardUtil = {};
             case 0xbc: return KeyTable.XK_comma;
             case 0xbd: return KeyTable.XK_minus;
             case 0xbe: return KeyTable.XK_period;
+            case 0xdb: return KeyTable.XK_bracketleft;
         }
 
         return nonCharacterKey({keyCode: keycode});


CendioOssman:
That doesn't look correct as you're adding a very layout sensitive key to the layout independent code.

What browser are you seeing this issue with?

matthewcmead:
I am seeing this issue with two browsers:

Firefox on Windows 7
Chrome on macOS 10.12 (sierra)

The commit corrects the issue in both cases.  Prior to the commit, xev (the X windows event viewer) sees no events when ctrl-[ is pressed in a session through VNC.  It only sees the control key event.  After the commit, xev sees the ctrl-[ event as expected and the key combination works appropriately in vim as well.

Is there a mechanism to do this from a layout sensitive portion of the code?

CendioOssman:
Maybe, I'm not sure. I'm actually trying to clean up that code right now since it is rather old and a lot of it deals with browsers too old to function with noVNC anyway.

What you should test is to use a different layout and make sure you still get the right character. E.g. Swedish which has `å` on that key.

CendioOssman:
Does this work on #766?

samhed:
@matthewcmead did you find time to try #766 ?

CendioOssman:
Works for me with #766 merged.

Issue #734, opened 2016-12-14
ie11 and edge no enter/backspace


flumm:
with current master, when pressing enter or backspace (possibly other keys too) nothing happens
with debug level on i can see on the console

onKeyPress down, keysym: 16777226(16777226)
onKeyPress up, keysym: 16777226(16777226)

on /tests/input.html i get

raw key event keydown (key: 13, char 0, which: 13)
RFB keypress up keysym: [object Object] key name: [object Object]

what can be wrong here?

flumm:
i found, that ie11 and edge send "\n" as char in the keyboardevent
which gets parsed as unicode and the keysym 0x100000a gets sent, but the enter gets not recognized

samhed:
I can somewhat reproduce your problem. When running xev in a session and pressing backspace or enter I get the same wierd 0x100000a keysyms. However, both keys work as expected when actually using them for typing etc.

CendioOssman:
A lot of applications will expect the proper keysym though.

flumm:
> I can somewhat reproduce your problem. When running xev in a session and pressing backspace or enter I get the same wierd 0x100000a keysyms. However, both keys work as expected when actually using them for typing etc.

what kind of vnc server do you use? i use the qemu builtin one, which does not work (qemu 2.7)

samhed:
I used TigerVNC

samhed:
This is a regression from commit 115eedf69c465548a44eab1b2a6492546b2ed6e1

zoon01:
With latest release the problem still exist.
No Enter or Backspace keys working.

Windows 10 with IE11 connecting a freebsd 11 VM with virtualbox and LibVNCServer-0.9.9
The browser chrome does work correctly.

CendioOssman:
> This is a regression from commit 115eedf

I reverted it and the problem still remains (albeit slightly different), so it's not that.

IE/Edge is setting `char` for its `keydown` events which is confusing our code. Will need to have a deeper look.

flumm:
are you sure?

in my local branch i reverted that commit, and it works fine since...

CendioOssman:
Perhaps not. I think I only got a revert on one of the files. Will have to look at it next week.

zoon01:
Where are we standing now? Is the problem already fixed in master?

samhed:
@zoon01 not fixed in master, but #766 fixes this at least for me.

CendioOssman:
#766 is merged so this is now fixed.

PR #766, opened 2017-01-31
Simplify keyboard code

CendioOssman:
A major cleanup of the keyboard code in order to bring it in to CURRENT YEAR. Some parts is to remove legacy compatibility that is no longer required, and some is sharing the experiences we've had from the TigerVNC development.

Testing so far shows that it can solve a bunch of current issues on several platforms. It merges the QEMU code path and should solve that bug as well, although I haven't tested it.

More testing will follow, but have a look and see what you think.

(Note that this adds two new globals, which would be nice to hide using #756)


samhed:
Fixes #734

oglueck:
Yey this fixes "qemu -k" and now noVNC behaves the same as other VNC clients.

CendioOssman:
Great. Thanks for testing. :)

samhed:
I guess our goal should be to fully test this and eventually get it merged as soon as #774 is in? Seeing as this rewrite fixes a lot of open issues

CendioOssman:
Indeed.

mjrussell:
If its helpful at all, I tried to do that on my fork - https://github.com/Simspace/noVNC/commit/56dced21126449c5d257aa03ff56a111551245a5

It seemed to work quite well, the only issue I had was I couldnt get safari to trigger commands on a Windows box that included holding down both the CTRL and another key (like CTRL + S for save). Using an on screen keyboard I could tell that I could send both CTRL and S but it appeared to not respect them both being pressed at the same time.

CendioOssman:
Is that a regression though? Did it work with the existing keyboard handling?

mjrussell:
Im not actually sure, I can try it and report back. I originally upgraded because the es6 module version didnt work with IE11

core/input/fixedkeys.js
```diff
@@ -0,0 +1,111 @@
...
+ * Licensed under MPL 2.0 or any later version (see LICENSE.txt)
+ */
+
+/* [module]
+ * import KeyTable from "./keysym";
+ */
+
+"use strict";
+
+var fixedkeys = {
```
DirectXMan12: I have a *huge* table of these in my PR from a while back that I did the research on.  We may just want to include that, since it's a bit more expansive, IIRC.

core/input/util.js
```diff
@@ -1,679 +1,170 @@
...
-
-    function isNumPadMultiKey(evt) {
-        return (numPadCodes.indexOf(evt.code) !== -1);
-    }
-
-    function getNumPadKeySym(evt) {
-        if (numLockOnKeyCodes.indexOf(evt.keyCode) !== -1) {
-            return numLockOnKeySyms[evt.code];
+        // We start with layout independent keys
+        var code = getKeycode(evt);
```
DirectXMan12: I'm not convinced this is the correct behavior -- keysym is supposed to be the *logical* keypress.  For example, if I remap `A` to `C`, we should send code `A` but the keysym for `C`.  This new behavior sends the keysym for `A` as well, which seems to be wrong, unless I've misinterpreted something.

We should just be relying on `key` for the keysym, if we can.

core/rfb.js
```diff
@@ -329,16 +330,33 @@
 
         // Send a key press. If 'down' is not specified then send a down key
         // followed by an up key.
-        sendKey: function (keysym, down) {
+        sendKey: function (keysym, code, down) {
```
DirectXMan12: this is breaking public API, so it'll need a release note (once we hit 1.0, which I'm hoping to soon, we can't break API this easily).

core/input/devices.js
```diff
@@ -154,7 +154,7 @@
         ['target',     'wo', 'dom'],  // DOM element that captures keyboard input
         ['focused',    'rw', 'bool'], // Capture and send key events
 
-        ['onKeyPress', 'rw', 'func'] // Handler for key press/release
+        ['onKeyEvent', 'rw', 'func'] // Handler for key press/release
```
DirectXMan12: need to remember to note this in the release notes too...

core/input/devices.js
```diff
@@ -50,59 +45,220 @@
...
+                    if (active[i] === keysym) {
+                        isCharModifier = true;
+                    }
+                }
+                if (!isCharModifier) {
+                    var escape = this._modifierState.activeCharModifier();
+                }
+            }
+
+            var last;
```
DirectXMan12: please comment this part as well, explain the list (it's not as clear as when things were broken up separately)

core/input/devices.js
```diff
@@ -136,10 +136,6 @@
 
             //Util.Debug(">> Keyboard.ungrab");
         },
-
-        sync: function (e) {
-            this._handler.syncModifiers(e);
-        }
```
DirectXMan12: it might not be a bad idea to keep a minimal form of this, where we release all modifiers if there are no modifiers at all pressed -- for embedded noVNC, (where noVNC could lose focus while a key was down), it might be useful, since accidentally pressing `Ctrl-SOMETHING` is probably not the best.

core/input/vkeys.js
```diff
@@ -0,0 +1,115 @@
+/*
+ * noVNC: HTML5 VNC client
+ * Copyright (C) 2017 Pierre Ossman for Cendio AB
+ * Licensed under MPL 2.0 or any later version (see LICENSE.txt)
+ */
+
+"use strict";
+
+var vkeys = {
```
CendioOssman: Sure. I'll add that.

core/input/fixedkeys.js
CendioOssman: If it's huge then it's probably not the same thing. This is **not** a mapping between DOM key codes and keysyms as such a thing would in most cases be incorrect. This is merely the subset of keys that we assume are mapped the same on every layout.

I can add a comment to this file as well.

core/rfb.js
CendioOssman: Indeed. If we were post-1.0 then it could have simply be handled by putting the new argument last and having some handling for it not being included.

core/input/devices.js
CendioOssman: We already release keys when we lose focus, so I'm not sure what scenario that would be needed?

core/input/devices.js
CendioOssman: This is internal between the keyboard handler and the RFB object, so it should be fine.

core/input/util.js
CendioOssman: We should, but unfortunately support for `key` isn't where we need it to be yet. As I mentioned in the other comment, this step is only for keys that are the same on every layout. So it will not do anything for e.g. `A` as that might be mapped to something else.

core/input/devices.js
CendioOssman: It gets nuked in a later commit though, so perhaps not necessary?

CendioOssman:
Rebased as modules now.

DirectXMan12:
> This is internal between the keyboard handler and the RFB object, so it should be fine.

NB: some people use our keyboard handler separately, apparently -- anything that's a public property should be considered public API.

> We should, but unfortunately support for key isn't where we need it to be yet. As I mentioned in the other comment, this step is only for keys that are the same on every layout. So it will not do anything for e.g. A as that might be mapped to something else.

Ah, but you can't assume *any* key is fixed -- that's my point.  I remap `CapsLock` as `Escape`, for instance, since I use vim and don't like pinky-related RSI.  Many emacs users do a similar thing, except to control.  When I'm working on a normal (non-VM) VNC session, when I press they key that I've defined as `Escape`, I want it to send `Escape`, not some other key that happens to have a different meaning on a "normal" keyboard layout.

I suggest that we use `key` when possible (I have a whole big map of `key` values to keysyms ~~somewhere around here~~ https://github.com/DirectXMan12/noVNC/blob/feature/use-ui-events/core/input/keysym-map.js), and fall back to "guessing" like this only when necessary.

Useful page, if you haven't seen it: https://dvcs.w3.org/hg/d4e/raw-file/tip/key-event-test.html.  It's the w3c key event tester page, which displays info about the legacy and new DOM input events.

CendioOssman:
> NB: some people use our keyboard handler separately, apparently -- anything that's a public property should be considered public API.

I don't think that's a use case we should support. I think we have enough on our plate making a good RFB implementation without also committing to providing various utilities for other things.

CendioOssman:
> Ah, but you can't assume any key is fixed -- that's my point. I remap CapsLock as Escape, for instance, since I use vim and don't like pinky-related RSI. Many emacs users do a similar thing, except to control. When I'm working on a normal (non-VM) VNC session, when I press they key that I've defined as Escape, I want it to send Escape, not some other key that happens to have a different meaning on a "normal" keyboard layout.

I'm fine not supporting those right now since that's going to be very messy and is such a very small group of users. Did that even work in the existing code?

> I suggest that we use `key` when possible...

I suggest "when practical", which is probably not for a few more years. At the very least we need to get Safari on board. And the other browsers still have lots of bugs in their reported names as the standard got changed so many times.

Trying to do it now would just result in basically two separate implementations, which I'm not looking forward to testing and maintaining.

CendioOssman:
> I'm fine not supporting those right now since that's going to be very messy and is such a very small group of users. Did that even work in the existing code?

I tested, and it did work in the previous version. The reason was however not our code, but that the browsers send symbols in `KeyboardEvent.keyCode` and not physical keys like in `KeyboardEvent.code`.

I'm still not convinced this is a worthwhile use case, but I can have a look for the sake of giving preference to browsers that implement `KeyboardEvent.key` properly.

DirectXMan12:
> I tested, and it did work in the previous version. The reason was however not our code, but that the browsers send symbols in KeyboardEvent.keyCode and not physical keys like in KeyboardEvent.code

> I'm still not convinced this is a worthwhile use case, but I can have a look for the sake of giving preference to browsers that implement KeyboardEvent.key properly.

Thanks!  Otherwise, this is good work.

core/input/domkeytable.js
```diff
@@ -0,0 +1,310 @@
...
+import KeyTable from "./keysym.js"
+
+/*
+ * Mapping between HTML key values and VNC/X11 keysyms for "special"
+ * keys that cannot be handled via their Unicode codepoint.
+ *
+ * See https://www.w3.org/TR/uievents-key/ for possible values.
+ */
+
+var DOMKeyTable = {};
```
DirectXMan12: This, IIRC, was the big ol' table that I mentioned a while back.

core/input/util.js
```diff
@@ -64,91 +65,74 @@ export function getKeycode(evt){
...
-        // Special key? (FIXME: Should have been caught earlier)
-        if (evt.key.length !== 1) {
-            return null;
-        }
-
-        codepoint = evt.key.charCodeAt();
-    } else if ('charCode' in evt) {
-        codepoint = evt.charCode;
+    // Special key? (FIXME: Should have been caught earlier)
+    if (key.length !== 1) {
```
DirectXMan12: probably want to log this, at least.

tests/test.helper.js
```diff
@@ -64,39 +64,60 @@ describe('Helpers', function() {
...
+                expect(KeyboardUtil.getKeysym({key: 'Semicolon'})).to.be.null;
+                expect(KeyboardUtil.getKeysym({key: 'BracketRight'})).to.be.null;
             });
-            it('should not recognize character keys', function() {
-                expect(KeyboardUtil.getKeysym({code: 'KeyA'})).to.be.null;
-                expect(KeyboardUtil.getKeysym({code: 'Digit1'})).to.be.null;
-                expect(KeyboardUtil.getKeysym({code: 'Period'})).to.be.null;
-                expect(KeyboardUtil.getKeysym({code: 'Numpad1'})).to.be.null;
+            it('should handle remappings', function() {
+                expect(KeyboardUtil.getKeysym({code: 'ControlLeft', key: 'Tab'})).to.be.equal(0xFF09);
```
DirectXMan12: 😄 Clearly, this needs a comment `// for crazy people like @directxman12` :-P

DirectXMan12:
(just needs a rebase to fix error-handler)

CendioOssman:
Tested on pretty much every combination of OS and browser. Should solve every issue that isn't a browser bug/limitation.

samhed:
Very nice! :+1: 

Could you update the Modules API wiki page? At least "sync" and "notify" should be removed. I recently spent time getting it in sync, it would be a shame to not keep it up.

PR #774, opened 2017-02-11
[Refactor] ES6 Modules

DirectXMan12:
This PR changes over to using native ES6 modules.  For normal development, the Browser ES Modules polyfill allows loading everything directly in the browser without any precompilation.  When we publish releases, we'll probably want to publish a SystemJS version (which this PR provides tools to do).


DirectXMan12:
@CendioOssman I've added a commit which optimizes the loader polyfill to run in Web Workers, so it shouldn't block the browser.  It also caches the compilation results.

DirectXMan12:
> I assume the module loader polyfill is this one: https://github.com/ModuleLoader/browser-es-module-loader I note on there "This project is only suitable for demonstrations / experimentation and is not designed for any production workflows at all."

So, the base module loader polyfill (the non-browser-specific variant) is prod-ready, and is used as the basis for SystemJS.  I suspect they say that because the browser-specific one does on-the-fly translation using Babel.  If/when this is approved, I'm going to add some Travis hooks to make our lives easier on release, and one of those will produce a version which uses the "precompiled" SystemJS variant.

> I think vnc_auto.html got missed. I suppose that raises a more general question of what we want to do with vnc_auto.html vs vnc.html.

That's completely intentional -- vnc_auto.html is currently broken, and we need to take some time to fix it and bring its UI up to date.

app/ui.js
```diff
@@ -11,14 +11,12 @@
...
- * import RFB from "../core/rfb";
- * import Display from "../core/display";
- * import WebUtil from "./webutil";
- */
+import Util from "../core/util.js";
+import KeyTable from "../core/input/keysym.js";
+import keysyms from "../core/input/keysymdef.js";
+import RFB from "../core/rfb.js";
+import Display from "../core/display.js";
+import WebUtil from "./webutil.js";
```
CendioOssman: Is the `.js` suffix required? It doesn't seem to be the recommended norm.

app/ui.js
```diff
@@ -1724,7 +1712,7 @@ var UI;
  */
     };
 
-    /* [module] UI.load(); */
+    UI.load();
```
CendioOssman: We should hold off on starting the UI until the translations are loaded.

vnc.html
```diff
@@ -323,9 +323,9 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br />VNC</h1>
     </audio>
 
     <!-- begin scripts -->
-    <script src="core/util.js"></script>
-    <script src="app/webutil.js"></script>
-    <script src="app/ui.js"></script>
+    <script src="node_modules/browser-es-module-loader/dist/babel-browser-build.js"></script>
+    <script src="node_modules/browser-es-module-loader/dist/browser-es-module-loader.js"></script>
+    <script type="module" src="app/ui.js"></script>
```
CendioOssman: Shouldn't this be part of the same commit that converts our modules to ES6 format? Otherwise we'll have a few commits that aren't usable.

CendioOssman: I think we should bundle the loader so that people don't have to run `npm install` in order to use noVNC.

CendioOssman: Can't these be put at the top now since only the module loader will be loaded before the page?

app/ui.js
```diff
@@ -66,9 +66,14 @@ var UI;
...
     var LINGUAS = ["de", "el", "nl", "sv"];
     Util.Localisation.setup(LINGUAS);
-    if (Util.Localisation.language !== "en") {
-        WebUtil.load_scripts(
-            {'app': ["locale/" + Util.Localisation.language + ".js"]});
+    if (Util.Localisation.language !== "en" && Util.Localisation.dictionary !== undefined) {
+        // NB: this only works with "native" ES6 modules and SystemJS modules -- the Babel translaters
+        // for CommonJS and AMD don't support this.  However, it is fairly easy to perform this logic
+        // yourself if you wish to package noVNC in CommonJS or AMD formats: simply load the appropriate
+        // language module and set Util.Localisation.dictionary to the default export of it.
```
CendioOssman: As discussed, we should probably fix this by loading it using `XMLHTTPRequest`.

core/display.js
```diff
@@ -84,790 +85,785 @@ export default function Display(defaults) {
...
-        this._cursor_uri = Util.browserSupportsCursorURIs();
+        this._cursor_uri = cursorURIsSupported();
     }
 
-    Util.Debug("<< Display.constructor");
+    Log.Debug("<< Display.constructor");
 };
 
-(function () {
-    "use strict";
```
CendioOssman: Is this added by the module loader, or do we need to have it globally still?

core/display.js
```diff
@@ -84,790 +85,785 @@ export default function Display(defaults) {
...
             this._cursor_uri === undefined) {
-        this._cursor_uri = Util.browserSupportsCursorURIs();
+        this._cursor_uri = cursorURIsSupported();
     }
 
-    Util.Debug("<< Display.constructor");
+    Log.Debug("<< Display.constructor");
 };
 
-(function () {
```
CendioOssman: I think it would be better to have the removal of function wrappers as a separate commit. Those parts are huge re-indents, so it's easy for other things to get lost in the noise.

core/websock.js
```diff
@@ -12,8 +12,7 @@
  * read binary data off of the receive queue.
  */
 
-import Util from "./util.js";
-
+import * as Log from './util/logging.js';
```
CendioOssman: We should remember to update websockify as well so we don't get out of sync again.

.travis.yml
```diff
@@ -7,7 +7,6 @@ node_js:
 - '6.1'
 env:
   matrix:
-  - TEST_BROWSER_NAME=PhantomJS
```
CendioOssman: This is a bit of an issue as it makes it difficult to run locally. What's preventing us from using PhantomJS 1.5?

karma.conf.js
```diff
@@ -103,39 +54,30 @@ module.exports = function(config) {
 
     // frameworks to use
     // available frameworks: https://npmjs.org/browse/keyword/karma-adapter
-    frameworks: ['mocha', 'sinon', 'chai', 'sinon-chai'],
-
+    frameworks: ['requirejs', 'mocha', 'chai'],
```
CendioOssman: Shouldn't this be done together with an earlier commit? Otherwise I assume there are a few commits that fail the tests?

app/styles/base.css
```diff
@@ -654,6 +654,79 @@ select:active {
 }
 
 /* ----------------------------------------
+ * Loading Animation
+ * ----------------------------------------
+ */
+
+#noVNC_loading_anim {
```
CendioOssman: Why not use our existing transition? The way I did it was add a new "state" class `noVNC_loading`, set that by default on `<html>` and have `ui.js` clear it on load.

core/display.js
CendioOssman: It seems you also missed `ui.js`. :)

vnc.html
DirectXMan12: Yep, that's done later in the PR (see the most recent commits).

core/display.js
DirectXMan12: Modules are strict-by-default, according to the spec (http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code <-- see the second bullet)

.travis.yml
DirectXMan12: all sorts of pieces of javascript that are present in all modern browsers are missing from PhantomJS < 2.5 (which is in beta).  You can still run locally with Karma -- I've updated the config so that you can run against any of your installed browsers.

karma.conf.js
DirectXMan12: yeah, a few commits fail the tests, but I wanted to make it easier to review each piece

app/styles/base.css
DirectXMan12: Ah, did not realize we had an existing `noVNC_loading` class.  I'll tweak things a bit.

core/display.js
DirectXMan12: > It seems you also missed `ui.js`. :)

Ack, will fix.

vnc.html
DirectXMan12: yes

app/ui.js
DirectXMan12: Yeah.  I think they realized in the module loader spec that they needed a good way to determine what file format a module was in when loading it, so they switched to recommending the `.js` extension there.

app/ui.js
DirectXMan12: Yep, I'll add a commit -- I realized I forgot to do that at like, 3 a.m. :-P.  It's a shame that `fetch` isn't supported in IE (only Edge), otherwise I'd just use that (it's *so* much nicer).

DirectXMan12:
I'll try and get to the fixed over the next couple of days.  Thanks for the feedback.

app/ui.js
CendioOssman: Got a reference? The ECMAScript spec is just horrible to read, but it doesn't seem to mention module name resolution at all. I found this:

https://github.com/whatwg/loader

But it just has a big **TODO** and examples so far.

core/display.js
CendioOssman: But until we have native support something needs to add that back. Did you check that the Babel conversion does so?

.travis.yml
CendioOssman: It requires you to manually switch over to the browser though, which is a much more cumbersome workflow.

Is it possible to depend on a beta version of PhantomJS? Is it on npm?

karma.conf.js
CendioOssman: Fair enough. The diff for the tests didn't look too big though so it might have been worth it.

app/styles/base.css
CendioOssman: `noVNC_loading` would be new, but we have a `noVNC_transition`. This diff is what I did:

```diff
diff --git a/app/styles/base.css b/app/styles/base.css
index d166bb3..087089e 100644
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -769,6 +769,7 @@ select:active {
   justify-content: center;
   flex-direction: column;
 }
+:root.noVNC_loading #noVNC_transition,
 :root.noVNC_connecting #noVNC_transition,
 :root.noVNC_disconnecting #noVNC_transition,
 :root.noVNC_reconnecting #noVNC_transition {
diff --git a/app/ui.js b/app/ui.js
index 98bb5f6..db4f167 100644
--- a/app/ui.js
+++ b/app/ui.js
@@ -170,6 +170,8 @@ var UI;
 
             document.getElementById('noVNC_setting_host').focus();
 
+            document.documentElement.classList.remove("noVNC_loading");
+
             var autoconnect = WebUtil.getConfigVar('autoconnect', false);
             if (autoconnect === 'true' || autoconnect == '1') {
                 autoconnect = true;
diff --git a/vnc.html b/vnc.html
index c6f0d25..de4d5bb 100644
--- a/vnc.html
+++ b/vnc.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html class="noVNC_loading">
 <head>
 
     <!--
```

Very minimal. :)

app/ui.js
DirectXMan12: it's scattered throughout their issues, IIRC.  I can try to dig up the actual spot at some point.  At any rate, it works according to the spec, AFAICT, and it's required for the polyfill.

.travis.yml
DirectXMan12: `karma run --browser chrome` will run the entire test suite without doing much of anything beyond that (for instance).  Unfortunately, the maintainers of the PhatomJS NPM modules we were depending on haven't updated to Phantom 2.5 yet, so we'll have to wait.

core/display.js
DirectXMan12: Yup, just checked (`./utils/use_require.js --as systemjs --with-app` and then check the generated files).

app/styles/base.css
DirectXMan12: done

/me is sad his fancy bouncing "no" won't make it in :-P

CendioOssman:
Something that just occurred to me, we probably need to move the error handler to it's on, non-module, .js file in order for it to grab errors during loading.

Btw, have you tested on all the major browsers, or do you need help with that?

app/ui.js
```diff
@@ -61,19 +61,6 @@ window.addEventListener('error', function(event) {
...
-if (l10n.language !== "en" && l10n.dictionary !== undefined) {
-    // NB: this only works with "native" ES6 modules and SystemJS modules -- the Babel translaters
-    // for CommonJS and AMD don't support this.  However, it is fairly easy to perform this logic
-    // yourself if you wish to package noVNC in CommonJS or AMD formats: simply load the appropriate
-    // language module and set l10n.dictionary to the default export of it.
-    import(`./locale/${l10n.language}.js`).then((language_mod) => {
-        l10n.dictionary = language_mod.default;
-    });
-}
-
```
CendioOssman: Perhaps move this patch earlier in the series to avoid ever having that warning?

app/ui.js
```diff
@@ -1714,6 +1701,21 @@ const UI = {
...
+var LINGUAS = ["de", "el", "nl", "sv"];
+l10n.setup(LINGUAS);
+if (l10n.language !== "en" && l10n.dictionary === undefined) {
+    WebUtil.fetchJSON('app/locale/' + l10n.language + '.js', function (translations) {
+        l10n.dictionary = translations;
+
+        // wait for translations to load before loading the UI
+        UI.load();
+    }, function (err) {
+        throw err;
```
CendioOssman: I guess this is keeping with the previous behaviour, but do we really want to fail fatally on failing to load translations?

app/ui.js
```diff
@@ -1714,6 +1701,21 @@ const UI = {
  */
 };
 
-export default UI;
+// Set up translations
+var LINGUAS = ["de", "el", "nl", "sv"];
+l10n.setup(LINGUAS);
+if (l10n.language !== "en" && l10n.dictionary === undefined) {
```
CendioOssman: Did you consider my approach of hiding the loading logic in the localisation code?

app/ui.js
```diff
@@ -1714,6 +1701,21 @@ const UI = {
...
+        UI.load();
+    }, function (err) {
+        throw err; 
+    });
+} else {
+    UI.load();
+}
+
 
-UI.load();
```
CendioOssman: I think you mentioned this in an earlier commit, so you may want to tweak the messages. :)

app/ui.js
DirectXMan12: I did consider it, but I think it's useful to let the noVNC users manually load in translations using their own mechanisms.

CendioOssman: Wrong thread? :)

DirectXMan12:
@samhed @CendioOssman I've done a couple of the fixes and started rebasing by hand -- git's merge algorithms are having a hard time dealing with the diffs, so I have to copy things into the PR by hand.  Can you please avoid merging anything this weekend so I can finish that up and we can get this merged?

samhed:
@DirectXMan12 will do!

DirectXMan12:
Alright, this should be rebased on top of the most recent changes.  All of the rebasing had to be done by hand, so let me know if I've missed something.  I also updated the "disable labels for disabled elements" functionality to work using only CSS.

DirectXMan12:
P.S. the last commit isn't strictly necessary, but I wasn't particularly fond of all of the javascript involved in styling work.

CendioOssman:
> P.S. the last commit isn't strictly necessary, but I wasn't particularly fond of all of the javascript involved in styling work.

-1 from me, I'm afraid. I'm usually against doing styling in Javascript as well, but in this case I consider it the lesser evil. Having to have a very strict structure and then visually reordering things is not particularly elegant, and prone to break whenever someone changes anything.

.travis.yml
CendioOssman: Doesn't work for me. First off, karma is hidden under `node_modules`, so it's not in $PATH. Secondly it still craps out:

```
$ PATH=$PATH:node_modules/karma/bin karma run  --browser firefox
27 02 2017 12:29:44.820:ERROR [runner]: There is no server listening on port 9876
```

Running `npm test` in another terminal still doesn't help:

```
$ PATH=$PATH:node_modules/karma/bin karma run  --browser firefox
No captured browser, open http://localhost:9876/
```

Can we improve on this? We don't want to discourage people from running the test suite.

core/display.js
CendioOssman: I take it it's a NAK on the split?

vnc.html
CendioOssman: Ping

app/ui.js
CendioOssman: Ping

app/ui.js
CendioOssman: Ping

CendioOssman:
I've poked a few threads that seemed to have been missed. :)

Did you also see my comment about error handling and browser testing?

DirectXMan12:
> -1 from me, I'm afraid. I'm usually against doing styling in Javascript as well, but in this case I consider it the lesser evil. Having to have a very strict structure and then visually reordering things is not particularly elegant, and prone to break whenever someone changes anything.

Fair enough.  I'll pull in the JS-based code and remove that commit.

> Did you also see my comment about error handling and browser testing?

Whoops, missed that last night when I was getting this up to speed.  I'll pull that in.  I tested last night on IE11, Edge, Firefox, and Chrome.  IE11 needs a polyfill for promises or `use_require --as amd --with-app` before run, it appears, but we probably want to do that with the published version anyway.  Otherwise, it seems to work well across browsers, after I fixed a couple of issues introduced by some recent commits.

DirectXMan12:
I'll try to get to the last bits tonight

samhed:
> P.S. the last commit isn't strictly necessary, but I wasn't particularly fond of all of the javascript involved in styling work.

Yeah, I don't like it either, but as Pierre said, the lesser evil. Hopefully we'll get the :has selector soon! Doesn't look too bright however:

http://caniuse.com/#feat=css-has

vnc.html
DirectXMan12: Yeah, a few of the commits in this PR are not-quite-usable standalone.  Splitting up the commits, like this, IMO, makes the individual commits easier to review and examine historically, since they're broken up by what they do (this commit simply uncomments the module syntax, this commit updates Pako to work with the module syntax, this commit adds in the polyfill, etc).

DirectXMan12:
Should be all set now, I think I've addressed all of your pings, and the error handling.  I think I've mostly covered the major browsers, but it wouldn't hurt to have someone else take things for a spin.

app/error-handler.js
```diff
@@ -0,0 +1,62 @@
...
+                .classList.add("noVNC_open");
+        } catch (exc) {
+            document.write("noVNC encountered an error.");
+        }
+        // Don't return true since this would prevent the error
+        // from being printed to the browser console.
+        return false;
+    }
+    window.addEventListener('error', function (evt) { handleError(event, event.error); });
+    window.addEventListener('unhandledrejection', function (evt) { handleError(event.reason, event.reason); });
```
CendioOssman: This looks wrong. The argument is `evt`, but you are referencing `event`.

app/styles/base.css
```diff
@@ -247,6 +247,16 @@ select:active {
...
 }
 
+#noVNC_fallback_errormsg .noVNC_message {
+  margin-left: auto;
+  margin-right: auto;
+  display: inline-block;
+}
+
+#noVNC_fallback_errormsg p {
+  text-align: left;
```
CendioOssman: Can't you put this on `.noVNC_message` and avoid the nested `<p>`?

app/styles/base.css
```diff
@@ -247,6 +247,16 @@ select:active {
   font-weight: normal;
 }
 
+#noVNC_fallback_errormsg .noVNC_message {
+  margin-left: auto;
+  margin-right: auto;
+  display: inline-block;
```
CendioOssman: Why `inline-block`? The next `<div>` should be below it.

CendioOssman:
Oh right, I guess `vnc_auto.html` needs to be fixed before we merge this to `master`?

CendioOssman:
And there's more errors with the error handler... Please test it a bit more as it is utterly broken right now. Two commits are always nice for these things, one for the move and one for the new features. Right now it's difficult to see what has been changed.

CendioOssman:
I'm a bit uneasy about the fact that only Chrome shows errors as I can't even find a time frame for support in other browsers. :/

I did this change, which improved things a bit at least:

```diff
diff --git a/app/ui.js b/app/ui.js
index 23c1733..a6991a2 100644
--- a/app/ui.js
+++ b/app/ui.js
@@ -1700,6 +1700,7 @@ const UI = {
  */
 };
 
+window.setTimeout(function () {
 // Set up translations
 var LINGUAS = ["de", "el", "nl", "sv"];
 l10n.setup(LINGUAS);
@@ -1715,5 +1716,6 @@ if (l10n.language !== "en" && l10n.dictionary === undefined) {
 } else {
     UI.load();
 }
+});
 
 export default UI;
```

The next issue is that no one seems to apply the source maps to the stack trace, making it very annoying to use, possibly to the point of useless. I'll play around with some tool to apply the source map in JavaScript.

CendioOssman:
This change might also be worthwhile:

```diff
diff --git a/vnc.html b/vnc.html
index c044673..dd9836f 100644
--- a/vnc.html
+++ b/vnc.html
@@ -62,10 +62,6 @@
     <!-- this is included as a normal file in order to catch script-loading errors as well -->
     <script type="text/javascript" src="app/error-handler.js"></script>
 
-    <!-- begin scripts -->
-    <script src="vendor/browser-es-module-loader/dist/browser-es-module-loader.js"></script>
-    <script type="module" src="app/ui.js"></script>
-    <!-- end scripts -->
 </head>
 
 <body>
@@ -75,6 +71,12 @@
         <div id="noVNC_fallback_errormsg"></div>
     </div>
 
+    <!-- The error handler is now fully in place, so load the rest of the scripts -->
+    <!-- begin scripts -->
+    <script src="vendor/browser-es-module-loader/dist/browser-es-module-loader.js"></script>
+    <script type="module" src="app/ui.js"></script>
+    <!-- end scripts -->
+
     <!-- noVNC Control Bar -->
     <div id="noVNC_control_bar_anchor" class="noVNC_vcenter">
```

app/error-handler.js
DirectXMan12: yeah, I missed it, since it works because window.event is a thing :-/

CendioOssman:
IE11 doesn't work for me. First and foremost it fails to load browser-es-module-loader.js. Secondly it fails to print the error as it doesn't have `ChildNode.remove()`. Use `Node.removeChild()` instead.

app/styles/base.css
DirectXMan12: Easier to get centering without an explicit size set.  It occurs to me that the `margin-left` and `margin-right` are unnecessary here.

app/styles/base.css
DirectXMan12: yeah, I could.  I'll switch around the JS a bit.

DirectXMan12:
> Please test it a bit more as it is utterly broken right now

I tested before I pushed on FF and Chrome, and just retested, and it seems to work fine (noVNC loads, and when I have `app/ui.js` throw an error while loading, it shows up in Chrome).

> The next issue is that no one seems to apply the source maps to the stack trace

It seems to have worked fine for me in Chrome, but Firefox seems to have some hiccups with it.  I'll see if I can figure out why that is.

> IE11 doesn't work for me. First and foremost it fails to load browser-es-module-loader.js.

There's a polyfill needed for IE11, I discovered.  MS seem to have mostly abandoned adding things to IE11, and never got around to promises.  I can add in the polyfill, or for the moment you can just test the output version from `./use_require.js --as amd --with-app` (as I noted above).

DirectXMan12:
Alright, polyfill in place, and I've worked around the lack of `unhandledrejection` events by having the loader fire an error event on the window object, and I've tested that loader errors fire in Firefox, Chrome, and IE11.

I'm not able to reproduce whatever error was alleviated with the setTimeout handler.  Can you provide some more info if it reproduces with the latest set of updates?

DirectXMan12:
Right, so we should now have a functional vnc_auto.html as well as vnc_playback.html.  You'll need to prefix the recording file variables with `export` now, but that's easily done with sed/by hand.  In the future, we probably want to look into generating recordings that are just JSON blobs instead of JS files.

DirectXMan12:
> Chrome only does it in the console AFAICT, not in the Error event. But I found this:
> https://github.com/novocaine/sourcemapped-stacktrace

I'll look into that, but we may want to consider it as a follow-up PR.  Thankfully, with what we're doing, source maps aren't *too* necessary, because the translation basically just shifts the code down by 20 lines, without really changing anything else

CendioOssman:
> I'm not able to reproduce whatever error was alleviated with the setTimeout handler. Can you provide some more info if it reproduces with the latest set of updates?

That was to move the startup out of the Promise execution so that errors could be caught. But it sounds like that isn't necessary anymore. I'll test.

CendioOssman:
> Right, so we should now have a functional vnc_auto.html as well as vnc_playback.html. You'll need to prefix the recording file variables with export now, but that's easily done with sed/by hand.

Can't we just let them continue to be loaded in the global context?

> In the future, we probably want to look into generating recordings that are just JSON blobs instead of JS files.

Agreed

> > Chrome only does it in the console AFAICT, not in the Error event. But I found this:
> >     https://github.com/novocaine/sourcemapped-stacktrace
> 
> I'll look into that, but we may want to consider it as a follow-up PR.

Sure. We might want to add an issue so that it doesn't get forgotten.

CendioOssman:
IE11 still broken. Needs this fix:

```diff
diff --git a/vendor/browser-es-module-loader/src/browser-es-module-loader.js b/vendor/browser-es-module-loader/src/browser-es-module-loader.js
index 2a4b6a5..2a9325f 100644
--- a/vendor/browser-es-module-loader/src/browser-es-module-loader.js
+++ b/vendor/browser-es-module-loader/src/browser-es-module-loader.js
@@ -231,7 +252,7 @@ BrowserESModuleLoader.prototype[RegisterLoader.instantiate] = function(key, proc
   }).then(function (data) {
     // evaluate without require, exports and module variables
     // we leave module in for now to allow module.require access
-    if (!data.key.endsWith('#nocache')) {
+    if (!data.key.slice(-8) === '#nocache') {
         localStorage.setItem(key+'!raw', data.source);
         localStorage.setItem(data.key+'!transpiled', data.code);
     }
```

samhed:
Interesting side note: 

With regards to iOS and Safari on Mac, today I get an error in the console:

> QuotaExceededError: Dom exception 22: An attempt was made to add something to storage that exceeded the quota

And it seems like local storage is disabled in private browsing in Safari - not related to this PR. However, at the moment I can't reproduce my error of infinite loading.. I'll keep looking into it.

CendioOssman:
Testing out the built versions:

**CommonJS**

Broken:

 a) Errors out on load with ```TypeError: selectbox is null```
 b) Installs all .js files even though everything is bundled up in `app.js`
 c) Fails to produce a source map for `app.js`

**AMD**

Source maps are generated, but not referenced in files. Source maps also reference the wrong source file.

Otherwise seems to work.

**SystemJS**

Same as AMD, except it doesn't work in IE11 because it needs Promises.

CendioOssman:
> a) Errors out on load with TypeError: selectbox is null

I think this might be because the script is executing before the page is loaded, hence it cannot find any DOM elements. The other formats load asynchronously, but not with browserify.

> b) Installs all .js files even though everything is bundled up in app.js

I guess this is a necessity of how the build works. We could consider adding a separate removal step afterwards.

> c) Fails to produce a source map for app.js

Urgh. browserify has really sucky support for source maps. We'll need at least two more tools it seems to deal with this. We'll probably have to fix this afterwards...

DirectXMan12:
> I guess this is a necessity of how the build works. We could consider adding a separate removal step afterwards.

We really only want to remove things when `--with-app` is specified.  Otherwise, we want to generate all files for CommonJS so they can be imported separately.

DirectXMan12:
So, this latest set of changes should fix the following issues:

- IE11 being broken due to not having `endsWith` (grr...)
- CommonJS failing to load (loading the UI is now done with `UI.prime`, which basically either goes ahead with loading if the ready state is appropriate, or adds a `DOMContentLoaded` event listener)
- AMD and SystemJS referencing source maps (note: source maps, when in separate form, appear to need to have absolute paths.  I generate them with paths *not* including the `/build` part, since this is more useful when packaging.  If you need to run from the /build directory, use inline source maps).  You can now also specify that you want inline source maps.
- SystemJS not working on IE11 (we now include the promise polyfill in the generated SystemJS `vnc.html`)
- bonus source map internal name hijinks (babel actually mutates the passed-in options object :-/)

Outstanding issues (which can probably be pushed to follow up PRs):

- browserify source maps
- respecting source maps in our error dialogs
- Safari local storage (I very little way to test this -- at least with IE11 I can use my gaming desktop, for Safari I have to actually get the SauceLabs connector working properly)

Please let me know if I've missed anything.

app/ui.js
```diff
@@ -11,1745 +11,1755 @@
...
+    rememberedClipSetting: null,
+    lastKeyboardinput: null,
+    defaultKeyboardinputLen: 100,
+
+    inhibit_reconnect: true,
+    reconnect_callback: null,
+    reconnect_password: null,
+
+    prime: function(callback) {
+        if (document.readyState === "interactive" || document.readyState === "complete") {
```
CendioOssman: `document.readyState !== 'loading'`?

CendioOssman:
> We really only want to remove things when --with-app is specified. Otherwise, we want to generate all files for CommonJS so they can be imported separately.

Agreed.

> note: source maps, when in separate form, appear to need to have absolute paths.

They worked just fine for me with relative paths. What problems were you seeing? Having absolute paths prevents people from putting noVNC where they'd like so that's not a very good solution.

Did you see my suggested patch? That also solved the issue of the opts object being modified.

DirectXMan12:
> They worked just fine for me with relative paths. What problems were you seeing? Having absolute paths prevents people from putting noVNC where they'd like so that's not a very good solution.

They do, I was just really tired last night and not quite thinking about the word "relative" properly :-/.

> Did you see my suggested patch? That also solved the issue of the opts object being modified.

Yeah, I fixed it similarly.

app/ui.js
DirectXMan12: Yes, but no -- it's more future-proof to say the states where you know you'll work, as opposed to the state where you won't.  Suppose there's a new readyState added in the future (there probably won't be, because it would break a massive number of things that are written like this) that doesn't have the DOM fully loaded?

CendioOssman: I was thinking that we wanted to avoid the loading state, but you're right; we're really interested in when the DOM is available.

CendioOssman:
Looks good. Tested "un-built", CommonJS, AMD, SystemJS on:

 - Firefox on Linux and Android
 - Chrome on Linux and Android
 - IE11 on Windows 2008 R2
 - Safari on iOS

Did you see my comment about loading recordings? Other than that I'd say we're good to go.

mjrussell:
This is awesome work, Im really looking forward to support for this. Is there a recommended way for someone else to try importing this into a project? Im assuming I cant just point NPM at the repo with this branch because you go through some manual steps for packaging. Should I just fork it, run the build, remove the build artifacts from `.gitignore` and try it that way?

DirectXMan12:
> Should I just fork it, run the build, remove the build artifacts from .gitignore and try it that way?

Yeah.  If you're already running babel in your project, no need to run any builds, just make sure you have one of the es6 module plugins loaded (I think the presets include one).  Otherwise, in the mean time, yeah, you can do that.  Once this merges, I'll set up Travis builds to publish the built versions under GitHub releases, and we'll put the CommonJS version (with an included raw ES6 modules directory for working with things like rollup) on NPM.

DirectXMan12:
> Did you see my comment about loading recordings?

The reason I didn't do that is because it would necessitate some basic form of WebUtil.load_scripts going back in, and I wasn't if we wanted to do that.  I figured it was easy enough to `sed -i s/var (\w)+ =/export var $1 =/` (or do it by hand), and we'll probably want to have a broader discussion about how we import the recordings separately (e.g. switch over to making them JSON objects).

DirectXMan12:
@samhed any comments, or does it seem good to go on your end?

mjrussell:
In my project the deps are setup to be ES5 so I ended up running the pre-publish to create the lib and then pushing that to a forked branch. It worked pretty well, except for a couple of broken paths:

https://github.com/Simspace/noVNC/blob/es6-dist/lib/inflator.js#L8-L10

These are referenced too high (they pull in ES6 code and it needs to import ES5), it should be `./vendor`

mjrussell:
I also had to change the `.npmignore` like [this](https://github.com/Simspace/noVNC/commit/560fa0a8eb4849268e929a63ecf080730c237e92). It was excluding a util file from pako with the path `util/common.js`.

I think you could also have used the `files` option in `package.json`.

core/util/localization.js
```diff
@@ -0,0 +1,170 @@
...
+                }
+                if (elem.hasAttribute("value") &&
+                    elem.tagName === "INPUT" &&
+                    isAnyOf(elem.getAttribute("type"), ["reset", "button"])) {
+                    translateAttribute(elem, "value");
+                }
+            }
+
+            for (var i = 0;i < elem.childNodes.length;i++) {
+                let node = elem.childNodes[i];
```
mjrussell: Do you mean to use es6 here? If you want to export compatible es5 you need to run this through babel-node or something.

Otherwise this should probably be changed to var

CendioOssman:
Indeed. Do a `git diff -w` and you'll see that the commits reintroduce some removed lines.

CendioOssman:
Some unnecessary returns are also restored in devices.js

CendioOssman:
@DirectXMan12, you still with us? :)

DirectXMan12:
lol, yeah, I went on vacation, and managed to not actually do any work (egads!).  Sometimes I think that GitHub needs an out-of-office indicator :-P.

I'll try and fix this up tonight before my flight back, or tomorrow.

DirectXMan12:
ok, that should fix the rebase issues (I've removed the returns properly and fixed the ui.js changes)

DirectXMan12:
merging, this should let us move forward with other things as well :-).

I'm going to file a couple of follow-up issues as well.

CendioOssman:
Crap. Something went wrong with this along the way. Firefox cannot load `master`:

```
17:46:34.255 SyntaxError: in strict mode code, functions may be declared only at top level or immediately within another function1 browser-es-module-loader.js:1172:11
```

DirectXMan12:
> Crap. Something went wrong with this along the way. Firefox cannot load `master`

Errm... works for me (just tested -- Firefox 55 nightly)?  Which version are you testing on?

CendioOssman:
Firefox ESR (45.8)

PR #776, opened 2017-02-17
Disable labels for settings

samhed:
When a setting is disabled, it makes sense to disable the labels as well.


app/styles/base.css
```diff
@@ -582,6 +582,9 @@ select:active {
 #noVNC_setting_path {
   width: 100px;
 }
+.noVNC_disabled_setting {
```
CendioOssman: Perhaps just `.noVNC_disabled` to allow use for more things and reduce clutter?

app/ui.js
```diff
@@ -235,6 +236,27 @@ var UI;
...
+        setupSettingLabels: function() {
+            var labels = document.getElementsByTagName('LABEL');
+            for (var i = 0; i < labels.length; i++) {
+                var label = labels[i];
+                var htmlFor = label.htmlFor;
+                if (htmlFor != '') {
+                    var elem = document.getElementById(htmlFor);
+                    if (elem) elem.label = label;
+                } else {
+                    // If 'for' isn't set, let's use the first non-text child
```
CendioOssman: Shouldn't you look for input like elements rather than just anything? I.e. check if `Element.form !== undefined`.

app/ui.js
```diff
@@ -860,6 +890,18 @@ var UI;
             return val;
         },
 
+        disableSetting: function(name) {
+            var ctrl = document.getElementById('noVNC_setting_' + name);
```
CendioOssman: A comment about why we have this helper would probably be nice. I.e. it compensates for lack of selectors in CSS.

vnc.html
```diff
@@ -222,7 +222,7 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br />VNC</h1>
                                     </li>
                                     <li>
                                         <label for="noVNC_setting_port">Port:</label>
-                                        <input id="noVNC_setting_port" />
+                                        <input id="noVNC_setting_port" type="number" />
```
CendioOssman: There are more number fields. E.g. reconnect delay.

app/styles/base.css
samhed: Sure, I can't find any good place to put it in the css though. Perhaps just at the top below .noVNC_only_touch.noVNC_hidden?

app/ui.js
samhed: Good point!

app/ui.js
samhed: I will add that

app/styles/base.css
CendioOssman: Do we still have that? I thought that was removed...

Wherever is fine. Close to the top probably makes sense as it overrides the default text colour and nothing else.

samhed: Yeah, we don't seem to use noVNC_only_touch anywhere

samhed:
Good comments, all fixed now!

vnc.html
CendioOssman: You need to fix the commit message as well. :)

samhed:
Fixed the commit message now.

PR #780, opened 2017-03-01
Autoconnect without url query parameters

mkunz7:
Connecting with
http://127.0.0.1/vnc.html?autoconnect=true&host=127.0.0.1&port=5901
is kind of ugly

With some quick javascript you can autoconnect without adding query parameters. If we provide this new users to novnc may find starting out easier. 

http://127.0.0.1/vnc_auto_ui.html

### DIFF
--- a/vnc_auto_ui.html
+++ b/vnc_auto_ui.html
@@ -1,0 +1,337 @@
+<!DOCTYPE html>
+<html>
+<head>
+
+    <!--
+    noVNC example: simple example using default UI that autoconnects
+    Copyright (C) 2012 Joel Martin
+    Copyright (C) 2016 Samuel Mannehed for Cendio AB
+    Copyright (C) 2016 Pierre Ossman for Cendio AB
+    noVNC is licensed under the MPL 2.0 (see LICENSE.txt)
+    This file is licensed under the 2-Clause BSD license (see LICENSE.txt).
+    Autoconnect parameters are set in a javascript block at the bottom of this file:
+        host = 127.0.0.1, port = 5901, password = "", path = "websockify"
+    -->
+    <title>noVNC</title>
+
+    <meta charset="utf-8" />
+
+    <!-- Always force latest IE rendering engine (even in intranet) & Chrome Frame
+                Remove this if you use the .htaccess -->
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+
+    <!-- Icons (see Makefile for what the sizes are for) -->
+    <link rel="icon" sizes="16x16" type="image/png" href="app/images/icons/novnc-16x16.png">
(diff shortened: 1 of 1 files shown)


samhed:
$ diff -u vnc.html vnc_auto_ui.html


    --- vnc.html	2017-03-02 08:25:05.426788266 +0100
    +++ vnc_auto_ui.html	2017-03-02 08:25:05.427788260 +0100
    @@ -3,17 +3,14 @@
     <head>
     
         
         <title>noVNC</title>
     
    @@ -325,7 +322,16 @@
         <script src="core/util.js"></script>
         <script src="app/webutil.js"></script>
         <script src="app/ui.js"></script>
    +    
    +    <script>
    +    window.onload = function () {
    +        document.getElementById("noVNC_setting_host").value = "127.0.0.1"
    +        document.getElementById("noVNC_setting_port").value = "5901";
    +        document.getElementById("noVNC_setting_password").value = "";
    +        document.getElementById("noVNC_setting_path").value = "websockify";
    +        UI.connect();
    +    }
    +    </script> 
         
    -
      </body>
     </html>


So you add a completely new file for that? You are free to do that for your own use, an addition like this is not interesting for this project.

mkunz7:
Yes the pr is minimal code. I think there's a bigger issue here that needs to be addressed. 

Even with renaming vnc_auto to vnc_lite you are still gonna have people trying to use it since it looks exactly like vnc_auto did. 

The comments suggest it uses the default ui and not a stripped down interface. The code also suggests it supports mobile (viewport), it does not. Those can be fixed in quick changes to lite, but the past mindset exists.

This is pretty widely used software and other people have integrated code using vnc_auto, the new release is gonna roll out and they aren't gonna like that their file is gone without a better alternative. 

My proposed solution is minimal it stacks on top the rename while alleviating some technical barrier to entry into novnc. Hopefully alleviate some user headaches down the road.

CendioOssman:
This kind of thing is probably what we want replacing vnc_auto.html long term. But we can't really maintain two almost identical files. This is probably more of a documentation issue. Perhaps an example in the wiki?

Using `onload` is also a bad idea as the scripts will load after that once #774 is in place. We might need a new mechanism for customisation like this.

samhed:
A better name will clarify what the file actually is. A simple name change is not very difficult to adapt to.

The word "default" does not have to mean "same as vnc.html". The interface in vnc.html has evolved over the years, vnc_auto has not. The comment clearly states that it is a simple example.

> The code also suggests it supports mobile (viewport), it does not.

Those are metadata lines, why would you read so much into that? vnc_auto works fine on iOS if you plug in a mouse and keyboard to your device.

I'm open for making it easier for people to change the default settings, but having different html files for every little variation is not a good approach.

mkunz7:
I'm all for better documentation on the matter and having a comment in the code refer to a doc on customization. Some additional cleanup to the comments in the HTML files would help too.

samhed:
Agreed!

PR #782, opened 2017-03-06
Use PNG for mouse cursors

### LINKED ISSUE #740: Windows 7 + VMware SVGA II driver.  black box where cursor should be
Using vncviewer for linux we have no issues.

Using novnc we can see a black box drawn around where the cursor should be.

If we do not install the VMWare SVGA II driver, novnc works, there is no black box.

The guest is a qemu windows vm (with a vmware vga adapter), using qemu's VNC option with websocket -> tcp proxy.

Using vnc_auto.html 0.6.0.

CendioOssman:
The browsers' support for Microsoft's cursor format is a bit spotty,
so use the more common PNG format instead. This also allows us to
use a Canvas to generate the image, rather than coding it by hand.

Fixes #740.

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -763,110 +763,52 @@
     ]);
 
     // Class Methods
-    Display.changeCursor = function (target, pixels, mask, hotx, hoty, w0, h0, cmap) {
-        var w = w0;
-        var h = h0;
-        if (h < w) {
-            h = w;  // increase h to make it square
-        } else {
-            w = h;  // increase w to make it square
+    Display.changeCursor = function (target, pixels, mask, hotx, hoty, w, h, cmap) {
+        if ((w === 0) || (h === 0)) {
+            target.style.cursor = 'none';
+            return;
         }
 
-        var cur = [];
-
-        // Push multi-byte little-endian values
-        cur.push16le = function (num) {
-            this.push(num & 0xFF, (num >> 8) & 0xFF);
-        };
-        cur.push32le = function (num) {
-            this.push(num & 0xFF,
-                      (num >> 8) & 0xFF,
-                      (num >> 16) & 0xFF,
-                      (num >> 24) & 0xFF);
-        };
-
-        var IHDRsz = 40;
-        var RGBsz = w * h * 4;
-        var XORsz = Math.ceil((w * h) / 8.0);
-        var ANDsz = Math.ceil((w * h) / 8.0);
-
-        cur.push16le(0);        // 0: Reserved
-        cur.push16le(2);        // 2: .CUR type
-        cur.push16le(1);        // 4: Number of images, 1 for non-animated ico
-
-        // Cursor #1 header (ICONDIRENTRY)
-        cur.push(w);            // 6: width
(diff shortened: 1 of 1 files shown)


core/display.js
```diff
@@ -763,110 +763,52 @@
...
+        if (SUPPORTS_IMAGEDATA_CONSTRUCTOR) {
+            img = new ImageData(new Uint8ClampedArray(cur), w, h);
+        } else {
+            img = ctx.createImageData(w, h);
+            img.data.set(new Uint8ClampedArray(cur));
         }
+        ctx.clearRect(0, 0, w, h);
+        ctx.putImageData(img, 0, 0);
 
-        var url = 'data:image/x-icon;base64,' + Base64.encode(cur);
```
samhed: Can we remove Base64.encode now?

CendioOssman: Still used in `imageRect()` unfortunately.

kanaka:
@CendioOssman This would certainly be a nice improvement. I do recall when I implemented this that browsers had mixed support for dataURLs and image formats, especially in the cursor style context. I'm sure this has improved in the years since though. Do you have any testing data? Seems like as of a year ago, IE only supported .cur and .ani image types: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_User_Interface/Using_URL_values_for_the_cursor_property In particular I would be interested if you've verified this for mobile browsers.

Apart from verifying browser compatibility, the change looks good.

samhed:
I'm not sure if I misunderstood something here, but IE doesn't support data URI's for cursors anyway?

kanaka:
@samhed you're right, I wasn't very clear. I was giving an example of the worst case (IE) as an example of the varied support even as of 1 year ago. But as long this does not regress any of the browsers we want to support, then I'm completely fine with it. I was just curious if we have any data points to confirm we haven't regressed (especially mobile where I've personally done basically 0 local cursor testing). When I implemented local cursor support originally (and it was a royal PITA), support was all over the place and what I ended up with was the least common denominator of functionality.

BTW, regarding the base64 issue, we could consider using window.atob/btoa for imageRect. However, this is one of those surprising cases where the JS implementation of Base64 was significantly faster than the builtin functions. And to top it off, if I'm recalling correctly (it's been a while) doing a direct comparison of simple microbenchmarks did not reveal the different as starkly. The performance difference may longer be there or it might be ever worse now with even more optimized JS JITs. But that's the reason why we have Base64 in the code base and aren't just using the builtin atob/btoa functions.

CendioOssman:
> I was just curious if we have any data points to confirm we haven't regressed (especially mobile where I've personally done basically 0 local cursor testing).

More testing is needed. I just first wanted to confirm that it solves the issue.

> BTW, regarding the base64 issue, ...

Long term we probably want to use blobs rather than data URIs, so it might not be worth changing until then.

DirectXMan12:
marking as v0.7.0 since this fixes a bug that we probably want fixed.

generally LGTM :+1: once testing confirms that it has broad support.

CendioOssman:
Tested on all the major browsers and it works well.

PR #789, opened 2017-03-11
How to connect vmmachine to novnc 

shubham436:
When iam try to get access of virtual machine it is only shows background Wallpaper and dont shows any icon or menubar


samhed:
What are you doing?

We already gave you some hints in other threads where you asked the exact same question.

PR #800, opened 2017-03-25
Allow users to specify CERT, WEB and PORT etc, through env variable

Neilpang:
Allow users set CERT etc by env variable.

### DIFF
--- a/utils/launch.sh
+++ b/utils/launch.sh
@@ -140,7 +140,7 @@ fi
 
 echo "Starting webserver and WebSockets proxy on port ${PORT}"
 #${HERE}/websockify --web ${WEB} ${CERT:+--cert ${CERT}} ${PORT} ${VNC_DEST} &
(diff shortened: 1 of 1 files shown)


DirectXMan12:
What's the use case here?  If you're doing this in production, you probably shouldn't be using launch.js anyway, and I'm not entirely certain as to why you can't just pass the flag.

Neilpang:
@DirectXMan12 

For example, if I run noVNC in docker,  I have to  set those values throw env variable.

Neilpang:
@DirectXMan12 
Here is my use case: 
https://github.com/Neilpang/novncbox#4-access-via-ssl

I need to specify `ssl_only` and cert file by the env variable.

CendioOssman:
Just modify `startup.sh` there to do what you need. It can easily add arguments based on environment.

PR #805, opened 2017-04-02
Fixed the issue where novnc can't be relocated

soichih:
If I try to serve noVNC under a sub directory, like
```
https://somewhere.com/noVNC/
```

It will fail due to browser-es-module-loader.js trying to load babel.js from root. This patch makes noVNC to be relocatable to sub-URL.

### DIFF
--- a/vendor/browser-es-module-loader/dist/browser-es-module-loader.js
+++ b/vendor/browser-es-module-loader/dist/browser-es-module-loader.js
@@ -1350,7 +1350,7 @@ WorkerPool.prototype = {
 };
 
 var promiseMap = new Map();
-var babelWorker = new WorkerPool('/vendor/browser-es-module-loader/dist/babel-worker.js', 3);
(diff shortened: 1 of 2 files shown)


CendioOssman:
Thanks good find. Unfortunately the files under `vendor/dist/` are generated files and should not be modified directly. Please look at `vendor/browser-es-module-loader/` instead. There's also a `README.md` there that explains how to generate things.

Lastly, please try to keep your initial commit line shorter. As you can see github does a weird truncation if it's too long, as do many other tools. :)

soichih:
@CendioOssman I've applied the fix on the src/. I also ran rollup and browserify, but it updated a tons of stuff on the dist (probably due to some version / environment change?) so I didn't include in my commit. All I am changing is to remove the / prefix from the URL.

Thanks for the tip on commit line!

CendioOssman:
I merged the two commits and did a new rollup here so the source map gets updated. It's now commited as dd44a8e. Thanks for your contribution.

PR #819, opened 2017-05-08
Fix license name in package.json

martinpitt:
It's spelled "MPL-2.0" according to https://spdx.org/licenses/.

This is important for license scanners as cockpit's (this came up in https://github.com/cockpit-project/cockpit/pull/5932).


samhed:
Thanks

PR #837, opened 2017-06-03
fix typscript error - the only one.

shaneharris:
TypeScript error: novnc/core/rfb.js(1012,17): Error TS7027: Unreachable code detected.

I was able to import the ES modules into typescript OK apart from this line. It is right the line is unreachable.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1009,7 +1009,7 @@ RFB.prototype = {
                 } else {
                     return this._fail("Authentication failure");
                 }
-                return false;
+                //return false; causes TypeScript error: novnc/core/rfb.js(1012,17): Error TS7027: Unreachable code detected.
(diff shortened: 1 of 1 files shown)


core/rfb.js
```diff
@@ -1009,7 +1009,7 @@ RFB.prototype = {
                 } else {
                     return this._fail("Authentication failure");
                 }
-                return false;
+                //return false; causes TypeScript error: novnc/core/rfb.js(1012,17): Error TS7027: Unreachable code detected.
```
DirectXMan12: Go ahead and just delete this line.

DirectXMan12:
thanks.  I'm glad we're importable from TypeScript (mildly surprised this was the only issue, but glad none-the-less ;-) )

PR #838, opened 2017-06-05
Fix Travis-Sauce Connection

DirectXMan12:
Somethings got messed up in the Travis-Sauce tunnel setup.  This should
fix it by re-adding the Sauce credentials.  It also updates the config
to explicitly pass auth information instead of via environment
variables.


DirectXMan12:
Will merge once tests re-run.  Travis is being slowish today.

PR #843, opened 2017-06-08
Remove hard coded Sauce authentication

CendioOssman:
This prevents forks to have their own Sauce accounts. The authentication
information should come from the Travis account configuration, not from
what's in the repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,4 @@ env:
   - TEST_BROWSER_NAME=safari TEST_BROWSER_OS='OS X 10.11'
 before_script: npm install -g karma-cli
 addons:
-  sauce_connect:
-    username:
-      secure: "Y5+WRK1ThvctFQNchjW2jly9eLrxe9k8wkKZhS2OGT3SDAmFZnJADswbk/jEbhtl+ZclXybz0uy8Y5YT1PzVCySOqwvy5viIvMZ1xXlOMt1wwJIYbsb3AbOTt3lS+GBXB7RSk0Um5nVx+IMDDIjfS+kQ53NeR1+a/RrDaVNfDEE="
-  jwt:
-    secure: "MHYV2PhoZtNdeSNgRQjLRSn8T9BmwvwBNTgzwOfrUrwU4kVH16ETWOCnygybuRtYC/LNwQxe5W0hcb4rqOzlOO+fWr7OUdRAg8MPXVK0v3WVRu0BpgDIqMXlAbSp2IyCADjqNxG4pmGyrlfAhJ5yO7moaHpjU4QkkXX6QrvjmH4="
-
+  sauce_connect: true


CendioOssman:
@DirectXMan12 ?

DirectXMan12:
apologies, I must have somehow missed this notification -- I just saw it.  Doing review now...

DirectXMan12:
I think this looks reasonable, assuming it actually works.  Let me configure the settings on Travis and sauce and check to see if this passes.  I seem to recall issues with it in the past...

DirectXMan12:
ok, env vars added to Travis, let me kick this PR and see if it works.

CendioOssman:
As one data point, I'm able to use sauce on my repo with this branch, but not with `master`. I've added the environment variables `SAUCE_ACCESS_KEY` and `SAUCE_USERNAME` and nothing else.

CendioOssman:
You can also trigger rebuilds from travis without fiddling with the project here. ;)

DirectXMan12:
> You can also trigger rebuilds from travis without fiddling with the project here. ;)

yep, that's what I did later ;-).

It doesn't appear to be working here, though -- I've added SAUCE_ACCESS_KEY and SAUCE_USERNAME to the env variables section of sauce.  Maybe they don't get updated until
a new PR is submitted?

DirectXMan12:
ahhh, there is an issue.  From the Travis docs:

> The most important restriction for pull requests is about secure environment variables and encrypted data.

> A pull request sent from a fork of the upstream repository could be manipulated to expose any environment variables. The upstream repository’s maintainer would have no protection against this attack, as pull requests can be sent by anyone with a fork.

> Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.

> Pull requests sent from forked repositories do not have access to encrypted variables or data.

I think this prevents us from using the access key and username in PR builds :-/

DirectXMan12:
Just to check, try pushing this branch to the main repo as a separate branch

CendioOssman:
If that was completely true then the currently checked in encrypted variables shouldn't work either?

Is it possible to remove this restriction?

CendioOssman:
> Just to check, try pushing this branch to the main repo as a separate branch

I pushed the branch `travis`, and it seems to be getting the environment variables.

DirectXMan12:
Yeah, I don't know if it's possible to remove this restriction.  I *think* it probably works with encrypted values because it's Travis that's interpreting them, so it's fine, because they can't leak, but anyone can access environment variables in the script, so that could leak.

CendioOssman:
I don't understand how. The point of the protection (as far as I can understand) is to protect against someone doing a pull request with this test script:

```sh
echo $SAUCE_ACCESS_KEY
```

In that case it doesn't matter if the secret comes from the repo or the config, it will be printed just the same.

Have you tried asking the Travis people?

CendioOssman:
This is what the settings look like for my Travis account (where this branch works fine):

screenshot from 2017-08-14 16-18-42

DirectXMan12:
> I don't understand how. The point of the protection (as far as I can understand) is to protect against someone doing a pull request with this test script:

When it's injected directly in the config, it's not injected via an environment variable, but rather a config option for the sauce addin.  Since that's parsed and decrypted by Travis and passed directly to Sauce by the Travis internals, it's not exposed elsewhere.

CendioOssman:
Ah, I see. That makes sense.

Still, I think we should try to talk to the Travis people. We want the repo configuration to be the fallback, not the primary source of the secrets. Otherwise we discourage people from proper testing before they send pull requests.

DirectXMan12:
> Still, I think we should try to talk to the Travis people

Yeah, it's worth a shot to see if they have anything in the works.  Are you fine closing this PR, or do you want to leave it open?

CendioOssman:
We can close this for now since it doesn't solve all our issues.

PR #885, opened 2017-08-28
Add FreeBSD support and remove bashisms from launch.sh

ocochard:
Small update to launch.sh.
Does the replacement of the [[ ]] by simple [ ] for being compliant with a standard /bin/sh should be reviewed by some security peoples ?

### DIFF
--- a/utils/launch.sh
+++ b/utils/launch.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # Copyright 2016 Joel Martin
 # Copyright 2016 Solly Ross
@@ -36,6 +36,7 @@ CERT=""
 WEB=""
 proxy_pid=""
 SSLONLY=""
+PORTUSED=false
 
 die() {
     echo "$*"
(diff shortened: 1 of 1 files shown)


utils/launch.sh
```diff
@@ -73,8 +74,12 @@ done
 which netstat >/dev/null 2>&1 \
     || die "Must have netstat installed"
 
-netstat -ltn | grep -qs ":${PORT} .*LISTEN" \
-    && die "Port ${PORT} in use. Try --listen PORT"
+case $(uname -s) in
+   "FreeBSD") netstat -lan | grep -qs ".${PORT} .*LISTEN" && PORTUSED=true ;;
+   "Linux") netstat -ltn | grep -qs ":${PORT} .*LISTEN" && PORTUSED=true ;;
```
CendioOssman: `netstat -lan` seems to work on Linux as well, so you should be able to merge these instead.

utils/launch.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
CendioOssman: Why restrict ourselves to just Posix shell? Requiring bash is hardly a massive burden.

CendioOssman:
No response.

PR #888, opened 2017-09-01
Clipboard setting

### LINKED ISSUE #859: Ability to disable clipboard integration
I have a use case where an application I am testing becomes unresponsive if there is a nvc client connected sharing the clipboard. I am able to reproduce the issue using realvnc viewer, but if I disable send clipboard changes to server, the issue stops.

I tried looking into the RFB source as well as the client but didn't see anything about the clipboard at the RBF level. I tried disabling the events for clipboard, but that didn't help. 

I am guessing is a property that has to be sent with the connection.

Please let me know if there is a way to work around this or what we would need to do to implement this.

image

Thanks,

Hansel

samhed:
Simple addition that adds a setting for clipboard.

Fixes #859

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -38,7 +38,6 @@ var UI = {
     controlbarMouseDownOffsetY: 0,
 
     isSafari: false,
-    rememberedClipSetting: null,
     lastKeyboardinput: null,
     defaultKeyboardinputLen: 100,
 
--- a/app/ui.js
+++ b/app/ui.js
@@ -383,6 +383,8 @@ var UI = {
         UI.addSettingChangeHandler('encrypt');
         UI.addSettingChangeHandler('cursor');
         UI.addSettingChangeHandler('cursor', UI.updateLocalCursor);
+        UI.addSettingChangeHandler('clipboard');
+        UI.addSettingChangeHandler('clipboard', UI.updateClipboard);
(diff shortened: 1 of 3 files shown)


CendioOssman:
When is this needed? The clipboard sync isn't automatic in noVNC, so a user that doesn't want to use the clipboard can just ignore it.

samhed:
@hanselb seemed to have found some application that was misbehaving due to clipboard send/recieve in #859 .

CendioOssman:
I'm very sceptical about cluttering up the interface in order to work around bugs. Especially if the bug isn't even in noVNC. We should explore other options. E.g.

 a) Getting the bug fixed instead of papered over
 b) Disabling the clipboard in the server, or some other workaround closer to the misbehaving application
 c) Investigate why this is even happening, since no clipboard transfer should occur if the user isn't interacting with the clipboard field

samhed:
Sounds reasonable to me

DirectXMan12:
I don't mind the UI changes, but I'm not certain the changes to `core/rfb.js` are needed.  We should just be able to set the clipboard field to disabled, or something like that.

samhed:
As per @CendioOssman 's comments, I don't think merging this is the way forward. Let's see if we get an answer in #859 . I'm closing this.

Issue #860, opened 2017-07-15
Connecting to Intel AMT fails


grepaly:
Dear All,

is there a trick to connect to Intel AMT? What i see on the server side is: "handler exception: [Errno 107] Transport endpoint is not connected". On the browser side: Server native red-shift is not 16 (warning) and WebSocket on-close event. It does ask for the password and it says disconnected.

CendioOssman:
Those kind of servers generally don't fully support the VNC protocol, so I would suspect it kicks out noVNC when it asks for a different pixel format. Have you tried some other VNC clients?

DirectXMan12:
Yeah, noVNC tends to be a bit strict with the pixel formats that it supports, for performance reasons (because most "normal" VNC servers use one or two different ones, so we can get away with it).  I believe the iKVM support PR expands this support a bit, so you might want to try with that branch (see #614 and friends).

haqa:
If you go back to v0.6.2 it works fine. I suspect a more recent change is incompatible with the AMT VNC KVM Redirector.

DirectXMan12:
huh, if that's the the case, then it's a regression.  Do you think you could send us a recording (run websockify with the `--record` option) so that we could track it down?

haqa:
Regrettably I'm not allowed to do that (This is being used within a corporate environment). If I can get access to an AMT device outside of work I'll post a dump here.

kimsto:
I think this is directly related to removing colormap support in commit 26586b9.
This removed the explicit test for AMT to disable true_color mode:
if (this._true_color && this._fb_name === "Intel(r) AMT KVM") {

CendioOssman:
Well that is annoying. So the question is do we want to fix and keep all that colour map code just to support Intel AMT? (which is technically in violation of the VNC protocol by doing this)

Does this happen on current systems as well, or just older ones?

TigerVNC has also dropped colour map support, so which client to people normally use for these systems?

DirectXMan12:
looks like there are still a few color map servers floating around these days.  We might have to re-implement that support (see also #882)

haqa:
My personal feeling is that removing compatibility with hardware implementations of the server is just making peoples lives hard.

Oracle (and then Chrome and Firefox) did this with Java and overnight hundreds of thousands of DRAC and iLO servers became inaccessible.

The strength and power of Open Source projects such as this is that they are run by people who know doing this is a bad thing, and so don't.

(Just my opinion)

[...]

PR #893, opened 2017-09-08
Basic support for Intel AMT

CendioOssman:
This restores basic support for Intel AMT servers. They refuse clients that request more than 16 bits per pixels, so implement a fallback in just the "Raw" encoding.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1971,7 +1971,7 @@ RFB.encodingHandlers = {
         return true;
     },
 
-    display_tight: function (isTightPNG) {
+    TIGHT: function () {
         this._FBU.bytes = 1;  // compression-control byte
         if (this._sock.rQwait("TIGHT compression-control", this._FBU.bytes)) { return false; }
 
@@ -2203,11 +2203,6 @@ RFB.encodingHandlers = {
                                "Illegal tight compression received, " +
                                "ctl: " + ctl);
 
-        if (isTightPNG && (cmode === "filter" || cmode === "copy")) {
-            return this._fail("Unexpected server message",
-                              "filter/copy received in tightPNG mode");
-        }
-
         switch (cmode) {
             // fill use depth because TPIXELs drop the padding byte
             case "fill":  // TPIXEL
@@ -2282,9 +2277,6 @@ RFB.encodingHandlers = {
 
         return true;
     },
-
-    TIGHT: function () { return this._encHandlers.display_tight(false); },
-    TIGHT_PNG: function () { return this._encHandlers.display_tight(true); },
 
     last_rect: function () {
         this._FBU.rects = 0;
@@ -2398,12 +2390,4 @@ RFB.encodingHandlers = {
         } catch (err) {
         }
     },
-
-    JPEG_quality_lo: function () {
-        Log.Error("Server sent jpeg_quality pseudo-encoding");
-    },
-
-    compress_lo: function () {
-        Log.Error("Server sent compress level pseudo-encoding");
-    }
 };
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -22,6 +22,7 @@ import DES from "./des.js";
 import KeyTable from "./input/keysym.js";
 import XtScancode from "./input/xtscancodes.js";
 import Inflator from "./inflator.js";
+import { encodings, encodingName } from "./encodings.js";
 
 /*jslint white: false, browser: true */
 /*global window, Util, Display, Keyboard, Mouse, Websock, Websock_native, Base64, DES, KeyTable, Inflator, XtScancode */
@@ -76,7 +77,6 @@ export default function RFB(defaults) {
     ];
 
     this._encHandlers = {};
-    this._encNames = {};
     this._encStats = {};
 
     this._sock = null;              // Websock object
@@ -183,7 +183,6 @@ export default function RFB(defaults) {
     // Create lookup tables based on encoding number
     for (var i = 0; i < this._encodings.length; i++) {
         this._encHandlers[this._encodings[i][1]] = this._encHandlers[this._encodings[i][0]];
-        this._encNames[this._encodings[i][1]] = this._encodings[i][0];
         this._encStats[this._encodings[i][1]] = [0, 0];
     }
 
@@ -445,14 +444,14 @@ RFB.prototype = {
         for (i = 0; i < this._encodings.length; i++) {
             s = this._encStats[this._encodings[i][1]];
             if (s[0] + s[1] > 0) {
-                Log.Info("    " + this._encodings[i][0] + ": " + s[0] + " rects");
+                Log.Info("    " + encodingName(this._encodings[i][1]) + ": " + s[0] + " rects");
             }
         }
 
         Log.Info("Encoding stats since page load:");
         for (i = 0; i < this._encodings.length; i++) {
             s = this._encStats[this._encodings[i][1]];
-            Log.Info("    " + this._encodings[i][0] + ": " + s[1] + " rects");
+            Log.Info("    " + encodingName(this._encodings[i][1]) + ": " + s[1] + " rects");
         }
     },
 
@@ -1348,7 +1347,8 @@ RFB.prototype = {
                     {'x': this._FBU.x, 'y': this._FBU.y,
                      'width': this._FBU.width, 'height': this._FBU.height,
                      'encoding': this._FBU.encoding,
-                     'encodingName': this._encNames[this._FBU.encoding]});
+                     'encodingName': encodingName(this._FBU.encoding)});
+            }
 
                 if (!this._encNames[this._FBU.encoding]) {
                     this._fail("Unexpected server message",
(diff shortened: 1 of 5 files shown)


CendioOssman:
This is an attempt at fixing #860.

samhed:
Looks nice! Great with the encoding cleanup as well. As you mentioned before, it would be nice to split the encoding handlers to separate modules/files as well.

DirectXMan12:
I'd like to do a review of this before merge.  Can't promise that I'll get in done today, but I'll try and get it done by the end of this week or early next week.

CendioOssman:
Someone also still needs to test this as I haven't been able to get my hands on an AMT device.

core/rfb.js
```diff
@@ -1971,7 +1971,7 @@ RFB.encodingHandlers = {
         return true;
     },
 
-    display_tight: function (isTightPNG) {
+    TIGHT: function () {
```
DirectXMan12: Does the server not send the TIGHT_PNG encoding explicitly?  If so, we should probably have comments on which encodings are just for advertising capabilities, and which are for actually being sent by the server.

core/rfb.js
```diff
@@ -1788,6 +1772,34 @@ RFB.genDES = function (password, challenge) {
 };
 
 RFB.encodingHandlers = {
+    getHandler: function (rfb, encoding) {
+        switch (encoding) {
```
DirectXMan12: Please, let's not do this.  We should have a map of some sort (whether via the Map type or just a Javascript object) and do lookup in there, instead of having a giant switch statement.  If the names of the functions in the encodingHandlers object isn't correct, then we should fix the names.

core/rfb.js
```diff
@@ -1788,6 +1772,34 @@ RFB.genDES = function (password, challenge) {
...
+            case encodings.encodingRaw:
+                return RFB.encodingHandlers.RAW.bind(rfb);
```
DirectXMan12: IIRC, calling bind isn't a particularly fast operation.  Since this could be called in a hotloop, we really should pre-bind.

core/rfb.js
```diff
@@ -1675,33 +1685,45 @@ RFB.messages = {
         sock.flush();
     },
 
-    pixelFormat: function (sock, bpp, depth, true_color) {
+    pixelFormat: function (sock, depth, true_color) {
```
DirectXMan12: what's the reasoning behind dropping the `bpp` argument here?  I'd prefer to be more explicit here, like we were doing before.

core/rfb.js
```diff
@@ -1805,19 +1827,34 @@ RFB.encodingHandlers = {
             this._FBU.lines = this._FBU.height;
         }
 
-        this._FBU.bytes = this._FBU.width * 4;  // at least a line
+        var pixelSize = this._fb_depth == 8 ? 1 : 4;
```
DirectXMan12: it feels hacky to hardcode this.  We should calculate it from by dividing depth by 8, and then cache if necessary (or change depth to be bytes, not bits).

core/rfb.js
```diff
@@ -1805,19 +1827,34 @@ RFB.encodingHandlers = {
...
+        this._FBU.bytes = this._FBU.width * pixelSize;  // at least a line
         if (this._sock.rQwait("RAW", this._FBU.bytes)) { return false; }
         var cur_y = this._FBU.y + (this._FBU.height - this._FBU.lines);
         var curr_height = Math.min(this._FBU.lines,
-                                   Math.floor(this._sock.rQlen() / (this._FBU.width * 4)));
+                                   Math.floor(this._sock.rQlen() / (this._FBU.width * pixelSize)));
+        var data = this._sock.get_rQ();
+        var index = this._sock.get_rQi();
+        if (this._fb_depth == 8) {
```
DirectXMan12: again, it feels a bit weird to special case 8 bits, but this is probably better perf-wise.  Maybe in the long term we should think about a better solution.

core/rfb.js
```diff
@@ -1805,19 +1827,34 @@ RFB.encodingHandlers = {
...
+            var pixels = this._FBU.width * curr_height
+            var newdata = new Uint8Array(pixels * 4);
```
DirectXMan12: ooof.  Allocations like this are *slow* (like, really slow in some cases), that's why we preallocate a buffer in a bunch of places.  We should do that here, and grow if necessary.

CendioOssman:
> I'm a bit uneasy about hardcoding a special case for 8 bits, and only on RAW. It seems like it might be easy to loose track of.

The idea is to avoid having to force all code having to deal with this non-standard corner case. IOW, get AMT to work with as small maintenance cost as possible.

core/rfb.js
CendioOssman: Yup. That's why I tagged it as a pseudo-encoding and not a "real" encoding. It's similar to how the quality "encodings" turn on JPEG in Tight.

core/rfb.js
CendioOssman: We weren't explicit before either as only some of the necessary values were provided and the rest were hard coded (possibly creating conflicting info). The idea here was to make it a bit more consistent in that we derive most fields from what is provided.

core/rfb.js
CendioOssman: But depth 24 bits has a pixel size of 32 bits, so we have to have special code for this.

core/rfb.js
CendioOssman: The idea is to make 8 bit the only exception to code that otherwise assumes 24 bit. But perhaps we can make that more explicit. E.g. `this._low_bit_mode`?

core/rfb.js
CendioOssman: This is Raw, so it's going to have mediocre performance anyway. :)
So ideally I'd prefer simpler code.

core/rfb.js
CendioOssman: I don't think it's a good idea to restrict the code to follow the naming of protocol constants. But would you prefer something like this in the init routine:

```Javascript
this._encoding_handlers = []
this._encoding_handlers[encodings.encodingRaw] = RFB.encodingHandlers.RAW.bind(this)
...
```

core/rfb.js
CendioOssman: I seem to recall it's the calling of bound functions that is slow (*was* slow most likely these days), so I'm not sure that will change much. However my suggestion on the previous comment would perform that change as a side effect.

core/rfb.js
DirectXMan12: ack, alright.  Cool

core/rfb.js
DirectXMan12: yeah.  I'd either prefer that or just having something like
`RFB.encodingHandlers.registerEncoding(encodings.encodingRaw, function(...) { ... })`.  I don't like the giant switch statement, and the ability to register new encodings dynamically is nice.

P.S. won't it be nice when we can drop IE support in favor of edge at some nebulous time in the future and just write `RFB.encodingHandlers = { ..., [encodings.encodingRaw]: function(...) { ... } }`).  All the other browsers support it.

core/rfb.js
DirectXMan12: We used to have a difference between depth and pixel size, and we passed that in explicitly, so the normal combo was `(4, 3)` (4 bytes per pixel, 3 bytes of depth).  I'd prefer explicitly passing those as separate arguments.

core/rfb.js
DirectXMan12: nice pun (a *bit* more consistent), but I don't think I'm following.  We did hard-code the color locations, but we don't have to do that.  Passing in bpp makes it more explicit that we're choosing this value for pixel size, and this other value for depth.

DirectXMan12:
> The idea is to avoid having to force all code having to deal with this non-standard corner case. IOW, get AMT to work with as small maintenance cost as possible.

Ack.

core/rfb.js
CendioOssman: But they're not really independent. There's pretty much only a single correct bpp for each depth value. The function should either:

 a) Be really stupid and just send what is given
 b) Be somewhat high level and ask only for the values it cannot derive itself

Right now we've had some mix of the two. This patch pushes it further along the b) route.

core/rfb.js
CendioOssman: See above. They're not independent and having both as arguments give the illusion that they are.

core/rfb.js
CendioOssman: One can dream. :)
Unfortunately Edge isn't much better in many regards...

I've pushed a new version now with the suggested model. Please have a look.

core/rfb.js
DirectXMan12: ack.

DirectXMan12:
:-/ Travis is borked again.  Let me take a look

DirectXMan12:
*much* better.  Feel free to merge once you get someone with an actual AMT box to test.

haqa:
Yes, I can confirm that works on AMT on an Intel NUC. Thank you.

grepaly:
I can also confirm that it is working. Thanks!

samhed:
Great! Thanks for testing

PR #896, opened 2017-09-12
Combine small mouse wheel events

### LINKED ISSUE #577: Scrolling with a trackpad is extremely sensitive on OS X and Chrome OS
Scrolling using a trackpad in noVNC running on Safari or Google Chrome (OS X 10.11) is perhaps 10 times as sensitive as when scrolling locally.

The problem does not seem to exist in Firefox on OS X.

samhed:
The VNC protocol can't handle different deltas or speeds for a mouse wheel event. When using a device that sends a lot of small mouse wheel events, instead of fewer larger steps, the effect was that mouse wheel scrolling was way to sensitive.
This patch looks at the delta of wheel events and doesn't send events until the combined delta has passed a threshold. Single events that doesn't pass the threshold get sent after a timeout in order to not loose any events.

Fixes #577


core/input/keyboard.js
```diff
@@ -10,7 +10,7 @@
 
 import * as Log from '../util/logging.js';
 import { isTouchDevice } from '../util/browsers.js';
```
CendioOssman: Did you check these imports? `isTouchDevice` doesn't seem like something the keyboard code would need?

core/input/keyboard.js
```diff
@@ -354,222 +354,4 @@ make_properties(Keyboard, [
...
-    ['target',         'ro', 'dom'],   // DOM element that captures mouse input
-    ['focused',        'rw', 'bool'],  // Capture and send mouse clicks/movement
-
-    ['onMouseButton',  'rw', 'func'],  // Handler for mouse button click/release
-    ['onMouseMove',    'rw', 'func'],  // Handler for mouse movement
-    ['touchButton',    'rw', 'int']    // Button mask (1, 2, 4) for touch devices (0 means ignore clicks)
-]);
-
-export { Keyboard, Mouse };
+export { Keyboard };
```
CendioOssman: Perhaps these should be default exports now that it's just one thing in each?

core/input/mouse.js
```diff
@@ -12,12 +12,20 @@ import * as Log from '../util/logging.js';
...
 
     this._doubleClickTimer = null;
     this._lastTouchPos = null;
 
+    this._wheelStepTimer = null;
+    this._accumulatedWheelDeltaX = 0;
+    this._accumulatedWheelDeltaY = 0;
+
+    this.WHEEL_STEP = 10; // Delta threshold for a mouse wheel step
+    this.WHEEL_LINE_HEIGHT = getScrollLineHeight();
```
CendioOssman: These are constants I assume? So why not put them at the top level?

core/input/mouse.js
```diff
@@ -108,27 +116,74 @@ Mouse.prototype = {
...
-            }
+        var dX = e.deltaX;
+        var dY = e.deltaY;
+
+        // Pixel units unless it's non-zero.
+        // Note that if deltamode is line or page won't matter since we aren't
+        // sending the mouse wheel delta to the server anyway.
+        // The difference between pixel and line can be important however since
+        // we have a threshold that can be smaller than the line height.
+        if (e.deltaMode) {
```
CendioOssman: This makes it look like it's a boolean. I would prefer a `!== 0`.

core/input/mouse.js
```diff
@@ -108,27 +116,74 @@ Mouse.prototype = {
         this._handleMouseButton(e, 0);
     },
 
+    // Mouse wheel events are sent in steps over VNC. This means that the VNC
+    // protocol can't handle a wheel event with specific distance or speed.
+    // Therefor, if we get a lot of small mouse wheel events we combine them.
+    _generateWheelStepX: function (pos) {
```
CendioOssman: Perhaps we should keep the position stored on the object rather than passed along as an argument? In theory the mouse could have moved before the timer fires.

core/input/mouse.js
```diff
@@ -108,27 +116,74 @@ Mouse.prototype = {
...
+
+        this._accumulatedWheelDeltaY = 0;
+    },
+
     _handleMouseWheel: function (e) {
-        if (!this._focused) { return; }
+        if (!this._focused || !this._onMouseButton) { return; }
+
+        window.clearTimeout(this._wheelStepXTimer);
+        window.clearTimeout(this._wheelStepYTimer);
```
CendioOssman: Don't you need to clear the variables as well in order to avoid the risk of stopping unrelated timers the next round?

core/input/mouse.js
```diff
@@ -108,27 +116,74 @@ Mouse.prototype = {
...
+
+        // Generate a mouse wheel step event when the accumulated delta
+        // for one of the axes is large enough.
+        // Small delta events that do not pass the threshold get sent
+        // after a timeout.
+        if (Math.abs(this._accumulatedWheelDeltaX) > this.WHEEL_STEP) {
+            this._generateWheelStepX(pos);
+        } else {
+            this._wheelStepXTimer =
+                window.setTimeout(this._generateWheelStepX.bind(this, pos), 50);
```
CendioOssman: Since we already have constants, then perhaps this 50ms should also be a constant? Keeps it in sync for the two calls if nothing else.

core/input/mouse.js
CendioOssman: We also need to clear these timers on "ungrab".

core/input/util.js
```diff
@@ -14,6 +14,20 @@ function isEdge() {
     return navigator && !!(/edge/i).exec(navigator.userAgent);
 }
 
+// Get the height of a un-styled line of text
+export function getScrollLineHeight() {
```
CendioOssman: Urgh. Is this really necessary? Can't we have something hard coded? Or evaluate some CSS?

CendioOssman: Since this is rather heavy it should do the work when loaded and store the result in a variable, which can then efficiently be returned on each call.

core/input/mouse.js
```diff
@@ -108,27 +116,74 @@ Mouse.prototype = {
         this._handleMouseButton(e, 0);
     },
 
+    // Mouse wheel events are sent in steps over VNC. This means that the VNC
+    // protocol can't handle a wheel event with specific distance or speed.
+    // Therefor, if we get a lot of small mouse wheel events we combine them.
+    _generateWheelStepX: function (pos) {
+
+        if (this._accumulatedWheelDeltaX < 0) {
```
CendioOssman: Any scenario where we might need to send multiple events?

core/input/keyboard.js
samhed: I did check them but must've missed that one, good eye!

core/input/keyboard.js
samhed: Yes, I'll fix that.

core/input/mouse.js
samhed: Yes. You mean outside the mouse object? I'll move them

core/input/mouse.js
samhed: Agreed, that feels more natural.

core/input/mouse.js
samhed: Absolutely, but I think that is another discussion. That could make the behavior deviate from how things worked previously, even when using an input device that does not send many small delta events.

A goal for this was to keep the old scroll wheel behavior where it wasn't too sensitive.

core/input/mouse.js
samhed: Good point!

core/input/mouse.js
samhed: Fair enough

core/input/mouse.js
samhed: +1

core/input/mouse.js
samhed: I tested it quite extensively and even printing timestamps, it seemed to behave properly. But what do you mean with unrelated timers?

core/input/util.js
samhed: good idea!

samhed: Depends on how accurate we want to be. This is the only way to get the correct value.

core/input/mouse.js
CendioOssman: Timer ids are simple integers, so it is possible for re-use. So if you get id 45 and then clear it, some other code might register a new timeout and also get 45 since that id is now free.

core/input/util.js
CendioOssman: Do we need, or even want, accurate value for this though? VNC doesn't concern itself with pixels or lines, so line height isn't the important value here. What we really want is the physical distance the scroll wheel (or finger on a touch pad) moved. And from what I understood Firefox always sent the same line count for the same physical distance, no matter the line height?

core/input/mouse.js
samhed: Oh, I thought the id was a longer almost unique identifier. I'll just null the variables after clearing the timeout then.
Seems like null-checks are not needed since the functions don't error when called with null.

core/input/util.js
samhed: That is absolutely true, I just verified it. Let's just use the default value (19) then.

DirectXMan12:
Modulo @CendioOssman's comments, this looks good (@CendioOssman basically said most of what I would have said anyway ;-) )

samhed:
Updated according to the comments.

core/input/mouse.js
```diff
@@ -108,27 +117,82 @@ Mouse.prototype = {
...
 
-        if (this._onMouseButton) {
-            if (e.deltaX < 0) {
-                this._onMouseButton(pos.x, pos.y, 1, 1 << 5);
-                this._onMouseButton(pos.x, pos.y, 0, 1 << 5);
-            } else if (e.deltaX > 0) {
-                this._onMouseButton(pos.x, pos.y, 1, 1 << 6);
-                this._onMouseButton(pos.x, pos.y, 0, 1 << 6);
-            }
+        this._wheelPos = this._getMousePosition(e);
```
CendioOssman: This doesn't really solve the position problem as there might be a normal movement of the mouse before the timer fires. You should keep track of the last known position, no matter what event it was.

(and the naming should probably reflect that as well :))

CendioOssman:
Oh, right. You also forgot to fix the tests. Didn't you use those while developing? ;)

Should probably add some new tests for this scenario.

core/input/mouse.js
samhed: Right, so basically update the position in all mouse handlers?

samhed:
Yeah, I'll fix the broken tests and add some new ones

core/input/mouse.js
CendioOssman: Indeed. Perhaps the helper can be changed for this?

samhed: how do you mean?

samhed: How do you mean?

CendioOssman: We have `_getMousePosition()`. Perhaps it should be changed to something like `_updateMousePosition()`?

samhed: right, of course

samhed:
All done

PR #905, opened 2017-09-29
Only load ES6 module loader polyfill if needed

qmfrederik:
Safari and Chrome now have native support for modules, so we need to make sure we only load our polyfill when actually needed.


qmfrederik:
This is a port of 36bdf136541599c0b7620cdece2cd6639d4b90e3 to `vnc_lite.html`

Issue #692, opened 2016-10-19
IE and Edge unusable - Pressing SPACE toggles UI buttons


samhed:
After you have used any button in the control bar SPACE will, when pressed, toggle that button. It continues to happen even after you have used the mouse inside the VNC screen. It only stops happening if you click somewhere outside any button in the UI, or if you click on any element that belongs to the browser GUI.

Even though the SPACE is being sent to the server, the toolbar buttons are still being toggled. For example, if the button you most recently pressed was "connect" and you want to send a SPACE to the server, you will disconnect. This is a horrible experience as you might imagine and I am extremely surprised that we haven't had anyone reporting this before.

Found using IE 11 and Edge 38 on Windows 10. Can reproduce using both using code from 2 years ago and with the current master with the new UI, so this is not a regression in noVNC.

samhed:
Quick tests seem to indicate that this does not happen on Windows 10 mobile. It does however happen both on a microsoft surface device with a touch screen, and on a computer without touch screen.

geocode:
I am seeing a somewhat similar issue with F10 and firefox.  In some but not all instances of firefox (we have not been able to come up with a consistent work around).  F10 will display the firefox menu instead of sending F10 to the application on the other end of the vnc connection.  Sometimes it works, sometimes it doesn't.  Does this sound like it is caused by the same issue that is scheduled for v0.7.0?  Any help on this would be greatly appreciated!

samhed:
> Does this sound like it is caused by the same issue that is scheduled for v0.7.0?

No, not at all

samhed:
I have no direct idea for how to work around this. I propose to remove this from the 0.7 milestone.

samhed:
This is not a regression, it's been like this for ages. Fixing this will likely require some major rethinking of focus handling. Let's postpone this.

samhed:
Note that with master I can still reproduce the issue on Edge but not on IE. I used private browsing and inspected the code to make sure IE was running master and it sure seems like the issue doesn't exist on IE anymore.

Anyway, still a problem on Edge, and #916 fixes it!

CendioOssman:
Fixed by #916.

Issue #915, opened 2017-10-06
Arrow keys affect setting elements in Firefox


CendioOssman:
Click on the scaling mode setting, and then try to use the up or down key in the session. This results in the scaling mode changing state when the key should only be sent to the session.

I suspect this is a general focus issue, so it is probably related to #914 and #692.

Tested with Firefox 55 on Linux.

CendioOssman:
Fixed by #916.

Issue #914, opened 2017-10-06
Context menu isn't blocked in Firefox


CendioOssman:
Pressing the context menu key results in a local menu popping up in Firefox.

Tested with Firefox 55 on Linux.

Possibly related to #692.

DirectXMan12:
is this a 1.0 blocker?

CendioOssman:
For the problem itself, I'd say no. But #916 does fix some API warts which I think we want to get rid of before 1.0.

CendioOssman:
Fixed by #916.

PR #916, opened 2017-10-06
Properly give focus to canvas

CendioOssman:
Possible fix for #914, #915 and #692. Needs testing.


CendioOssman:
Need to check how it works on mobile devices where that extra input field is used as well.

CendioOssman:
Works well with Chrome and Firefox on Pixel C with on screen keyboard. Physical keyboard has some quirks though:

 a) Touching the canvas doesn't move focus there (need to add a `touchbegin` listener)
 b) It's not readily apparent that the focus is on input elements

samhed:
Running this makes IE and Edge work well! Can verify that it fixes #692.

One should also note that this PR brings a change in behavior for users. If the user clicks in the settings and then proceeds to type with the keyboard, the input will no longer be sent to the server. The user has to change focus by clicking on the canvas first.

samhed:
You have only seen issues with a physical keyboard on android so far right?

CendioOssman:
Yup. But there's also more testing to be done.

I also realised it changed the behaviour of the "touch" input field in that we'll no longer listen to key events on that. IOW it becomes the normal path rather than a fallback.

CendioOssman:
Works well on all browsers and platforms now. Even tested physical keyboard on Android and iOS.

samhed:
nice!

PR #920, opened 2017-10-10
updated german translations

flumm:
Signed-off-by: Dominik Csapak <d.csapak@proxmox.com>


samhed:
Thanks, but the javascript  (.json file) should be updated as well. You can find instructions here: https://github.com/novnc/noVNC/wiki/Translations#updating

flumm:
@samhed done

PR #922, opened 2017-10-11
Update Greek translation

kosmasgiannis:
.po and .json files updated.


samhed:
I saw now that there are still 2 strings marked as fuzzy, but it looks like the strings were updated. I guess you edited the files by hand and just forgot to remove the fuzzy-marks?

kosmasgiannis:
Yes, you are right, I edited them by hand and did't pay attention to the fuzzy marks. 
I 'll remember next time.

samhed:
Ok, no problem, I'll remove the marks then.

Issue #924, opened 2017-10-12
Finalize public API


samhed:
Before we release a 1.0 of noVNC we really should take a look at our public API, mainly on the RFB module. If we want to remove or change anything, now is the time.

A few things that we might want to think about:
* Do we need all of these public functions?
* Are we satisfied with all the names?
* Use the `Event` type for our callbacks. More in line with general JavaScript API's. Also, currently with the "onEvent" style we can only have one listener at a time.
* Can we try to bring in #916 before a release? It fixes a number of known issues but changes the API a bit

PS:

I think that we should stop looking at Mouse, Keyboard and Display the same way. In my eyes the main API we have towards other applications that use our code is the RFB module. Mouse, Keyboard and Display are more internal help-modules. Therefore I don't think we have to preserve those API's as much. I don't think we need to document the API of Mouse, Keyboard and Display either.

kanaka:
Short version: I have no objection to making all the modules internal, although I think there is value in the Keyboard module that make it interesting to other people/projects.

Longer version:

My original intention was that each of these modules would be generic enough that others could use them independently. I do think there is value to the maintainability of the project to keep those modules modular and independent of each other module. Documenting the public API of each module was a way of keeping myself honest about that modularity.

However, at this point, I don't think any external projects would be likely use Mouse or Display. There are a lot of projects/frameworks that have similar functionality and the current Mouse and Display modules are somewhat specific to what we need in noVNC. On the other hand, I think all the effort that we have put into the Keyboard module to work around browser limitations and quirks is fairly unique and solves a problem that a number of different types of web applications have (and will continue to have in the future). So I still think there is some value in keeping it documented for possible external use. However, that would require a level of commitment from our team and if none of us want to own that, then we don't want to mislead people into thinking it's something it's not (and then filing bugs that would just be noise for us).

On the topic of RFB module APIs: I recently ran across https://github.com/mikeal/webtorrent-element. I think something like that might be interesting for noVNC as a future feature.

CendioOssman:
> On the topic of RFB module APIs: I recently ran across https://github.com/mikeal/webtorrent-element. I think something like that might be interesting for noVNC as a future feature.

If you are referring to the shadow DOM thing, then yes please. :)
I'm keeping my eye on that an eagerly awaiting when it is rolled out in most browsers.

CendioOssman:
> * Can we try to bring in #916 before a release? It fixes a number of known issues but changes the API a bit

[...]

PR #927, opened 2017-10-13
Adjust public API

CendioOssman:
These are the changes I'd like to make before closing #924 and releasing 1.0. I'm still looking over things, but this branch should include what I've found so far.


core/rfb.js
```diff
@@ -243,15 +240,13 @@ export default function RFB(defaults) {
...
+    connect: function (url, creds) {
+        this._url = url;
         this._rfb_credentials = (creds !== undefined) ? creds : {};
-        this._rfb_path = (path !== undefined) ? path : "";
 
-        if (!this._rfb_host) {
-            return this._fail(
-                _("Must set host"));
+        if (!url) {
+            this._fail(_("Must specify URL"));
```
samhed: 👎 translation change :(

CendioOssman: Yes, that's unfortunate. But it's not a major issue for most users as we have a check in `ui.js` as well. So it is people using `RFB` directly that will be affected. And we haven't sorted out translations in that scenario.

CendioOssman:
I'm not entirely happy with the capabilities API, so I'll have another look at that once everything else is done.

core/rfb.js
CendioOssman: Ideally, yes. We could also have some compromise with codes covering broad areas and an explaining text similar to how the `Error` exception is used.

Do you know how browsers handle translation in those cases? Or are they always in English?

samhed: Chrome does not have translations for warnings and errors that aren't intended for end-users. But I think we can be better, it's always a bit shitty to have mixed languages if you can avoid it.

core/display.js
```diff
@@ -88,7 +81,39 @@ try {
 }
 
 Display.prototype = {
-    // Public methods
+    // ===== PROPERTIES =====
+
+    _scale: 1.0,
+    get scale() { return this._scale; },
```
DirectXMan12: this is ES6 syntax.  Do all of our intended platforms support it?

DirectXMan12: (P.S. I'm fine with using ES6 syntax if our intended platforms support it, or even saying that we should transpile for older platforms, but it's something we need to decide on.  If we do, we can probably just switch to class syntax...)

CendioOssman:
I've now changed the documentation to resemble the style used by MDN. Hopefully that will make it easier for people to understand it. What do you think?

There's also a lot missing, so feel free to pitch in. :)

core/display.js
CendioOssman: MDN claims so. Haven't tested it yet.

CendioOssman:
> 👎 on the moving all the public properties from separate objects to RFB -- I think we should keep those on the separate objects, but treat what's returned from the getters like an interface (think typescript style w/o actually using it):

I'm not sure I understand what you are suggesting here. Can you give a code example?

Right now I really prefer what I've suggested, because:

 - It avoid exposing our internal structure, which would otherwise force us to keep that structure
 - It makes it more obvious what is stable API and what is not. I.e. everything on the `RFB` object that isn't prefixed by `_` is fair game.
 - Having to write `rfb.get_display().scale = 0.5;` instead of just `rfb.scale = 0.5;` is just meaningless extra characters for anyone using the API as the display object is just some magical internal thing anyway.

samhed:
I like the MDN style API docs! 

With regards to the API for the display (and friends) towards the app; I think it ties into the discussion of whether we should treat these other modules as separately valuable or not. To me it feels like they have grown (aside from maybe keyboard) quite specific for our use case. If you agree with this it would probably be more natural to keep the API for the UI limited to the RFB module.

DirectXMan12:
> I'm not sure I understand what you are suggesting here. Can you give a code example?

Without using something like TypeScript (which I'm not necessarily advocating, although it does make part of this issue much easier to deal with), it's much more just a documentation issue -- we document/explicitly mark that `scale` and a couple other things are considered stable, and everything else is not.  If someone were to provide TypeScript definition files, they'd do something like what I did above, which enforces that `let b = new RFB(); b.display.blitImage()` is a compile-time error, while `let b = new RFB(); b.display.scale = 0.5` is not.  For us, it'd just be documentation.

I'm *really* not a huge fan of piling a bunch of methods into RFB -- even if display and friends aren't usable apart from RFB, it makes sense to me to separate them.  It's easier to organize for us, and for the user, it's clearer that `rfb.display.scale` is scaling the display, as opposed to something else (the remote machine, etc).  If it's just `rfb.scale`, it's not clear what that means (IMO).

DirectXMan12:
(P.S. I agree with @samhed -- the MDN-style docs are really nice)

CendioOssman:
> I'm really not a huge fan of piling a bunch of methods into RFB -- even if display and friends aren't usable apart from RFB, it makes sense to me to separate them.

Code wise, I agree wholeheartedly. But for the interface I don't for the reasons above. And if we need documentation, then maintaining that is similar effort to maintaining the indirection in the code. And I think it is more likely that people will check our code when figuring things out. :)

> It's easier to organize for us, and for the user, it's clearer that rfb.display.scale is scaling the display, as opposed to something else (the remote machine, etc). If it's just rfb.scale, it's not clear what that means (IMO).

I can agree with that. But that could be handles as a naming issue. So I did some changes to the names in my latest iteration of this patch set. Have a look and see what you think. :)

samhed:
This PR is getting huge, would it be possible to split into a few smaller ones? Would make it a lot easier to review.

CendioOssman:
Not easily since they build upon each other. :/

I tried to keep each commit as small as possible at least.

app/ui.js
```diff
@@ -13,7 +13,7 @@
 
 import * as Log from '../core/util/logging.js';
 import _, { l10n } from '../core/util/localization.js';
-import { isTouchDevice, browserSupportsCursorURIs as cursorURIsSupported } from '../core/util/browsers.js';
+import { isTouchDevice } from '../core/util/browsers.js';
 import { setCapture, getPointerEvent } from '../core/util/events.js';
 import KeyTable from "../core/input/keysym.js";
 import keysyms from "../core/input/keysymdef.js";
```
DirectXMan12: make sure all of these imports are still needed.

app/ui.js
```diff
@@ -207,10 +207,9 @@ var UI = {
                               'onUpdateState': UI.updateState,
                               'onDisconnected': UI.disconnectFinished,
                               'onCredentials': UI.credentials,
-                              'onCapabilities': UI.updatePowerButton,
+                              'onCapabilities': function () { UI.updatePowerButton(); UI.initialResize(); },
```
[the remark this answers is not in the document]
CendioOssman: Probably. I was more focused on getting the API ready than cleanups.

app/ui.js
```diff
@@ -277,8 +277,8 @@ var UI = {
         document.getElementById("noVNC_keyboard_button")
             .addEventListener('click', UI.toggleVirtualKeyboard);
 
-        UI.touchKeyboard = new Keyboard({target: document.getElementById('noVNC_keyboardinput'),
-                                         onKeyEvent: UI.keyEvent});
+        UI.touchKeyboard = new Keyboard(document.getElementById('noVNC_keyboardinput'),
+                                        {onKeyEvent: UI.keyEvent});
```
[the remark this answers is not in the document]
CendioOssman: Sure. Feel free to add something to `master`. :)

core/input/keyboard.js
```diff
@@ -18,14 +18,14 @@ import KeyTable from "./keysym.js";
 // Keyboard event handler
 //
 
-export default function Keyboard(defaults) {
+export default function Keyboard(target, defaults) {
+    this._target = target || document;
```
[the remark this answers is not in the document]
CendioOssman: It makes sure there's a default. I'm just keeping the existing behaviour.

core/input/mouse.js
```diff
@@ -30,7 +31,6 @@ export default function Mouse(defaults) {
 
     // Configuration attributes
     set_defaults(this, defaults, {
-        'target': document,
```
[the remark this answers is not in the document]
CendioOssman: No idea. Didn't dig in to those details.

core/rfb.js
```diff
@@ -1478,6 +1493,8 @@ make_properties(RFB, [
     ['local_cursor', 'rw', 'bool'],         // Request locally rendered cursor
     ['shared', 'rw', 'bool'],               // Request shared mode
     ['view_only', 'rw', 'bool'],            // Disable client mouse/keyboard
+    ['scale', 'rw', 'float'],               // Display area scale factor
```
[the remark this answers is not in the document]
CendioOssman: This one got fixed later. Anything you feel was overlooked?

core/rfb.js
```diff
@@ -1282,7 +1277,8 @@ RFB.prototype = {
             case 1:  // XVP_INIT
                 this._rfb_xvp_ver = xvp_ver;
                 Log.Info("XVP extensions enabled (version " + this._rfb_xvp_ver + ")");
-                this._onXvpInit(this._rfb_xvp_ver);
+                this._capabilities.power = true;
+                this._onCapabilities(this, this._capabilities)
```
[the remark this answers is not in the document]
CendioOssman: I had a look at other APIs in MDN, and this seems to be a more common model.

docs/API-internal.md
```diff
@@ -0,0 +1,141 @@
+# 1. Internal Modules
+
+The noVNC client is a composed of several internal modules that handle
```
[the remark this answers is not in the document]
CendioOssman: Inherited from the original, but I've fixed it now. :)

core/rfb.js
```diff
@@ -324,63 +339,55 @@ RFB.prototype = {
...
-            this._sock.flush();
-            return true;
-        } else {
-            return false;
+        if (!this._supportsSetDesktopSize) {
+            return;
         }
+
+        RFB.messages.setDesktopSize(this._sock, width, height,
+                                    this._screen_id, this._screen_flags);
```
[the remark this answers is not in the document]
CendioOssman: We did a cleanup in 89d2837fa8e3f0498bbac7b7194f453d88bcbc1f but apparently we missed this one.

CendioOssman:
> Method names look better, but I'm still 👎 versus just making a couple of the methods and properties on display part of the public API.
> 

Fair enough. Which variant does @samhed and @kanaka prefer?

> Additionally, IMO, anything called from app/ui.js and app/webutil.js need to be listed as public APIs. The "app" should exist as both a standalone app and a good example of how to use all the features of core, so if someone copied over app/* as their own, they should be able to use it with the official releases of noVNC without having to worry about it using internals.

I agree that the goal should be that all features only rely on public API, but I also think that it would be very helpful if we can have `vnc.html` as a test bed for new features before we finalise the API. So I'm afraid I don't agree with the goal of being able to drop it in to any future version of `rfb.js`.

Is there something specific you noticed?

core/rfb.js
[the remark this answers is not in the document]
CendioOssman: `Base64` is the only one not used. But it's not used on `master` either, so feel free to remove it there.

`KeyTable` is only used by `sendCtrlAltDel()`, which I considered removing since it is just a wrapper around `sendKey()`. Does anyone of you want to keep it?

app/ui.js
CendioOssman: `Base64` is the only one not used. But it's not used on `master` either, so feel free to remove it there.

`KeyTable` is only used by `sendCtrlAltDel()`, which I considered removing since it is just a wrapper around `sendKey()`. Does anyone of you want to keep it?

docs/API.md
```diff
@@ -1,178 +1,32 @@
...
     if (rfb.get_encrypt()) {
         alert("Encryption is set");
     }
     rfb.set_encrypt(false);
 
-Some attributes are read-only and cannot be changed. For example, the
-Display 'render_mode' option will throw an exception if an attempt is
-made to set it. The attribute mode is one of the following:
+Some attributes are read-only and cannot be changed. An exception will
+be thrown if an attempt is made to set one of these attributs. The
```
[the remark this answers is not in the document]
CendioOssman: This was removed in the conversion to MDN style docs.

app/ui.js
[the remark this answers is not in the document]
samhed: fair enough

core/input/keyboard.js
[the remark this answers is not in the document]
samhed: Yeah, but the default was null

app/ui.js
samhed: I'm fine with removing sendCtrlAltDel

samhed:
@CendioOssman With the new names of the display functions I'm happy.

@DirectXMan12 

> Additionally, IMO, anything called from app/ui.js and app/webutil.js need to be listed as public APIs. The "app" should exist as both a standalone app and a good example of how to use all the features of core, so if someone copied over app/* as their own, they should be able to use it with the official releases of noVNC without having to worry about it using internals.

I'm not sure I follow here, you mean that app/ui.js would use internals indirectly by using for example RFB.viewportScale that in turn is using Display.scale directly?

core/rfb.js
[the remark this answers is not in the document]
samhed: No, good!

kanaka:
> Fair enough. Which variant does @samhed and @kanaka prefer?

I lean towards the variant suggested by @DirectXMan12, but not strongly. My intuition is that it will be easier to maintain going forward but I don't have any concrete evidence. And since I'm not personally doing much maintenance for the foreseeable future, take that as you will.

core/input/keyboard.js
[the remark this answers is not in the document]
CendioOssman: Ah. Will fix.

core/rfb.js
```diff
@@ -230,47 +225,76 @@ export default function RFB(defaults) {
...
+    get capabilities() { return this._capabilities; },
+
+    get touchButton() { return this._mouse.touchButton; },
+    set touchButton(button) { this._mouse.touchButton = button; },
+
+    get viewportScale() { return this._display.scale; },
+    set viewportScale(scale) { this._display.scale = scale; },
+
+    get clipViewport() { return this._display.clipViewport; },
+    set clipViewport(viewport) { this._display.touchButton = clipViewport; },
```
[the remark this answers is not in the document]
CendioOssman: Oops. Will fix. :)

CendioOssman:
I've updated the branch with the found problems now. I've also added a commit changing events to something more like `EventTarget`. Unfortunately it was not possible to use the browser's code for this, so something similar in JavaScript was added.

core/rfb.js
```diff
@@ -512,7 +502,9 @@ RFB.prototype = {
         // State change actions
 
         this._rfb_connection_state = state;
-        this.onupdatestate(this, state, oldstate);
+        var event = document.createEvent("CustomEvent");
+        event.initCustomEvent("updatestate", false, false, { state: state });
+        this.dispatchEvent(event);
```
samhed: Can't we make a helper function for this?

CendioOssman: Possibly. But it was not that much code, so I figured it might be better to keep it obvious what is happening.

samhed: fair enough

app/ui.js
DirectXMan12: `Display` doesn't seem to be used either.

core/rfb.js
```diff
@@ -512,7 +502,9 @@ RFB.prototype = {
         // State change actions
 
         this._rfb_connection_state = state;
-        this.onupdatestate(this, state, oldstate);
+        var event = document.createEvent("CustomEvent");
+        event.initCustomEvent("updatestate", false, false, { state: state });
```
DirectXMan12: we should *really* be using the `CustomEvent` constructor when possible -- the `initCustomEvent` has been deprecated and removed from the standards (https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/initCustomEvent).  With our target browser support, I don't think we can use it all the time, but we can use the polyfill listed here: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent

core/util/eventtarget.js
```diff
@@ -0,0 +1,45 @@
+/*
+ * noVNC: HTML5 VNC client
+ * Copyright 2017 Pierre Ossman for Cendio AB
+ * Licensed under MPL 2.0 (see LICENSE.txt)
+ *
+ * See README.md for usage and integration instructions.
+ */
+
+export var EventTargetMixin = {
```
DirectXMan12: just `export default` this

core/util/eventtarget.js
```diff
@@ -0,0 +1,45 @@
...
+    _listeners: null,
+
+   addEventListener: function(type, callback) {
+      if (!this._listeners) {
+         this._listeners = {};
```
DirectXMan12: should just use a `Map` to avoid accidental collisions with built-in properties, etc.

core/util/eventtarget.js
```diff
@@ -0,0 +1,45 @@
...
+      this._listeners[type].push(callback);
+   },
+
+   removeEventListener: function(type, callback) {
+      if (!this._listeners || !(type in this._listeners)) {
+         return;
+      }
+      var stack = this._listeners[type];
+      for (var i = 0, l = stack.length; i < l; i++) {
+         if (stack[i] === callback) {
```
DirectXMan12: make this a `Set` to allow easy removal.

core/rfb.js
CendioOssman: I was using the constructor to start with, but that went south once I tested on IE. :sob:

I can try the polyfill and see how that goes.

core/util/eventtarget.js
CendioOssman: How would a `Map` improve things in that regard?

core/util/eventtarget.js
CendioOssman: Not sure why I didn't in the first place...

core/util/eventtarget.js
CendioOssman: Alright, I'll give it a try.

app/ui.js
CendioOssman: Ehm? It's instantiated in the RFB constructor. :)

core/util/eventtarget.js
CendioOssman: Ah, I see what you mean now. Will fix.

CendioOssman:
Comments fixed, thanks.

I think we're at a point where we can look at merging this. I've been running the tests on most browsers, but I should probably do some manual tests as well.

We seem to be divided on the display/mouse/keyboard API issue, so unless there's a massive objection I'll refer to "he who writes the code gets to decide" to break the tie. :)

samhed:
LGTM!

CendioOssman:
Found a bunch of bugs (which I've since fixed and pushed here), but now everything seems to work fine where I can test.

I'll give you guys a few days to have a final look and then I'll merge it. :)

DirectXMan12:
Whoops, looks like one of my replies didn't submitted.  I've reproduced it below.

I'll try and do a final review again soon.   Can you push this to a branch on the repo before merge?  I want to play around with submitting a PR with a couple of things that I've been thinking about, but want to keep the discussion in this repo for posterity.

> I'm not sure I follow here, you mean that app/ui.js would use internals indirectly by using for example RFB.viewportScale that in turn is using Display.scale directly?

I'm suggesting that anything we call from `app/ui.js` should be public API.  It should be theoretically possible for someone to completely reproduce `vnc.html` without using any of the internal APIs.

> I agree that the goal should be that all features only rely on public API, but I also think that it would be very helpful if we can have `vnc.html` as a test bed for new features before we finalise the API. So I'm afraid I don't agree with the goal of being able to drop it in to any future version of `rfb.js`.

That's not precisely what I was suggesting ;-).  I merely meant what I stated above -- if someone wanted to reproduce vnc.html without using internal APIs, it should be possible.  I don't think that conflicts with the goal of using `vnc.html` as a testbed.

CendioOssman:
> I'll try and do a final review again soon. Can you push this to a branch on the repo before merge? I want to play around with submitting a PR with a couple of things that I've been thinking about, but want to keep the discussion in this repo for posterity.

I'm not sure what you're trying to do, but I've pushed an `api` branch here now.

> That's not precisely what I was suggesting ;-). I merely meant what I stated above -- if someone wanted to reproduce vnc.html without using internal APIs, it should be possible. I don't think that conflicts with the goal of using vnc.html as a testbed.

I still don't think I understand you in that case because test bed and public api are fundamentally in conflict in my world. :)

CendioOssman:
@DirectXMan12, do you need some more time or can I merge this?

DirectXMan12:
hey, appologies, got caught up in something at work, then got a bit sick :-/.  Anyway, I thought most of it was good to go :-).

Issue #881, opened 2017-08-16
Better alignment of console window


rh-jelabarre:
One problem I've seen with the console window on noVNC is since the console starts immediately at the top of the page, it looks as though the top of the window is being cut off.  It took a bit of experimenting on my part to determine it isn't, but even with that it continues to appear as it is being cut off, or that you should be able to scroll upward.
I would suggest the window should have some small spacer above the console window to make it clear the console is being shown in it's entirety.

DirectXMan12:
cc @CendioOssman who does most of the visual work on noVNC these days.

Can you post a screenshot with what you're suggesting?

rh-jelabarre:
![novnc_console_current](https://user-images.githubusercontent.com/29484746/29432402-77222fca-8369-11e7-9e0c-6818da9d673b.jpeg)
![novnc_console_modification](https://user-images.githubusercontent.com/29484746/29432403-77254ce6-8369-11e7-8301-dce5f0367ba1.jpeg)

I have attached a couple jpeg images, one which shows a current view of an instance console in Openstack, and a second one with the suggested modification of dropping the console window jown just enough to separate it from the "header".  I believe Openstack makes use of the minimal (novnc_auto ?) form of NoVNC.

DirectXMan12:
ah, I see what you mean.  That makes a bit less sense when fullscreen, but we can probably work out a good design that works in both cases.

CendioOssman:
Ah, so this is `vnc_lite`. The standard frontend already centers the display. Perhaps the same should be done for lite?

And perhaps we should add a label so we can separate lite stuff from the rest?

DirectXMan12:
> Ah, so this is vnc_lite. The standard frontend already centers the display. Perhaps the same should be done for lite?

Yeah, seems reasonable to me.

> Ah, so this is vnc_lite. The standard frontend already centers the display. Perhaps the same should be done for lite?

Done.  The labels are `ui/full` and `ui/lite`.

samhed:
Fixed in #947

PR #947, opened 2017-11-01
Properly center things in vnc_lite.html

samhed:
Problems fixed:
* The canvas wasn't centered vertically (#881)
* The status text was previously only centered in the remaining space from the buttons
* There were conflicting styles in html and css


samhed:
I changed one commit and added a dummy element to the left now instead of using a transform to center the status text.

PR #950, opened 2017-11-12
Abstract information from RFB to the UI 

samhed:
We had a problem with strings being passed from RFB to UI using the notification and disconnect events. These were strings that were intended to be shown directly to the end user. Since these strings were defined in RFB the UI did not have control over translations.

This PR removes the `notification` event, and the `updatestate` event. It adds a `connect` event, a `securityfailure` event, and modifies the `disconnect` event:
 * The `notification` event was unnecessary and was easily replaced by logging
 * The `updatestate` event was being sent at alongside the `disconnect` event, causing unnecessary complexity for the applications. Applications do normally not have any use for the more detailed internal state information anyway.
 * There is now a `connect` event that works nicely with the `disconnect` event.
 * The `disconnect` event no longer includes a reason. Instead it includes a boolean that marks if the disconnection was clean (normal) or not clean (exited with errors).
 * A `securityfailure` event is now available for applications that want to display the security result failure string that can be sent directly from the server. One example of this is authentication failures. 

I still have to update the API documentation, but feel free to review.


app/ui.js
```diff
@@ -1063,16 +1055,46 @@ var UI = {
...
+        UI.connected = false;
+
         if (typeof e.detail.reason !== 'undefined') {
             UI.showStatus(e.detail.reason, 'error');
+            UI.updateVisualState('disconnected');
+
         } else if (UI.getSetting('reconnect', false) === true && !UI.inhibit_reconnect) {
-            document.getElementById("noVNC_transition_text").textContent = _("Reconnecting...");
-            document.documentElement.classList.add("noVNC_reconnecting");
+
```
CendioOssman: Extra blank lines in this general area

vnc_lite.html
```diff
@@ -200,8 +180,7 @@
         };
 
         function updatePowerButtons() {
-            var powerbuttons;
-            powerbuttons = document.getElementById('noVNC_power_buttons');
+            let powerbuttons = document.getElementById('noVNC_power_buttons');
```
CendioOssman: Bonus change? ;)

core/rfb.js
```diff
@@ -11,7 +11,6 @@
  */
 
 import * as Log from './util/logging.js';
-import _ from './util/localization.js';
```
CendioOssman: We should probably move `localization.js` to the app side of things now.

app/ui.js
```diff
@@ -479,21 +479,37 @@ var UI = {
...
+        if (statusElem.classList.contains("noVNC_open")) {
+            if (statusElem.classList.contains("noVNC_status_error")) {
+                visible_status_type = 'error';
+            } else if (statusElem.classList.contains("noVNC_status_warn")) {
+                visible_status_type = 'warn';
+            } else {
+                visible_status_type = 'normal';
+            }
+        }
+        if (visible_status_type === 'error') { return; }
```
CendioOssman: How about if `warn` is visible and a `normal` wants to be shown?

samhed: Will fix

app/ui.js
samhed: I'll remove them

core/rfb.js
samhed: Yeah, one small thing to note is that tests/test.util.js is generally written for core/util and will now test app/localization.js as well, but not a big problem

vnc_lite.html
samhed: I'll remove this change

samhed:
I discovered an issue with the focus handling that was introduced with the 118d929 commit. It seems like running .focus() on an element that isn't rendered have no effect. I'll just move the focus call to after showing the canvas (like it worked before).

samhed:
All fixed now. API documentation still remaining though

core/rfb.js
CendioOssman: Or just split that test?

samhed: yeah, sounds good

samhed:
API documentation updated as well now, I'm satisfied.

docs/API.md
```diff
@@ -177,24 +177,11 @@ connection to a specified VNC server.
...
-| `"connected"`     | connected normally
-| `"disconnecting"` | starting to disconnect
-| `"disconnected"`  | disconnected
-
-Note that a `RFB` objects can not transition from the disconnected
-state in any way, a new instance of the object has to be created for
-new connections.
+The `connect` event is fired after all the handshaking with the server
+is completed and the connection is fully established. After this event
+the `RFB` object is ready to recieve grahpics updates and to send input.
```
CendioOssman: Typo

core/rfb.js
```diff
@@ -799,6 +786,59 @@ RFB.prototype = {
...
+        if (strlen > 0) {
+            if (this._sock.rQwait("reason", strlen, 8)) { return false; }
+            reason = this._sock.rQshiftStr(strlen);
+        }
+
+        if (reason !== "") {
+
+            let event = new CustomEvent(
+                "securityfailure",
+                { detail: { result: security_result, reason: reason } });
```
CendioOssman: The entire message is called `SecurityResult`, but the code field is called `status`, so perhaps we should use the same terminology here?

samhed:
Any thoughts @DirectXMan12?

docs/API.md
```diff
@@ -198,6 +202,15 @@ credentials than were specified to [`RFB()`](#rfb-1). The `detail`
...
 `Array` of `DOMString` listing the credentials that are required.
 
+#### securityfailure
+
+The `securityfailure` event is fired when the handshaking process with
+the server fails during the security negotiation step. The `detail`
+property is an `Object` optionally containing the property `reason`
+which is a `DOMString` specifiying the reason for the failure. The
+server can choose to not send a reason and in these cases the property
+`reason` will be omitted.
```
CendioOssman: Can you try to make it clearer that this is a server provided message, rather than something internal? Perhaps also mention that the language is not fully known (although most likely english)?

CendioOssman: You haven't documented `reason`/`status`.

core/rfb.js
samhed: I missed that, I'll change it

docs/API.md
samhed: Yes, will do

docs/API.md
samhed: good eye

docs/API.md
samhed: What do you mean?

samhed: Ah, you mean `result`. Yeah, I missed to document that since we don't currently use it in the UI

samhed: To clarify; `result` which will be renamed to `status` wasn't documented

samhed:
Fixed all your comments now @CendioOssman, I also found that I had forgotten to add tests for the new `securityfailure` event, so I added that.

docs/API.md
```diff
@@ -198,6 +202,25 @@ credentials than were specified to [`RFB()`](#rfb-1). The `detail`
...
+The `securityfailure` event is fired when the handshaking process with
+the server fails during the security negotiation step. The `detail`
+property is an `Object` containing the following properties:
+
+| Property | Type        | Description
+| -------- | ----------- | -----------
+| `status` | `long`      | The failure status code
+| `reason` | `DOMString` | The **optional** reason for the failure
+
+The property `status` corresponds to the *SecurityResult* status code in
```
CendioOssman: Users of the API might not be that familiar with the protocol. A link to the relevant section of rfbproto would be nice here.

samhed: good idea!

samhed:
Added a link to SecurityResult in rfbproto for the `securityfailure` event status as per @CendioOssman comment

samhed:
I'll give people until tomorrow to review this, then I will go ahead and merge.

DirectXMan12:
@samhed @CendioOssman hey, sorry, was out sick.  FYI, next week is a holiday in the US, so I'll probably be mostly unresponsive ;-).

Issue #951, opened 2017-11-13
vnc_lite.html broken


dtcooper:
Hi there, there's a reference error in `vnc_lite.html`, in particular lines (205-213)[https://github.com/novnc/noVNC/blob/b8dfb983/vnc_lite.html#L205-L213] in function `updatePowerButtons()`: `powerbuttons` needs to be renamed to `xvpbuttons`.

dtcooper:
Fix available on my PR. Thanks!

PR #952, opened 2017-11-13
Fix vnc_lite.html (powerbuttons -> xvpbuttons)

dtcooper:
This fixes issue #951. Thanks!


CendioOssman:
Thank you, but the rename should actually be the other way around. Fixed in 56f21bf72c4331a2db5fde4428d82e295355db80.

PR #953, opened 2017-11-14
Fixed --cert Argument Typo in launch.sh

Bubbl3H3d:
Fixed a transposition error on line 145 affecting the --cert argument.


CendioOssman:
That was no error. The existing code makes sure the cert argument is only added when needed. Please open an issue with the problem you are seeing instead.

PR #960, opened 2017-11-24
Move resize handling in to RFB object

CendioOssman:
Makes the API simpler and makes it easier for other frontends to get this functionality. Just look at the diff for `ui.js`. And the second commit demonstrates how easy it is for other frontends to start using these features. :)

This is the final portion I want in for #924.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -66,7 +66,7 @@ export default function RFB(target, url, options) {
 
     this._fb_name = "";
 
-    this._capabilities = { power: false, resize: false };
+    this._capabilities = { power: false };
 
     this._supportsFence = false;
 
@@ -88,6 +88,7 @@ export default function RFB(target, url, options) {
 
     // Timers
     this._disconnTimer = null;      // disconnection timer
+    this._resizeTimeout = null;     // resize rate limiting
 
     // Decoder states and stats
     this._encHandlers = {};
@@ -140,15 +141,29 @@ export default function RFB(target, url, options) {
     // Bound event handlers
     this._eventHandlers = {
         focusCanvas: this._focusCanvas.bind(this),
+        windowResize: this._windowResize.bind(this),
     };
 
     // main setup
     Log.Debug(">> RFB.constructor");
 
-    // Target canvas must be able to have focus
-    if (!this._target.hasAttribute('tabindex')) {
-        this._target.tabIndex = -1;
-    }
+    // Create DOM elements
+    this._screen = document.createElement('div');
+    this._screen.style.display = 'flex';
+    this._screen.style.width = '100%';
+    this._screen.style.height = '100%';
+    this._screen.style.overflow = 'auto';
+    this._screen.style.backgroundColor = 'rgb(40, 40, 40)';
+    this._canvas = document.createElement('canvas');
+    this._canvas.style.margin = 'auto';
+    // Some browsers add an outline on focus
+    this._canvas.style.outline = 'none';
+    // IE miscalculates width without this :(
+    this._canvas.style.flexShrink = '0';
+    this._canvas.width = 0;
+    this._canvas.height = 0;
+    this._canvas.tabIndex = -1;
+    this._screen.appendChild(this._canvas);
 
     // populate encHandlers with bound versions
     this._encHandlers[encodings.encodingRaw] = RFB.encodingHandlers.RAW.bind(this);
@@ -166,7 +181,7 @@ export default function RFB(target, url, options) {
     // NB: nothing that needs explicit teardown should be done
     // before this point, since this can throw an exception
     try {
-        this._display = new Display(this._target);
+        this._display = new Display(this._canvas);
     } catch (exc) {
         Log.Error("Display exception: " + exc);
         throw exc;
@@ -174,10 +189,10 @@ export default function RFB(target, url, options) {
     this._display.onflush = this._onFlush.bind(this);
     this._display.clear();
 
-    this._keyboard = new Keyboard(this._target);
+    this._keyboard = new Keyboard(this._canvas);
     this._keyboard.onkeyevent = this._handleKeyEvent.bind(this);
 
-    this._mouse = new Mouse(this._target);
+    this._mouse = new Mouse(this._canvas);
     this._mouse.onmousebutton = this._handleMouseButton.bind(this);
     this._mouse.onmousemove = this._handleMouseMove.bind(this);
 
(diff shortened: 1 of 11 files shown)


core/rfb.js
```diff
@@ -447,7 +475,96 @@ RFB.prototype = {
...
+        } else {
+            let size = this._screenSize();
+            this._display.autoscale(size.w, size.h);
+        }
+        this._fixScrollbars();
+    },
+
+    // Requests a change of remote desktop size. This message is an extension
+    // and may only be sent if we have received an ExtendedDesktopSize message
+    _applyResize: function () {
```
samhed: rename to something like `requestRemoteResize`?

docs/API.md
```diff
@@ -32,24 +32,26 @@ protocol stream.
...
   - Is a `boolean` indicating if mouse events should control the
-    relative position of a clipped canvas. Only relevant if
+    relative position of a clipped session. Only relevant if
     `clipViewport` is enabled. Disabled by default.
 
-`isClipped` *Read only*
-  - Is a `boolean` indicating if the framebuffer is larger than the
-    current canvas, i.e. it is being clipped.
+`scaleViewport`
+  - Is a `boolean` indicating if the session should be scaled locally
```
samhed: I'm not sure if 'session' is a good term in general, do you think 'remote' works? This would mean changes to a number of places.

core/rfb.js
```diff
@@ -140,16 +141,39 @@ export default function RFB(target, url, options) {
...
+    if (!found) {
+        let link = document.createElement('link');
+        link.rel = "stylesheet";
+        link.href = styleurl;
+        document.head.appendChild(link);
     }
 
+    // Create DOM elements
+    this._screen = document.createElement('div');
+    this._screen.className = 'noVNC_screen';
```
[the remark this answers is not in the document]
CendioOssman: Sure. Suggestions? `noVNC_container`?

core/rfb.js
```diff
@@ -391,20 +419,31 @@ RFB.prototype = {
...
     _disconnect: function () {
         Log.Debug(">> RFB.disconnect");
-        this._target.removeEventListener("mousedown", this._eventHandlers.focusCanvas);
-        this._target.removeEventListener("touchstart", this._eventHandlers.focusCanvas);
-        this._cleanup();
+        this._canvas.removeEventListener("mousedown", this._eventHandlers.focusCanvas);
+        this._canvas.removeEventListener("touchstart", this._eventHandlers.focusCanvas);
+        window.removeEventListener('resize', this._eventHandlers._windowResize);
+        this._keyboard.ungrab();
+        this._mouse.ungrab();
```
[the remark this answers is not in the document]
CendioOssman: > Now ungrab gets called even in view only, I assume this isn't a problem?

Shouldn't be, no. Keeps things simpler and more fool proof that we properly unlink ourselves from the element.

> And the cursor doesn't get reset on disconnect now?

No need. The canvas element gets removed now.

core/rfb.js
CendioOssman: Sure

docs/API.md
CendioOssman: Just "remote"? Or do you mean "remote session"?

core/rfb.js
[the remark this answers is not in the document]
samhed: The div with id `noVNC_screen` is inside a div called `noVNC_container` in vnc.html. So that name is also taken. Maybe `noVNC_canvas_container`? But I don't really have any good suggestions

docs/API.md
samhed: I was thinking only "remote", but "remote session" could work too. What phrase are you using in TigerVNC for example?

CendioOssman: TigerVNC uses "remote session".

samhed: Ok, good enough for me

CendioOssman:
> So, I don't mind the general concept here (making it easier to consume resize functionality), but I'm definitely against this in it's current form. noVNC shouldn't modify the surrounding page. At the very least, it's unexpected to have an embeded component start trying to add stuff to head. At worst, it could create headaches when trying to embed noVNC in non-standard ways (especially with the brittle link checking, which expect noVNC files to be at the root).
> 

Right, the `<link>` tag was just a temporary thing to get the ball rolling. I've now changed that to a `<style>` block with inline CSS. However it still places it in head as we only want a single copy in the case of multiple RFB instances. I'm fixing up some tests and then I'll update the PR.

Do you see any other problems with this variant? I think this is cleaner than putting style attributes directly on the elements, which would be the alternative.

> Why not make the resizing a helper library? Opt-in for people who want it, and people who don't can easily write their own stuff.

Because it is not easy to write your own stuff, and I don't feel comfortable freezing that API. The suggestion here is much more high level and much easier to commit to keeping in future versions.

Do you have an idea where this approach could cause problems?

core/rfb.js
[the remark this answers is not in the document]
CendioOssman: I found a simple solution; `#noVNC_screen` is no longer needed so I just removed it. :)

CendioOssman:
Alright, CSS fixed, tests fixed, and suggested names changed. Should be good to go from my point of view.

core/rfb.js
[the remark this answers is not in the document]
samhed: great!

DirectXMan12:
> Do you see any other problems with this variant? I think this is cleaner than putting style attributes directly on the elements, which would be the alternative.

As much as it pains me to say it, I'd prefer the style directly on the element, at least until we get ShadowDOM support (unless you feel like requiring the shadowdom and shadycss polyfills from https://www.webcomponents.org/polyfills).  My reasoning is that it's generally bad form to introduce something to the global namespace (all non-shadow-dom CSS is effectively in the global namespace) in a library like this.  We don't want this accidentally applying to other things, overlapping with existing names, etc.

CendioOssman:
I'm not terribly worried given the `noVNC_` prefix, but it's not a big issue so I've pushed a new version with the styles inline. Have a look.

samhed:
Any other differences? In some ways the method of opening a new PR for each revision of the patch set would make things a lot easier to compare in cases like this..

CendioOssman:
> Any other differences?

No

> In some ways the method of opening a new PR for each revision of the patch set would make things a lot easier to compare in cases like this..

That just makes it much more difficult to find things in the future, and you have to go digging for comments and discussion. :-1:

github has some "show differences since last review" function. Does that handle rebase?

samhed:
> github has some "show differences since last review" function. Does that handle rebase?

That sounded promising, but no, it seems it doesn't handle rebase. You have made changes since my last review, and the button is greyed out for me and says "No new changes".

DirectXMan12:
> In some ways the method of opening a new PR for each revision of the patch set would make things a lot easier to compare in cases like this..

Ooof, please don't do that.  It's really hard to track.  We could use fixup commits, and then ask to squash at the end with `--autosquash` or something, if you want.

DirectXMan12:
With the inline style, this is fine.

samhed:
> Ooof, please don't do that. It's really hard to track. We could use fixup commits, and then ask to squash at the end with --autosquash or something, if you want.

Yeah that sounds fine. It would be very nice to be able to use the github "show differences since last review" feature after rebase as well.. Didn't you guys have the same issues as me when reviewing a PR thats being modified?

@CendioOssman all done for merging then?

PR #963, opened 2017-11-30
Initalize cached 16x16 tile.

josiah2009:
Without this, noVNC will display a blank screen when dealing with a server that sends 16 by 16 tiles.

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -69,6 +69,7 @@ export default function Display(target) {
         throw new Error("Canvas does not support createImageData");
     }
 
+    this._tile16x16 = this._drawCtx.createImageData(16, 16);
     Log.Debug("<< Display.constructor");
 };
 


CendioOssman:
Hmm.. It should be crashing rather than give a blank screen...

Anyway, it seems like this got removed by accident in 1e13775bd5e37173b6c8a814811ea7f16e56f892. Nice catch. :)

DirectXMan12:
we should probably have a test for this too.

CendioOssman:
Indeed. But I'm thinking why are we even having a cached tile? It doesn't seem to be a common code path given that this bug has been around since 2014 without anyone hitting it.

Maybe nuke it instead?

DirectXMan12:
I don't know about nuking it.  I recall it being a significant perf improvement to something qemu was sending, but I'm not certain.  I'd prob just test it normally.

Issue #961, opened 2017-11-27
better documentation on npmjs.com


CendioOssman:
This comment from #910 was overlooked:

> Can we modify the info shown at npmjs.com? It currently shows stuff relevant for the app, not the library.

DirectXMan12:
From a quick googling, the answer may actually be "no".  I can't seem to find a way to override which readme is used for npm publishing.  We could *maybe* do some trickery on before-publish to rename files (e.g. have a `README-LIBRARY.md` and rename that to `README.md` just before publish)...

CendioOssman:
Annoying. But I guess we need a section in the normal README then?

DirectXMan12:
Yeah, I'll be out at a conference next week, but I can try to write one when I get back, unless you'd prefer to .

CendioOssman:
I've merged #971 now. I'm pretty content with it for now. Anything else we need before closing this issue?

samhed:
LGTM

PR #971, opened 2017-12-08
Documentation improvements

CendioOssman:
Suggestions on how to improve the documentation in preparation for a release.


docs/EMBEDDING.md
```diff
@@ -0,0 +1,49 @@
...
+
+The most basic deployment consists of simply serving these files from a web
+server and setting up a WebSocket proxy to the VNC server.
+
+## Parameters
+
+The noVNC application can be controlled by including certain settings in the
+query string. Currently the following options are available:
+
+FIXME
```
samhed: Is this work in progress or are you fine leaving this FIXME until after the release?

README.md
```diff
@@ -67,63 +72,45 @@ See more screenshots
...
 ### Server Requirements
 
-Unless you are using a VNC server with support for WebSockets connections (such
-as [x11vnc/libvncserver](http://libvncserver.sourceforge.net/),
-[QEMU](http://www.qemu.org/), or
-[MobileVNC](http://www.smartlab.at/mobilevnc/)), you need to use a
-WebSockets to TCP socket proxy. There is a python proxy included
-('websockify').
+noVNC follows the standard VNC protocol, but unlike other VNC clients it does
+require WebSockets support. Many server include support (e.g.
```
kanaka: server -> servers

CendioOssman:
> Maybe we can tune the 'library' aspect of things down a bit? Perhaps phrase it like its mainly a VNC client but with a core library that can be used separately. That's how I view noVNC anyway. No strong feelings though, I'm fine either way!

We're putting a lot of effort in to making this a general library, so I think I'd be a waste not to advertise that. :)

docs/EMBEDDING.md
CendioOssman: Not sure. :)

I mainly wanted to get the PR up before the weekend so that @DirectXMan12 could see I started looking at #961 and friends.

But it would be nice to have that section filled in as well.

README.md
CendioOssman: Thanks. Will fix.

PR #973, opened 2017-12-11
improve use_require.js

### LINKED ISSUE #970: creating a browserify bundle leaves .js files around
The easiest way to distribute the noVNC application is by doing:

```
$ utils/use_require.js  --with-app --as commonjs
```

However that leaves you with a `build/` directory with lots of extra crap in it, namely all the `.js` files that were bundled up in to `app.js`.

We should clean those files so people can easily just take the entire `build/` directory.

CendioOssman:
Some improvements to our conversion script to make it more useful when converting the application:

 - Fixes some races by using promises
 - Cleans up unnecessary files when bundling (fixes #970)
 - Lets converted modules be a fallback rather than forced on all browsers

### DIFF
--- a/utils/use_require.js
+++ b/utils/use_require.js
@@ -12,6 +12,7 @@ program
     .option('--as [format]', `output files using various import formats instead of ES6 import and export.  Supports ${Array.from(SUPPORTED_FORMATS)}.`)
     .option('-m, --with-source-maps [type]', 'output source maps when not generating a bundled app (type may be empty for external source maps, inline for inline source maps, or both) ')
     .option('--with-app', 'process app files as well as core files')
+    .option('--only-legacy', 'only output legacy files (no ES6 modules) for the app')
     .option('--clean', 'clear the lib folder before building')
     .parse(process.argv);
 
@@ -86,7 +87,7 @@ var walkDir = function (base_path, cb, filter) {
     });
 };
 
-var transform_html = function (new_script) {
+var transform_html = function (legacy_scripts, only_legacy) {
     // write out the modified vnc.html file that works with the bundle
     var src_html_path = path.resolve(__dirname, '..', 'vnc.html');
     var out_html_path = path.resolve(paths.out_dir_base, 'vnc.html');
@@ -99,6 +100,37 @@ var transform_html = function (new_script) {
         var start_ind = contents.indexOf(start_marker) + start_marker.length;
         var end_ind = contents.indexOf(end_marker, start_ind);
 
+        new_script = '';
+
+        if (only_legacy) {
+            // Only legacy version, so include things directly
+            for (let i = 0;i < legacy_scripts.length;i++) {
+                new_script += `    <script src="${legacy_scripts[i]}"></script>\n`;
+            }
+        } else {
+            // Otherwise detect if it's a modern browser and select
+            // variant accordingly
+            new_script += `\
+    <script type="module">\n\
+        window._noVNC_has_module_support = true;\n\
+    </script>\n\
+    <script>\n\
+        window.addEventListener("load", function() {\n\
+            if (window._noVNC_has_module_support) return;\n\
+            let legacy_scripts = ${JSON.stringify(legacy_scripts)};\n\
+            for (let i = 0;i < legacy_scripts.length;i++) {\n\
+                let script = document.createElement("script");\n\
+                script.src = legacy_scripts[i];\n\
+                script.async = false;\n\
+                document.head.appendChild(script);\n\
+            }\n\
+        });\n\
+    </script>\n`;
+
+            // Original, ES6 modules
+            new_script += '    <script type="module" crossorigin="anonymous" src="app/ui.js"></script>\n';
+        }
+
         contents = contents.slice(0, start_ind) + `${new_script}\n` + contents.slice(end_ind);
 
         return contents;
@@ -109,7 +141,7 @@ var transform_html = function (new_script) {
     });
 }
 
-var make_lib_files = function (import_format, source_maps, with_app_dir) {
+var make_lib_files = function (import_format, source_maps, with_app_dir, only_legacy) {
     if (!import_format) {
         throw new Error("you must specify an import format to generate compiled noVNC libraries");
     } else if (!SUPPORTED_FORMATS.has(import_format)) {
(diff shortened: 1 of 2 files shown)


utils/use_require.js
```diff
@@ -38,33 +38,58 @@ const no_transform_files = new Set([
 
 no_copy_files.forEach((file) => no_transform_files.add(file));
 
+function readFile(path) {
+    return new Promise((resolve, reject) => {
```
DirectXMan12: https://nodejs.org/dist/latest-v9.x/docs/api/util.html#util_util_promisify_original

DirectXMan12: basically everywhere with manual promises can be cleaned up with this

utils/use_require.js
```diff
@@ -75,9 +100,7 @@ var transform_html = function (new_script) {
         contents = contents.slice(0, start_ind) + `${new_script}\n` + contents.slice(end_ind);
 
         console.log(`Writing ${out_html_path}`);
-        fs.writeFile(out_html_path, contents, function (err) {
-            if (err) { throw err; }
-        });
+        return writeFile(out_html_path, contents);
```
DirectXMan12: this flows a bit nicer as 

```javascript
return readFile(src_html_path).then((contents_raw) => {
    ...
    return contents
}).then((contents) => writeFile(out_html_path, contents)
```

IMO

utils/use_require.js
DirectXMan12: plus, we should pre-declare promisified versions at the top for all promisified functions for readability inline

utils/use_require_helpers.js
```diff
@@ -21,17 +32,29 @@ module.exports = {
...
             var b = browserify(path.join(base_out_path, 'app/ui.js'), {});
-            b.bundle().pipe(fs.createWriteStream(out_path));
-            return `<script src="${path.relative(base_out_path, out_path)}"></script>`;
+            return new Promise((resolve, reject) => {
+                b.bundle((err, buf) => {
+                    if (err) return reject(err);
+                    resolve(buf);
+                });
+            })
+            .then((buf) => {
```
DirectXMan12: `.then((buf) => writeFile(out_path, buff))`

No need to use blockform for single-expression returns.  It's more readable without the block

utils/use_require.js
```diff
@@ -231,7 +235,29 @@ var make_lib_files = function (import_format, source_maps, with_app_dir) {
...
+            if (!helper.removeModules) return;
+            console.log(`Cleaning up temporary files...`);
+            let unlink = path => new Promise((resolve, reject) => {
+                fse.unlink(path, err => {
+                    if (err) return reject(err);
+                    resolve();
+                });
+            });
+            let rmdir_r = dir => new Promise((resolve, reject) => {
+                fse.rmdir(dir, err => {
```
DirectXMan12: https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove.md

Use this for recursive removal

samhed:
This is not necessarily a blocker for 1.0.0 in my mind. @CendioOssman do you want this merged before a release?

DirectXMan12:
I'd say we can hold this off for a patch release, if we want to get 1.0.0 out the door.

CendioOssman:
Yeah, it would be nice but it's not critical.

utils/use_require.js
CendioOssman: Nice. I'll try to convert everything to that.

utils/use_require.js
CendioOssman: I only want to remove empty directories, so that function would not work.

utils/use_require.js
CendioOssman: Sure. Will fix.

utils/use_require_helpers.js
CendioOssman: Will fix.

CendioOssman:
> As for the last commit, if someone explicitly asks for module format X, we should give it to them, not give them ES6 module and fallback. They've asked explicitly for it. If they want to have fallback, it should be up to them.

If that was the case, I'd agree. But I'd say in most cases they just want something that works\<tm>. So when bundling the app they are specifying the fallback/legacy format, not the exclusive format.

Converting the library is a different matter though and there I agree with you. Fallback only makes sense for the app.

utils/use_require.js
CendioOssman: Crap. No go. That requires a fairly new version of Node.js, and at least Fedora does not ship that. So I think it puts a too high requirement at this time.

I'll try to do something similar within the boundaries of  Node.js 6 though.

CendioOssman:
New push with the promise wrappers cleaned up. Should be no other changes except that, and I added `process.exit(1)` on failure for easier testing.

utils/use_require.js
```diff
@@ -38,46 +40,95 @@ const no_transform_files = new Set([
 
 no_copy_files.forEach((file) => no_transform_files.add(file));
 
+// util.promisify requires Node.js 8.x, so we have our own
+function promisify(original) {
```
DirectXMan12: Any reason why we need this?  NodeJS 8.x is LTS (It's in F27 for Fedora, and SCL for CentOS/RHEL/etc, it's in sid for Debian, AFAICT)

DirectXMan12: Since this is mostly for devs that want to produce their own distribution and *don't* have their own ES6-to-whatever toolchain, I'd expect it to be relatively rare to find one that can't be on the LTS release

utils/use_require.js
```diff
@@ -38,46 +40,95 @@ const no_transform_files = new Set([
...
 
         var start_marker = '<!-- begin scripts -->\n';
         var end_marker = '<!-- end scripts -->';
         var start_ind = contents.indexOf(start_marker) + start_marker.length;
         var end_ind = contents.indexOf(end_marker, start_ind);
 
+        new_script = '';
+
+        // Fallback for legacy browsers
+        new_script += `\
```
DirectXMan12: I'd still prefer to at least have an option to disable this, but don't consider it a blocker

utils/use_require.js
CendioOssman: Fedora 26 is still a current version. And Sid is the unstable Debian, so I don't think we should require that people are running that.

The API is the same here, so my idea is that we replace it with `util.promisify()` some time in the future.

DirectXMan12: that's fine with me.

CendioOssman:
I got an option in place now. It also avoids duplicate versions when you're build the library as this only makes sense for the application. That was my intention from the start, but I screwed it up along the way.

Waiting for travis and then I'll merge.

PR #985, opened 2018-01-04
Fix view_only setting with autoconnect

Deni42:
I found an issue where the view_only parameter in the url was not respected during automatic connection.

To reproduce, open an url similar to this: .../noVNC/vnc.html?host=...&autoconnect=1&view_only=true&resize=none
After successful connection, the checkbox in the UI on the left side is checked for "
View only", but it is not respected. If I uncheck it and check it again, it is respected.

The following change fixes the bug, although I am not very familiar with the codebase, so this might not be the best solution.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1019,8 +1019,6 @@ var UI = {
         UI.closeAllPanels();
         UI.closeConnectPanel();
 
-        UI.updateVisualState('connecting');
-
         var url;
 
         url = UI.getSetting('encrypt') ? 'wss' : 'ws';
@@ -1047,6 +1045,7 @@ var UI = {
         UI.rfb.scaleViewport = UI.getSetting('resize') === 'scale';
         UI.rfb.resizeSession = UI.getSetting('resize') === 'remote';
 
+        UI.updateVisualState('connecting');
         UI.updateViewOnly();
     },
 
--- a/app/ui.js
+++ b/app/ui.js
@@ -1021,8 +1021,6 @@ var UI = {
 
         UI.updateVisualState('connecting');
 
-        UI.updateViewOnly();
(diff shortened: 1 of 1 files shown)


app/ui.js
```diff
@@ -1048,6 +1046,8 @@ var UI = {
         UI.rfb.clipViewport = UI.getSetting('view_clip');
         UI.rfb.scaleViewport = UI.getSetting('resize') === 'scale';
         UI.rfb.resizeSession = UI.getSetting('resize') === 'remote';
+
+        UI.updateViewOnly();
```
samhed: Could you move the `UI.updateVisualState('connecting')` down here as well?

Deni42:
OK, I updated it.

samhed:
thank you

CendioOssman:
> UI.updateVisualState and UI.updateViewOnly depend on that UI.rfb exists. They should be called after UI.rfb has been set.

Ehm... `updateViewOnly()` does, but not `updateVisualState()` does not. So now we've delayed showing the connecting transition. Not sure by how much though.

samhed:
I was talking about: https://github.com/novnc/noVNC/blob/master/app/ui.js#L442

CendioOssman:
Ah alright.

PS. That code looks a bit broken. What if the extra keys panel is open when view only is set?

samhed:
That `if` should be moved to `updateViewOnly()`, this will allow us to move the `updateVisualState()` call back to where it was before. I'll fix.

samhed:
Fixed in f3763a010cf743fdc6f97767c3ab9cb8ef77f5af and 61f93180c8cbdc6259d0b083b51f169053f66f8c

PR #986, opened 2018-01-05
Create tr.po file and update Turkish lang

farukomercakmak:
Create tr.po file and update Turkish lang


farukomercakmak:
I forgot, I edited it again

PR #990, opened 2018-01-09
Optimize PNG images losslessly using Google zopflipng

PeterDaveHello:
The image assets are now extremely and losslessly re-compressed as below which can save distribution package size and the loading bandwidth/time a little bit!
```
 app/images/icons/novnc-120x120.png | Bin 4028 -> 2757 bytes
 app/images/icons/novnc-144x144.png | Bin 4582 -> 3090 bytes
 app/images/icons/novnc-152x152.png | Bin 5216 -> 3613 bytes
 app/images/icons/novnc-16x16.png   | Bin 675 -> 414 bytes
 app/images/icons/novnc-192x192.png | Bin 5787 -> 4005 bytes
 app/images/icons/novnc-24x24.png   | Bin 1000 -> 695 bytes
 app/images/icons/novnc-32x32.png   | Bin 1064 -> 781 bytes
 app/images/icons/novnc-48x48.png   | Bin 1397 -> 1107 bytes
 app/images/icons/novnc-60x60.png   | Bin 1932 -> 1560 bytes
 app/images/icons/novnc-64x64.png   | Bin 1946 -> 1568 bytes
 app/images/icons/novnc-72x72.png   | Bin 2699 -> 1733 bytes
 app/images/icons/novnc-76x76.png   | Bin 2874 -> 1932 bytes
 app/images/icons/novnc-96x96.png   | Bin 2351 -> 1988 bytes
```


CendioOssman:
Thanks. Although I'm not sure how big a problem this is in practice. The browsers usually just downloads one or two of these, and their sizes are also rather small compared to the VNC data that will be sent afterwards.

However I am concerned that this will be undone the next time someone updates the files. Could you see if `app/images/icons/Makefile` also could be updated?

PeterDaveHello:
The compression process takes a while, to integrate with the Makefile may not be a good idea, so maybe to drop this change, to confirm the issue here, are those PNG images dynamically generated from SVG images? Thanks!

CendioOssman:
Yes, the PNG files are generated from the SVG files. We've checked in the generated files so that people don't need to have the conversion tools installed in order to use noVNC.

PeterDaveHello:
Will the `svg` -> `png` process be triggered every release? Or it'll only be triggered once the svg sources were changed?

CendioOssman:
We have nothing automatic at the moment, and I don't think we will for the foreseeable future either. So just when someone changes something.

PeterDaveHello:
Okay, so I just need to update `Makefile`, am I correct? Thanks.

CendioOssman:
Right, that should be enough.

samhed:
So, we close this?

PR #992, opened 2018-01-09
Update tr.po

farukomercakmak:


CendioOssman:
Translators aren't necessarily developers, so I think we can consider helping out a bit as long as people provide us with a sane `.po` file. :)

CendioOssman:
I committed this as 690d07baba325ce7e362181d0d544d146938963a.

Issue #966, opened 2017-12-03
Please document how to compile better


agronick:
I spent a few hours searching around the codebase and docs until I found this post #945 explaining how to compile the project.

There is the file docs/notes that looks like it might be obsolete saying to use browserify. I couldn't find anything in docs or the readme regarding use_require.js.

I think something that says run
`npm install <project dir>`
Folowed by
`utils/use_require.js --as <format>`
would be helpful to a lot of people.

samhed:
the wiki is editable by anyone :)

And just to clarify here, you don't have to do this. noVNC will work with the module system in very new versions of Chrome, Safari and Edge, but the slow fallback version is still used in older browsers and Firefox. But in production it's recommended that you instead make use of use_require.js due to the lack of full support across the browser platforms at the moment.

agronick:
I know. It was more of a personal preference and to go along with the rest of the codebase. Awesome tool by the way. It really helps us a lot.

agronick:
I added it to the wiki: https://github.com/novnc/noVNC/wiki/Integration

CendioOssman:
I've added upon that documentation and put it in the repo so that it matches the code. Please check the README for links.

PR #995, opened 2018-01-09
Update translation po/template strings

PeterDaveHello:
Current pot & po files are all dated, let's update them!

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: noVNC 0.6.1\n"
 "Report-Msgid-Bugs-To: novnc@googlegroups.com\n"
-"POT-Creation-Date: 2017-11-24 07:16+0000\n"
+"POT-Creation-Date: 2018-01-09 23:53+0800\n"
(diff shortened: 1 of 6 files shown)


CendioOssman:
The plan is to update these things for release, not randomly during development. It would make a very noisy history otherwise.

So thank you for keeping an eye on things, but I'm afraid we won't be merging this.

PeterDaveHello:
@CendioOssman can we have it more flexible here? The problem is that I translated from the old template which is annoying ... and I don't think very translator will notice it, know how to deal with it.

CendioOssman:
Right, but I don't think we necessarily want people translating from `master` as it is a moving target. Right now is a bit special since we've never made a release with translations, so hopefully it will be more sane once we get things rolling properly.

PeterDaveHello:
Got it, if you can accept, maybe just update the `pot`? If not, never mind.

PeterDaveHello:
Just that you know that I translated the project(#966), and then, I found that I was translating from an old template, so I needed to update the template and translate the new strings, drop the translation for the dropped strings, I think it waste time a little bit.

CendioOssman:
The plane is to start the release work next week, so that will mean an updated `.pot` file by then.

PR #998, opened 2018-01-10
Fix slice usage in display tests on IE11

DirectXMan12:
IE11 doesn't support the `slice` method on ArrayBufferViews, so
we have to use the old `Array.prototype.slice.call` trick in the
tests instead.


DirectXMan12:
See, for instance https://travis-ci.org/novnc/noVNC/jobs/326891590#L1679

PR #1005, opened 2018-01-28
Add eslint and refactor to ES6

juanjoDiaz:
Add ESlint and refactor the entire library to ES6.

A clear next step is to add refine the eslint configuration (either custom or standard air-bnb or similar) and add the linting to the build.


DirectXMan12:
so, as much as I would like to do this, we wanted to keep our transpiling to a minimum.  Since we're already transpiling for modules, it's not a big jump, but it's worth discussion.  I have *really* wanted to do this for a while, though.

@kanaka @samhed

juanjoDiaz:
Those were my thought's precisely.
We are already transpiling so it's just a matter of how much stuff we want to transpile.

In general all the ES6 stuff works just fine in all browsers except for IE11 so we might even give the option of use the raw modules or the transpiled modules to the user.
It's  not crazy to think that most users are going to be packaging noVNC using webpack, parcel or some other module bundler and are going to be doing their own transpilation for their code which could also be applied to noVNC.

As advantages:
* The code is noticeably cleaner and more maintainable
* eslint will help capturing issues as soon as possible.

CendioOssman:
I'm mildly concerned that doing this all at once will make `git blame` a little annoying, but I do like the idea of using some of these more modern constructs. But we need to make sure we don't create new requirements on transpiling. The goal should be that if a browser supports modules then it shouldn't need any conversion.

So which browsers would require transpiling of these things?

app/error-handler.js
```diff
@@ -9,22 +9,22 @@
     // Fallback for all uncought errors
     function handleError (event, err) {
         try {
-            var msg = document.getElementById('noVNC_fallback_errormsg');
+            const msg = document.getElementById('noVNC_fallback_errormsg');
```
CendioOssman: This file isn't transpiled when using things directly, so it cannot use any constructs that don't work on all browsers.

app/localization.js
```diff
@@ -10,36 +10,34 @@
...
-        }
 
-        for (var i = 0;i < userLanguages.length;i++) {
-            var userLang = userLanguages[i];
-            userLang = userLang.toLowerCase();
-            userLang = userLang.replace("_", "-");
-            userLang = userLang.split("-");
+        const userLanguages = (typeof window.navigator.languages == 'object')
+            ? window.navigator.languages
+            : [navigator.language || navigator.userLanguage];
```
CendioOssman: This could be a bit excessive style enforcement. Can we turn such things off?

app/localization.js
```diff
@@ -10,36 +10,34 @@
...
+
+        for (let i = 0;i < userLanguages.length;i++) {
+            const userLang = userLanguages[i]
+                .toLowerCase()
+                .replace("_", "-")
+                .split("-");
```
CendioOssman: Same thing here. People can use these things as suggestions, but I don't think we should have hard enforcement.

app/localization.js
```diff
@@ -48,11 +46,11 @@ Localizer.prototype = {
             }
 
             // First pass: perfect match
-            for (var j = 0;j < supportedLanguages.length;j++) {
-                var supLang = supportedLanguages[j];
-                supLang = supLang.toLowerCase();
-                supLang = supLang.replace("_", "-");
-                supLang = supLang.split("-");
+            for (let j = 0;j < supportedLanguages.length;j++) {
```
CendioOssman: I believe WebKit/Blink still has performance issues with `let` in loops, so we need to check that we don't get any performance regressions by using this.

core/util/browsers.js
```diff
@@ -8,24 +8,48 @@
 
 import * as Log from './logging.js';
 
+export function isMac() {
+    return navigator && !!(/mac/i).exec(navigator.platform);
+}
```
CendioOssman: These things are unrelated to linting, so please put them in a separate commit.

core/input/keyboard.js
```diff
@@ -5,70 +5,49 @@
...
+            'keypress': this._handleKeyPress.bind(this),
+            'blur': this._allKeysUp.bind(this)
+        };
+    }
 
-Keyboard.prototype = {
     // ===== EVENT HANDLERS =====
 
-    onkeyevent: function () {},     // Handler for key press/release
+    onkeyevent() {}     // Handler for key press/release
```
CendioOssman: Can we avoid this? It's more of a member variable than a method.

core/input/mouse.js
```diff
@@ -5,59 +5,57 @@
...
+            'mousemove': this._handleMouseMove.bind(this),
+            'mousewheel': this._handleMouseWheel.bind(this),
+            'mousedisable': this._handleMouseDisable.bind(this)
+        };
 
-Mouse.prototype = {
     // ===== PROPERTIES =====
 
-    touchButton: 1,                 // Button mask (1, 2, 4) for touch devices (0 means ignore clicks)
+    this.touchButton = 1;                 // Button mask (1, 2, 4) for touch devices (0 means ignore clicks)
```
CendioOssman: This did not convert nicely.

core/util/browsers.js
```diff
@@ -8,24 +8,48 @@
...
     isTouchDevice = true;
     window.removeEventListener('touchstart', onFirstTouch, false);
 }, false);
 
-var _cursor_uris_supported = null;
+let _cursor_uris_supported = null;
 
 export function browserSupportsCursorURIs () {
-    if (_cursor_uris_supported === null) {
+    if (_cursor_uris_supported !== null) {
```
CendioOssman: Huh? How did this sneak in. I'm getting concerned that this commit is more than just a simple run of eslint. Please separate things or we cannot safely review this.

core/util/polyfill.js
```diff
@@ -38,17 +37,17 @@ if (typeof Object.assign != 'function') {
...
-        var evt = document.createEvent( 'CustomEvent' );
+        const evt = document.createEvent( 'CustomEvent' );
         evt.initCustomEvent( event, params.bubbles, params.cancelable, params.detail );
         return evt;
     }
 
     CustomEvent.prototype = window.Event.prototype;
 
-    if (typeof window.CustomEvent !== "function") {
+    if (typeof window.CustomEvent !== 'function') {
```
CendioOssman: This should probably also be disabled or we'll have a lot of noise once eslint is run.

docs/API.md
```diff
@@ -137,7 +137,7 @@ connection to a specified VNC server.
 
 ##### Syntax
 
-    var rfb = new RFB( target, url [, options] );
+    const rfb = new RFB( target, url [, options] );
```
CendioOssman: Bike shedding, but I'd say `let` is more general and should be in the examples.

tests/fake.websocket.js
```diff
@@ -1,69 +1,71 @@
+import Base64 from '../core/base64.js';
```
CendioOssman: Separate commit.

tests/playback-ui.js
```diff
@@ -1,18 +1,19 @@
...
-    console.log(str);
-    var cell = document.getElementById('messages');
+    const cell = document.getElementById('messages');
     cell.textContent += str + "\n";
     cell.scrollTop = cell.scrollHeight;
 }
 
 function loadFile() {
-    const fname = WebUtil.getQueryVar('data', null);
+    const fname = WebUtil.getQueryconst('data', null);
```
CendioOssman: Something went very wrong here. `getQueryconst` does not exist. Looks like some kind of manual search-and-replace for `var`.

juanjoDiaz:
That's for the review.
AFAIK, only IE11 needs transpilation. All other browsers work just fine.

I agree that maybe I've tackled too much with the PR.
I'l start sending smaller PRs and then we can review those case by case. End result should be the same :)

docs/API.md
juanjoDiaz: It's a common practice nowadays to use const whenever possible.
It makes code more solid since you know that the variable won't change since it's declaration.
Also helps to get more "functional" code.

but yeah. It's trivial... I can change to let if you prefer that :)

core/input/mouse.js
juanjoDiaz: Yup. Declaring variables as `touchButton = 1` as you can do with methods is ES7 so it would need transpilation to all browsers.

So the way to declare value properties in ES6 is to declare them like this in the constructor.
I've put them always at the bottom so it should be trivial to take them out once ES7 is more widespread.

core/input/keyboard.js
juanjoDiaz: It is a method isn't it?
It's just a method that can be overridden.

app/localization.js
juanjoDiaz: We can check that.
I assume that the performance difference is because using let make the loop into its own scope.

That's what you would expect if you come from any other language and makes things more solid in my opinion.

app/localization.js
juanjoDiaz: This is not really enforced. I change it so user lang is a const.

As I said below, most ES6 good practices guidelines recommend using const whenever possible.

app/localization.js
juanjoDiaz: Same as above.

app/error-handler.js
juanjoDiaz: Do you mean the whole app folder?

core/input/keyboard.js
DirectXMan12: it's a callback

app/localization.js
CendioOssman: Ah, I missed the const thing. That's fine in that case. But for clarity it would be good to separate the things you manually restructured to get things to be `const` and the things `eslint` automatically converted.

app/localization.js
CendioOssman: Right, I really like these new features. But we need to make sure we aren't causing problems in other ways. A quick test using `vnc_playback.html` should answer that.

core/input/keyboard.js
CendioOssman: Right, so perhaps use the same style as for other properties.

core/input/mouse.js
CendioOssman: Urgh. Are there any other patterns that can be used? I'd like to avoid the confusion of putting it in the constructor with other stuff. And for `RFB` it gets extra messy as it has getters and setters for some properties but not all.

docs/API.md
CendioOssman: Right. You just need to separate the case where it was truly meant to never change, and when the code just happened to end up that way. There is some intent implied when you use `const`.

I don't have any really strong emotions about this particular case. I was just thinking about how we use it in `ui.js` where it is very much not a constant.

core/input/mouse.js
juanjoDiaz: I think that we could use https://babeljs.io/docs/plugins/transform-class-properties/
```
class MyClass {
    prop = true

    function() { ... }
}
```

However we would need to transpile it for all browsers AFAIK. It's not even part of es2017 preset.

CendioOssman: That's a no-go in that case. We need something that everyone (except IE) supports.

juanjoDiaz: Agreed. That's why I opted for adding them at the bottom of the constructor with a comment stating that they are properties. Once the class properties become a standard, it's trivial to move them out of there. :)

juanjoDiaz:
Superseded by #1013

CendioOssman:
You can update existing PRs without having to open new ones. Avoids having to dig around for previous comments on things.

juanjoDiaz:
Yeah. I opened a new no to don't pile up changes or override existing commits in this one.

I know that at the end you can squash commits and that solves the issue. But the commits in #1013 are very self-contained and I think that they shouldn't be squashed in order to be able to see changes one by one and fix any regression bug that it might indadvertedly introduce.

CendioOssman:
If you do a force push you can completely nuke the old stuff. That allows us to continue the discussion here, even when changing a commit rather than adding new ones on top of it.

juanjoDiaz:
@CendioOssman I've been doing force pushes to #1013 

It was my bad to open a new one in  the first place but there are many comments there from the other noVNC maintainers so it's sort of too late to come back here, isn't it? :(

CendioOssman:
No worries. Next time. :)

PR #1007, opened 2018-01-30
Remove jshint comments

juanjoDiaz:
As mentioned in #1005 let's do a bunch of small PRs vs a huge one.

This remove jshint comments.
jshint is not used anyway and it will be replaced by eslint.

No code touched. Just comments.


vendor/sinon.js
```diff
@@ -9864,7 +9864,7 @@ process.umask = function() { return 0; };
         // Unlike global isNaN, this avoids type coercion
         // typeof check avoids IE host object issues, hat tip to
         // lodash
-        var val = value; // JsLint thinks value !== value is "weird"
```
DirectXMan12: No need to touch the vendored sinon.  It's more-or-less copied straight (the upstream package pulls in an unnecessary dep, so we don't use it).

juanjoDiaz: Undone!

CendioOssman:
So squash these to a single commit, and perhaps we can even merge this before 1.0?

juanjoDiaz:
Squashed!

PR #1009, opened 2018-01-30
Move browser checks to browser.js

juanjoDiaz:
Just move all the duplicate `isMac`, `isWindow`, etc. to `browser.js`

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -13,7 +13,7 @@
 
 import * as Log from '../core/util/logging.js';
 import _, { l10n } from './localization.js';
-import { isTouchDevice } from '../core/util/browsers.js';
+import { isTouchDevice } from '../core/util/browser.js';
 import { setCapture, getPointerEvent } from '../core/util/events.js';
 import KeyTable from "../core/input/keysym.js";
 import keysyms from "../core/input/keysymdef.js";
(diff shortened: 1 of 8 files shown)


juanjoDiaz:
Any ETA for that?
I have more PRs for the eslint stuff sort of waiting for this to avoid conflicts.

CendioOssman:
> Any ETA for that?
> I have more PRs for the eslint stuff sort of waiting for this to avoid conflicts.

I was hoping to get the beta out this week, but we probably want to let that be tested for a while before doing the final version. So it could be a while.

Perhaps it's better if you just use #1005 and put each suggestion as a separate commit rather than a separate PR? That way everything can be reviewed and done without having to wait for partial merges.

DirectXMan12:
I've been gradually merging stuff that's not majorly impactful.  This is relatively minor, so it might be fine to merge pre-1.0, especially with the plan to do a beta.

juanjoDiaz:
That would be great :)

DirectXMan12:
@CendioOssman @samhed any objections to just merging this now?

samhed:
works for me

CendioOssman:
Yeah, this isn't too scary. So go ahead.

Issue #799, opened 2017-03-24
Auto reconnect only reconnects to last connection


steve13814:
Open one vnc.html page, enable auto reconnect and connect to VM A (by specifying VM A's ip, port). And then open another vnc.html page and connect to VM B (by specifying VM B's ip port). When VM A's vnc.html page triggered reconnect, it actually reconnect to the VM B, not VM A.

I think this is because the local storage only save one instance of setting, so later connection override this setting. I've record a video to show how this can be reproduced.
https://www.youtube.com/watch?v=0b3zX-714Jk

samhed:
Yes, you are correct, this is due to local storage.

DirectXMan12:
I don't think there's much we can do about this.  I'm going to close, but feel free to reopen if you feel otherwise.

steve13814:
I think this is a serious problem if the user is an IT working on multiple VMs. Auto reconnect might mislead the user.
There are still some workarounds available on this issue. For example, generate unique id for each connection and save multiple instance in local storage. Another method is to reread those options from the form and the URL again.
What's your opinion about the two methods?

CendioOssman:
I'm with @steve13814 on this. This does not seem like desired behaviour. I think we need to have another look at how configuration works in the case of multiple tabs.

DirectXMan12:
we could try to prefer sessionStorage over localStorage (apologies, I failed to recall sessionStorage when I closed this), but only the local storage would persist across close and re-open, I believe.

Craigzyc:
Why does reconnect need to be persisted at all? If you set auto reconnect I would expect it to exist only for the current session, not after a reload.

samhed:
Because there are cases where you'd want a session to remain connected no matter how many times you get disconnected.

Craigzyc:
I get that but a disconnect is not a page refresh so storing it in a local variable still works in that situation. 
It only needs to go to localStorage to persist across full page reloads which seems counter intuitive from a security standpoint and is what's causing the issue mentioned. 

If it was stored in a variable each window would auto reconnect to the session its supposed to which seems like a far better idea

samhed:
Sorry I misunderstood you, I see your point. I tend to agree tbh.

CendioOssman:
Since #1010 is merged this should now be fixed. Could someone retest?

CendioOssman:
No response. Assuming it works now.

PR #1010, opened 2018-01-30
Use localstorage only to initialize settings map

andrwwbstr:
This makes query overrides temporary (i.e. they do not change the value, if any, in local storage).  I believe it also solves #799.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -731,7 +731,7 @@ var UI = {
 
         // Save the cookie for this session
         if (typeof value !== 'undefined') {
-            WebUtil.writeSetting(name, value);
+            WebUtil.setSetting(name, value);
         }
 
         // Update the settings control
--- a/app/ui.js
+++ b/app/ui.js
@@ -721,21 +721,17 @@ var UI = {
         if (val === null) {
             val = WebUtil.readSetting(name, defVal);
         }
-        UI.updateSetting(name, val);
+        WebUtil.setSetting(name, val);
+        UI.updateSetting(name);
         return val;
     },
 
     // Update cookie and form control setting. If value is not set, then
     // updates from control to current cookie setting.
-    updateSetting: function(name, value) {
-
-        // Save the cookie for this session
-        if (typeof value !== 'undefined') {
-            WebUtil.setSetting(name, value);
-        }
+    updateSetting: function(name) {
 
         // Update the settings control
-        value = UI.getSetting(name);
+        var value = UI.getSetting(name);
 
         var ctrl = document.getElementById('noVNC_setting_' + name);
         if (ctrl.type === 'checkbox') {
--- a/app/webutil.js
+++ b/app/webutil.js
@@ -117,21 +117,25 @@ export function initSettings (callback /*, ...callbackArgs */) {
             }
         });
     } else {
-        // No-op
+        settings = {};
         if (callback) {
             callback.apply(this, callbackArgs);
         }
     }
 };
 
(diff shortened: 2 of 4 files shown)


app/ui.js
```diff
@@ -734,7 +734,7 @@ var UI = {
 
         // Save the cookie for this session
         if (typeof value !== 'undefined') {
-            WebUtil.writeSetting(name, value);
+            WebUtil.writeSetting(name, value, true);
```
CendioOssman: The only place that provides a `value` is the call from `initSetting()`. So I think we can take this opportunity to move this chunk to `initSetting()` and get rid of the `value` argument.

app/webutil.js
```diff
@@ -121,22 +121,22 @@ export function initSettings (callback /*, ...callbackArgs */) {
             }
         });
     } else {
-        // No-op
+        settings = {};
```
CendioOssman: Why is this added?

app/webutil.js
```diff
@@ -121,22 +121,22 @@ export function initSettings (callback /*, ...callbackArgs */) {
...
         if (callback) {
             callback.apply(this, callbackArgs);
         }
     }
 };
 
 // No days means only for this browser session
-export function writeSetting (name, value) {
+export function writeSetting (name, value, temporary) {
```
CendioOssman: "write" and "temporary" seems like a confusing mix to me. Perhaps introduce a new `setSetting()` that only updates the `settings` object?

app/webutil.js
```diff
@@ -162,9 +163,9 @@ export function readSetting (name, defaultValue) {
 
 export function eraseSetting (name) {
     "use strict";
+    settings[name] = undefined;
```
CendioOssman: Why not `delete` to get things fully back to before it was ever set?

app/webutil.js
andrwwbstr: Mostly for the tests so that the settings object can be reset between tests.  However, if `initSettings` were to be called again (which it isn't), it makes sense to reset the settings object.

app/webutil.js
andrwwbstr: If the setting was deleted, then the next time it is read it would be restored back to the value in localstorage.  If localstorage was modified by another tab between the erase and the next read, then the read value would be inconsistent.

It may be a moot point though since it seems that `eraseSetting` is never used.  Would it just be better to remove it instead?

CendioOssman: Not entirely clear what we want here since it isn't called. But if I read the code correctly this will nuke it from local storage as well. So the possibly confusing case is the one you describe, when another tab does a write after the first tab has done an erase.

I think a `delete` would be safer for now and we can look over this once the API starts getting use. You could add a comment about this confusion there if you'd like.

app/webutil.js
CendioOssman: Ah, makes sense.

app/webutil.js
```diff
@@ -128,12 +128,17 @@ export function initSettings (callback /*, ...callbackArgs */) {
     }
 };
 
+// Update the settings cache, but do not write to permanent storage
+export function setSetting (name, value) {
+    "use strict";
```
CendioOssman: These are no longer needed, so let's not add new ones.

CendioOssman:
Well, rather it will be okay to merge once 1.0.0 is out.

CendioOssman:
Do you think you could rebase those into the original commits? :)

PR #1013, opened 2018-02-01
Add eslint and fix reported issues

juanjoDiaz:
New attempt to move the library to ES6.

I'd start pushing more simple commits so it's simple to review.


juanjoDiaz:
Ok. This is good now. I hope that the smaller and more focused commits help reviewing the changes.

This currently uses `eslint:recommended` config. That's the loosest configuration and doesn't really enforce anything (for example, it doesn't even enforce using let/const instead of var). So the next step is to put it a more strict eslint config and fix the code accordingly.

I'd suggest [airbnb config](https://github.com/airbnb/javascript) since it's widely used and imho it's rock solid. Wdyt??

juanjoDiaz:
Hei guys,

Any ETA of when this could be merge?
I have a couple other PRs coming after this one :)

CendioOssman:
> I'd suggest airbnb config since it's widely used and imho it's rock solid. Wdyt??

Do a separate commit with that config applied and we can see what it will do to the code.

> Any ETA of when this could be merge?
> I have a couple other PRs coming after this one :)

That would be after 1.0 is out. So unfortunately you got hit with some bad timing here. :/

app/ui.js
```diff
@@ -1564,15 +1548,14 @@ const UI = {
         }
 
         const blist = [0, 1,2,4];
-        for (let b = 0; b < blist.length; b++) {
-            const button = document.getElementById('noVNC_mouse_button' +
-                                                 blist[b]);
-            if (blist[b] === num && !view_only) {
+        blist.forEach((b) => {
+            const button = document.getElementById('noVNC_mouse_button' + b);
```
DirectXMan12: template literal

tests/test.display.js
```diff
@@ -353,15 +353,12 @@ describe('Display/Canvas Helper', function () {
             let large_checked_data = new Uint8Array(16*16*4);
             display.resize(16, 16);
 
-            for (let y = 0;y < 16;y++) {
-                for (let x = 0;x < 16;x++) {
-                    let pixel;
-                    if ((x < 4) && (y < 4)) {
+            for (let y = 0; y < 16; y++) {
+                for (let x = 0; x < 16; x++) {
+                    const pixel = ((x < 4) && (y < 4))
```
DirectXMan12: this is a fairly long use of the ternary operator.  It's clearer as an if statement

core/des.js
```diff
@@ -41,7 +41,7 @@
  * NAVIGATION OR COMMUNICATION SYSTEMS, AIR TRAFFIC CONTROL, DIRECT LIFE
  * SUPPORT MACHINES, OR WEAPONS SYSTEMS, IN WHICH THE FAILURE OF THE
  * SOFTWARE COULD LEAD DIRECTLY TO DEATH, PERSONAL INJURY, OR SEVERE
- * PHYSICAL OR ENVIRONMENTAL DAMAGE ("HIGH RISK ACTIVITIES").  WIDGET WORKSHOP
+ * PHYSICAL OR ENVIRONMENTAL DAMAGE ('HIGH RISK ACTIVITIES').  WIDGET WORKSHOP
```
DirectXMan12: changing the license is probably not the best idea

utils/genkeysymdef.js
```diff
@@ -61,6 +61,7 @@ const codepoints = str.split('\n')
               }
           }
       }
+      return codepoints;
```
DirectXMan12: squash this into the previous commit where this changed?

core/rfb.js
```diff
@@ -921,11 +921,9 @@ export default class RFB extends EventTargetMixin {
      */
     _handle_security_failure(context, security_result_status) {
 
-        if (typeof context === 'undefined') {
-            context = '';
-        } else {
-            context = ' on ' + context;
-        }
+        context = typeof context === 'undefined'
```
DirectXMan12: these ternary operators really don't make things clearer, IMO

core/rfb.js
```diff
@@ -942,24 +940,16 @@ export default class RFB extends EventTargetMixin {
...
+            'securityfailure',
+            { detail: reason === ''
+                ? { status: security_result_status }
+                : { status: security_result_status, reason }
+            });
+        this.dispatchEvent(event);
 
-            return this._fail('Security negotiation failed' + context);
-        }
+        return this._fail('Security negotiation failed' + context +
```
DirectXMan12: template literal

app/ui.js
```diff
@@ -8,15 +8,15 @@
...
-import * as Log from '../core/util/logging.js';
-import _, { l10n } from './localization.js';
-import { isTouchDevice } from '../core/util/browser.js';
-import { setCapture, getPointerEvent } from '../core/util/events.js';
-import KeyTable from '../core/input/keysym.js';
-import keysyms from '../core/input/keysymdef.js';
-import Keyboard from '../core/input/keyboard.js';
-import RFB from '../core/rfb.js';
-import * as WebUtil from './webutil.js';
+import * as Log from '../core/util/logging';
```
DirectXMan12: These need to stay in.  Support for bare names is unclear in the WHATWG spec, and the module loader polyfill doesn't permit it: https://github.com/ModuleLoader/browser-es-module-loader

app/localization.js
```diff
@@ -31,7 +31,7 @@ export class Localizer {
             ? window.navigator.languages
             : [navigator.language || navigator.userLanguage];
 
-        for (let i = 0; i < userLanguages.length; i++) {
+        for (let i = 0; i < userLanguages.length; i += 1) {
```
DirectXMan12: just because Crockford is afraid of the increment operator doesn't mean we should be too.  Please leave these in.

core/util/logging.js
```diff
@@ -17,7 +17,7 @@ let Info = () => {};
 let Warn = () => {};
 let Error = () => {};
 
-export function init_logging (level) {
+export function init_logging(level) {
```
DirectXMan12: this doesn't look normalized -- some have spaces between parens and the name, some don't.  I prefer spaces.

app/localization.js
```diff
@@ -152,5 +152,5 @@ export class Localizer {
     }
 }
 
-export let l10n = new Localizer();
+export const l10n = new Localizer();
```
DirectXMan12: squash all these into the previous let/const/var commit, if possible

app/ui.js
```diff
@@ -130,11 +128,11 @@ const UI = {
         }
     },
 
-    initSettings: function() {
+    initSettings() {
         // Logging selection dropdown
         const llevels = ['error', 'warn', 'info', 'debug'];
         for (let i = 0; i < llevels.length; i += 1) {
-            UI.addOption(document.getElementById('noVNC_setting_logging'), llevels[i], llevels[i]);
+           UI.addOption(document.getElementById('noVNC_setting_logging'), llevels[i], llevels[i]);
```
samhed: This line was properly indented before this change

app/ui.js
```diff
@@ -1643,12 +1640,12 @@ const UI = {
...
 l10n.setup(LINGUAS);
 if (l10n.language !== "en" && l10n.dictionary === undefined) {
-    WebUtil.fetchJSON('app/locale/' + l10n.language + '.json', function (translations) {
+    WebUtil.fetchJSON('app/locale/' + l10n.language + '.json', (translations) => {
         l10n.dictionary = translations;
 
         // wait for translations to load before loading the UI
         UI.prime();
-    }, function (err) {
+    }, (err) => {
```
samhed: Should be able to skip the '{' and '}' here right?

core/rfb.js
samhed: +1

core/rfb.js
```diff
@@ -942,24 +940,16 @@ export default class RFB extends EventTargetMixin {
...
-                'securityfailure',
-                { detail: { status: security_result_status } });
-            this.dispatchEvent(event);
+        const event = new CustomEvent(
+            'securityfailure',
+            { detail: reason === ''
+                ? { status: security_result_status }
+                : { status: security_result_status, reason }
+            });
+        this.dispatchEvent(event);
```
samhed: in other places you have removed the intermediate event variable in cases like these

app/localization.js
samhed: agreed

core/util/strings.js
```diff
@@ -9,6 +9,6 @@
 /*
  * Decode from UTF-8
  */
-export function decodeUTF8 (utf8string) {
+export function decodeUTF8(utf8string) {
```
samhed: this commit doesn't seem to bring consistency at all!

app/localization.js
```diff
@@ -48,7 +48,8 @@ Localizer.prototype = {
             }
 
             // First pass: perfect match
-            for (var j = 0;j < supportedLanguages.length;j++) {
+            var j;
```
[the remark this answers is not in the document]
juanjoDiaz: Yeah. I tried to keep every commit as small as possible.

core/rfb.js
```diff
@@ -1062,7 +1060,7 @@ RFB.prototype = {
         var serverSupportedTypes = [];
 
         for (var i = 0; i < subAuthCount; i++) {
-            var capNum = this._sock.rQshift32();
+            this._sock.rQshift32();
```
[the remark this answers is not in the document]
juanjoDiaz: Same as above. I added the comment in the next commit.

core/util/logging.js
```diff
@@ -24,8 +24,10 @@ export function init_logging (level) {
         _log_level = level;
     }
 
-    Debug = Info = Warn = Error = function (msg) {};
+    Debug = Info = Warn = Error = function () {};
```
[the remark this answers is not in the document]
juanjoDiaz: They do. In lines 15-18

package.json
```diff
@@ -7,7 +7,8 @@
     "test": "tests"
   },
   "scripts": {
-    "test": "PATH=$PATH:node_modules/karma/bin karma start karma.conf.js",
+    "lint": "eslint app core tests utils",
+    "test": "karma start karma.conf.js",
```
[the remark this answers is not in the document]
juanjoDiaz: scripts in NPM automatically look in the `node_modules` folder.

tests/test.keyboard.js
```diff
@@ -133,12 +132,12 @@ describe('Key Event Handling', function() {
...
-                var evt = keyevent('keydown', {code: 'KeyA', key: 'a'});
-                kbd._handleKeyDown(evt);
-                expect(evt.preventDefault).to.have.been.called;
-                evt = keyevent('keyup', {code: 'KeyA', key: 'a'});
-                kbd._handleKeyUp(evt);
-                expect(evt.preventDefault).to.have.been.called;
+                var evt1 = keyevent('keydown', {code: 'KeyA', key: 'a'});
+                kbd._handleKeyDown(evt1);
+                expect(evt1.preventDefault).to.have.been.called;
+                var evt2 = keyevent('keyup', {code: 'KeyA', key: 'a'});
```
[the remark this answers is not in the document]
juanjoDiaz: We can. This change was looking forward to the next commit and the javascript best practice of using immutable variables (const) whenever possible.

utils/use_require.js
```diff
@@ -97,11 +97,12 @@ var make_lib_files = function (import_format, source_maps, with_app_dir) {
     const babel = require('babel-core');
 
     var in_path;
+    var out_path_base;
```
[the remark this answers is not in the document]
juanjoDiaz: Yup. But it obscures code. It's sort of good practice that, if using vars, they all should be declare on top of the scope.
Anyway. This will change to `let` so it will be scoped.

app/ui.js
```diff
@@ -1651,7 +1640,7 @@ var UI = {
 };
 
 // Set up translations
-var LINGUAS = ["de", "el", "es", "nl", "pl", "sv", "tr", "zh"];
+const LINGUAS = ["de", "el", "es", "nl", "pl", "sv", "zh"];
```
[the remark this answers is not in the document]
juanjoDiaz: This might have been a rebase problem or something like that. I'll re-add it.

app/webutil.js
```diff
@@ -201,13 +204,14 @@ export function injectParamIfMissing (path, param, value) {
...
-    var req = new XMLHttpRequest();
+    const req = new XMLHttpRequest();
     req.open('GET', path);
 
     req.onload = function () {
         if (req.status === 200) {
+            let resObj;
             try {
-                var resObj = JSON.parse(req.responseText);
+                resObj = JSON.parse(req.responseText);
```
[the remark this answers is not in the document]
juanjoDiaz: Same thing about the scope.
It's counter-intuitive for most people (specially the ones coming for non-js languages) to declare a variable in the try (even though var is not scoped) and use it outside.

In any case, I improved this one in one of the next commits.

core/display.js
```diff
@@ -677,13 +671,13 @@ Display.changeCursor = function (target, pixels, mask, hotx, hoty, w, h) {
...
-    var canvas = document.createElement('canvas');
-    var ctx = canvas.getContext('2d');
+    const canvas = document.createElement('canvas');
+    const ctx = canvas.getContext('2d');
 
     canvas.width = w;
     canvas.height = h;
 
-    var img;
+    let img;
```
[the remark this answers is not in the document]
juanjoDiaz: As I mentioned above I think that we should completely avoid using var since its scoping is not 100% clear to everyone and error prone.

core/rfb.js
```diff
@@ -699,7 +697,7 @@ RFB.prototype = {
 
     _setCapability: function (cap, val) {
         this._capabilities[cap] = val;
-        var event = new CustomEvent("capabilities",
+        const event = new CustomEvent("capabilities",
```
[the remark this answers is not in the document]
juanjoDiaz: I will :)

core/rfb.js
```diff
@@ -1407,17 +1405,14 @@ RFB.prototype = {
...
-        if (this._FBU.rects > 0) {
-            msg_type = 0;
-        } else {
-            msg_type = this._sock.rQshift8();
-        }
+        const msg_type = this._FBU.rects > 0
+            ? 0
+            : this._sock.rQshift8();
 
+        let first, ret;
```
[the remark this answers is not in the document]
juanjoDiaz: Same thing about the scope. This is the eslint rule: https://eslint.org/docs/rules/no-case-declarations

This is one of the few rules that's automatically enforced by eslint:recommend and also by most Javascript guidelines (airbnb, facebook, google, etc.)

docs/API.md
```diff
@@ -137,7 +137,7 @@ connection to a specified VNC server.
 
 ##### Syntax
 
-    var rfb = new RFB( target, url [, options] );
+    const rfb = new RFB( target, url [, options] );
```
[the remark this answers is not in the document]
juanjoDiaz: The point is that you should normally use const in whatever variable is immutable. And rfb probably is.

But I'm happy changing it back. Would you be happier with `let`? Or do you want to keep `var`?

karma.conf.js
```diff
@@ -1,29 +1,32 @@
...
-  if (useSauce && process.env.TEST_BROWSER_NAME && process.env.TEST_BROWSER_NAME != 'PhantomJS') {
-    var names = process.env.TEST_BROWSER_NAME.split(',');
-    var platforms = process.env.TEST_BROWSER_OS.split(',');
-    var versions = [];
-    if (process.env.TEST_BROWSER_VERSION) {
-      versions = process.env.TEST_BROWSER_VERSION.split(',');
-    } else {
-      versions = [null];
-    }
+  // Only IE need transpilation
```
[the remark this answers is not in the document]
juanjoDiaz: I guess if you use a very old chrome or firefox, you might need it. But not in the newer versions.

I've avoided using ES7 and non-standard-yet type of features just for that.

tests/playback.js
```diff
@@ -153,9 +153,9 @@ RecordingPlayer.prototype = {
             return;
         }
 
-        const frame = this._frames[this._frame_index];
-        var start = frame.indexOf('{', 1) + 1;
-        var u8;
+        let frame = this._frames[this._frame_index];
```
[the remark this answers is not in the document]
juanjoDiaz: It's back to a `const` in a subsequent commit.

tests/test.display.js
```diff
@@ -1,42 +1,47 @@
...
-        0x00, 0x00, 0xff, 255, 0x00, 0x00, 0xff, 255, 0x00, 0xff, 0x00, 255, 0x00, 0xff, 0x00, 255,
-        0x00, 0xff, 0x00, 255, 0x00, 0xff, 0x00, 255, 0x00, 0x00, 0xff, 255, 0x00, 0x00, 0xff, 255,
-        0x00, 0xff, 0x00, 255, 0x00, 0xff, 0x00, 255, 0x00, 0x00, 0xff, 255, 0x00, 0x00, 0xff, 255
-    ];
-    checked_data = new Uint8Array(checked_data);
-
-    var basic_data = [0xff, 0x00, 0x00, 255, 0x00, 0xff, 0x00, 255, 0x00, 0x00, 0xff, 255, 0xff, 0xff, 0xff, 255];
-    basic_data = new Uint8Array(basic_data);
+    const checked_data = new Uint8Array([
+        0x00, 0x00, 0xff, 255, 0x00, 0x00, 0xff, 255,
```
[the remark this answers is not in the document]
juanjoDiaz: Just trying to keep the lines short and readable. I can put it back as it was before if you want.

core/display.js
```diff
@@ -10,109 +10,111 @@
...
         return this._fb_height;
-    },
-
-    logo: null,
+    }
 
     // ===== EVENT HANDLERS =====
 
-    onflush: function () {},        // A flush request has finished
+    onflush() {}        // A flush request has finished
```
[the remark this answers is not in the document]
juanjoDiaz: Yup. I changed this in one of the next commits.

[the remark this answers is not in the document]
juanjoDiaz: I thought I did but not yet. Will be my next commit.

core/websock.js
```diff
@@ -14,141 +14,136 @@
...
+            'message': () => {},
+            'open': () => {},
+            'close': () => {},
+            'error': () => {}
+        };
+    }
+
     // Getters and Setters
-    get_sQ: function () {
+    get_sQ() {
```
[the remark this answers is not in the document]
juanjoDiaz: yup. That's also a future commit.

app/ui.js
juanjoDiaz: Yup, there will be a future commit converting all string concatenations to string literals.
Just trying to go little by little :)

tests/test.display.js
juanjoDiaz: This is just to keep the variable as constant instead of keeping it mutable. As mentioned before const should be preferred since they make the code more solid in general.

But I can accept that this might be a bit big ternary operator.

app/ui.js
juanjoDiaz: I can bring it back.

It's weird though, that it hadn't fail any of the tests...

app/localization.js
juanjoDiaz: Sure.

I found `++i`, `i++` and `i +=1` all over the code. that's why I tried to normalize with one.

`++i` vs `i++` was making not so clear to me what the code was doing in some cases so I preferred `i += 1`.

In the case of for loops I agree that's irrelevant. But you can't tell eslint to enforce it everywhere except for for loops...

core/util/logging.js
```diff
@@ -12,10 +12,10 @@
 
 var _log_level = 'warn';
 
-var Debug = function (msg) {};
-var Info = function (msg) {};
-var Warn = function (msg) {};
-var Error = function (msg) {};
```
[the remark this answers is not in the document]
juanjoDiaz: Nope. Arguments are never required in javascript. And those are never used anyway. It's just a default no-op

app/localization.js
```diff
@@ -21,25 +21,21 @@ export function Localizer() {
...
          * Fall back to navigator.language for other browsers
          */
-        if (typeof window.navigator.languages == 'object') {
-            userLanguages = window.navigator.languages;
-        } else {
-            userLanguages = [navigator.language || navigator.userLanguage];
-        }
+        const userLanguages = (typeof window.navigator.languages === 'object')
+            ? window.navigator.languages
+            : [navigator.language || navigator.userLanguage];
```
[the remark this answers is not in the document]
juanjoDiaz: I see that you all really dislike ternary operators.

We differ on that. But I can keep the if-else blocks.

As mentioned above, the main advantage of if-else is to declare variable as const when they are inmutable after declaration.

app/ui.js
juanjoDiaz: yup. But this change was just changed on master

core/util/strings.js
juanjoDiaz: **Anonymous function**
function () {....} // One single space between function and opening parenthesis and between closing parenthesis and opening curly brace

**Named function**
function funcName() {....} // No space between function name and opening parenthesis and single space between closing parenthesis and opening curly brace

It's the airbnb recommended way. I can follow some other way if you would prefer that.

juanjoDiaz:
Thanks for the thorough review.

I think that our main point of disagreement is the ternary operator. I can live with that :)

I'll try to fix all the comments.

core/rfb.js
```diff
@@ -718,6 +715,7 @@ RFB.prototype = {
                 Log.Error("Got data while disconnected");
                 break;
             case 'connected':
+                // eslint-disable-next-line no-constant-condition
```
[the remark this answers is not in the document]
juanjoDiaz: Yup. We could disable it globally. However, I think that's reasonable to assume that a condition on a constant should be an exception and not the norm. So it's disabled globaly and enabled only for this line.

core/util/polyfill.js
```diff
@@ -10,7 +10,7 @@
 if (typeof Object.assign != 'function') {
     // Must be writable: true, enumerable: false, configurable: true
     Object.defineProperty(Object, "assign", {
-        value: function assign(target, varArgs) { // .length of function is 2
```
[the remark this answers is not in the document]
juanjoDiaz: It's an argument that's not used.

To add it and not have eslint failing we would need to either add a comment to disable the rule here or disable eslint checks for the entire file (which might be trouble if we add here other polyfills.)

docs/API.md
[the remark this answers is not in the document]
samhed: `let` is fine for me

core/util/polyfill.js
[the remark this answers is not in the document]
samhed: ok, it's fine then

app/localization.js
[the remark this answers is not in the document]
samhed: In my mind they are fine if the condition and results are simple and you can keep it short.

juanjoDiaz:
I think I fixed all the comments.

Also there are some open questions:
* **Function normalization** Was it ok the way I normalized spacing around function declarations?
* **Increments** Currently we different increments randomly distributed accross the entire codebase. How should this go? Should we normalize on `i++`, `i += 1` or `++i`? I find the second option the clearest except for few exceptions like the encoding logic. But up to you.
* **Dangling coma** Currently we have dangling comas randomly distributed accross the entire codebase. How should this go? Dangling coma always or never? Most guidelines recomend it because it make git diffing better?
* **Destructuring** Which version of Chrome are we testing on? It seems to fail when using destructuring assignments which are supported since Chrome 49.
* **Polyfills** Probably not part of this PR. Currently we are always adding the Object.assign and CustomEvent polyfills to RFB. Shouldn't it be option and added to the UI as we do with the Promise polyfill? So the user can decide when to add it and when to not add it?

Anything else that I'm missing?

DirectXMan12:
>  **Function normalization** Was it ok the way I normalized spacing around function declarations?

I prefer `function (args)` and `function foo (args)` in all cases.  My only complaint with yours was that there seemed to (initially) be a difference between the way you treated anonymous functions vs named functions.

> **Increments** Currently we different increments randomly distributed accross the entire codebase. How should this go? Should we normalize on i++, i += 1 or ++i? I find the second option the clearest except for few exceptions like the encoding logic. But up to you.

I'd normalize on `i++`.  We should make sure we're not using `i++` or `++i` as an expression anywhere, but it's fine as a statement.

> **Dangling coma** Currently we have dangling comas randomly distributed accross the entire codebase. How should this go? Dangling coma always or never? Most guidelines recomend it because it make git diffing better?

Always use the dangling comma, except in literal JSON (i.e. not in a JS file).

> **Destructuring** Which version of Chrome are we testing on? It seems to fail when using destructuring assignments which are supported since Chrome 49.

A fairly old version.  Probably want to bump that.

>  **Polyfills** Probably not part of this PR. Currently we are always adding the Object.assign and CustomEvent polyfills to RFB. Shouldn't it be option and added to the UI as we do with the Promise polyfill? So the user can decide when to add it and when to not add it?

Probably should be an option, yes

DirectXMan12:
Will take a look at the changes again later this week.

core/rfb.js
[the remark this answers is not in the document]
CendioOssman: I don't like that we need to disable it in this case though. What is the eslint recommended way for an infinite loop?

app/localization.js
```diff
@@ -84,40 +79,32 @@ Localizer.prototype = {
...
     // Retrieve localised text
     get: function (id) {
-        if (typeof this.dictionary !== 'undefined' && this.dictionary[id]) {
-            return this.dictionary[id];
-        } else {
-            return id;
-        }
+        return (typeof this.dictionary !== 'undefined' && this.dictionary[id])
+            ? this.dictionary[id]
+            : id;
```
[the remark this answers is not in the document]
CendioOssman: Agreed. This construct should only be used for very minimal things.

app/localization.js
```diff
@@ -153,8 +143,8 @@ Localizer.prototype = {
                 }
             }
 
-            for (var i = 0;i < elem.childNodes.length;i++) {
-                var node = elem.childNodes[i];
+            for (let i = 0; i < elem.childes.length; i++) {
```
CendioOssman: What?

app/localization.js
CendioOssman: This is the let/const/var commit, or am I missing something?

app/ui.js
```diff
@@ -594,15 +592,15 @@ var UI = {
...
+            const dragThreshold = 10 * (window.devicePixelRatio || 1);
+            const dragDistance = Math.abs(ptr.clientY - UI.controlbarMouseDownClientY);
 
             if (dragDistance < dragThreshold) return;
 
             UI.controlbarDrag = true;
         }
 
-        var eventY = ptr.clientY - UI.controlbarMouseDownOffsetY;
+        const eventY = ptr.clientY - UI.controlbarMouseDownOffsetY;
```
CendioOssman: I'm not a huge fan of using `const` in cases like these, but I suppose I can overlook it as part of the conversion from `var` and we'll sort things out as we go in the future.

app/ui.js
```diff
@@ -1452,20 +1447,14 @@ var UI = {
...
 
-        var backspaces;
-        var inputs = newLen - oldLen;
-        if (inputs < 0) {
-            backspaces = -inputs;
-        } else {
-            backspaces = 0;
-        }
+        let inputs = inputs;
+        let backspaces = inputs ? -inputs : 0;
```
CendioOssman: Does this even execute at all?

core/base64.js
```diff
@@ -13,33 +13,30 @@ export default {
...
-        for (var i = 0; i < (length - 2); i += 3) {
-            result += toBase64Table[data[i] >> 2];
-            result += toBase64Table[((data[i] & 0x03) << 4) + (data[i + 1] >> 4)];
-            result += toBase64Table[((data[i + 1] & 0x0f) << 2) + (data[i + 2] >> 6)];
-            result += toBase64Table[data[i + 2] & 0x3f];
+        for (let i = 0; i < (length - 2); i += 3) {
+            result += this.toBase64Table[data[i] >> 2];
+            result += this.toBase64Table[((data[i] & 0x03) << 4) + (data[i + 1] >> 4)];
+            result += this.toBase64Table[((data[i + 1] & 0x0f) << 2) + (data[i + 2] >> 6)];
+            result += this.toBase64Table[data[i + 2] & 0x3f];
```
CendioOssman: Why a bunch of `this.` all of a sudden?

core/display.js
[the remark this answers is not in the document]
CendioOssman: I agree. If we're doing this then we should nuke every `var`.

karma.conf.js
```diff
@@ -101,6 +104,7 @@ module.exports = function(config) {
 
     babelPreprocessor: {
       options: {
+        presets: transpile ? ['es2015'] : [],
```
CendioOssman: If this is needed for Karma then it is also needed for the runtime transpiler, as well as `use_require.js`.

tests/playback.js
```diff
@@ -9,12 +9,12 @@ import * as Log from '../core/util/logging.js';
 import Base64 from '../core/base64.js';
 
 // Immediate polyfill
-if (setImmediate === undefined) {
-    var _immediateIdCounter = 1;
-    var _immediateFuncs = {};
+if (window.setImmediate === undefined) {
```
CendioOssman: This isn't related to const/let so it should be in some other commit.

karma.conf.js
CendioOssman: I'd like to see the transpiling changes as a separate commit.

CendioOssman:
I'm also not seeing updates to the eslint configuration. Shouldn't there be additions there to make sure we don't regress in this cleanups?

.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "parserOptions": {
+    "sourceType": "module"
+  },
+  "extends": "eslint:recommended"
```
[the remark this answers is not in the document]
juanjoDiaz: ESlint doesn't really fix anything.
It has a fix option but it rarely works. On this project ti just says that can't fix anything.

So what I did on this first commit is to run eslint with the `eslint:recommended` configuration and fix the reported issues manually.

`eslint:recommended` only include issues that will most probably break your code.

app/ui.js
```diff
@@ -510,7 +509,7 @@ var UI = {
         document.getElementById('noVNC_status').classList.remove("noVNC_open");
     },
 
-    activateControlbar: function(event) {
+    activateControlbar: function() {
```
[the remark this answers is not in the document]
juanjoDiaz: we can definitely turn it off if we want to keep these.

core/base64.js
juanjoDiaz: Before we did on top: `var toBase64Table = this.toBase64Table;`

Which is unnecessary. You can just call from this.

juanjoDiaz:
Github doesn't let me answer some of your comments.

`I don't like that we need to disable it in this case though. What is the eslint recommended way for an infinite loop?`
eslint recommend that you shouldn't have infinite loops. Which makes sense to me. An infinite loop is an exception no the normal way of using loops. That's why I add the comment to indicate the exception. Maybe it's not super-beautiful to have the comment there but it indicates that it's an exception in our normal rules.

`I'm also not seeing updates to the eslint configuration. Shouldn't there be additions there to make sure we don't regress in this cleanups?`

True I didn't push any rule because I wasn't sure of how you wanted the rules to be. I'l start pushing them.

I'l  start doing the commits again. This time one by one until you guys review them so I don't have to be going back and forth.

juanjoDiaz:
Ok. First commit re-pushed.

It simply adds eslint with the recommended config and fix the issues.

I've overriden the config a bit so we can have unused function arguments (as requested by @CendioOssman )

Once we agree that it looks good. I'll push the let/const commit.

juanjoDiaz:
Just pinging to see if anyone is still looking at this :)

CendioOssman:
Not right now, but it's on my todo list. :)

core/rfb.js
```diff
@@ -477,7 +474,7 @@ RFB.prototype = {
         this.focus();
     },
 
-    _windowResize: function (event) {
+    _windowResize: function (e) {
```
CendioOssman: Why the name change?

core/rfb.js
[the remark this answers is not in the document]
CendioOssman: @juanjoDiaz, did you check if there was a better way that would make lint happy?

tests/test.keyboard.js
```diff
@@ -279,7 +278,7 @@ describe('Key Event Handling', function() {
             });
             it('should ignore compositing code', function() {
                 var kbd = new Keyboard(document);
-                kbd.onkeyevent = function(keysym, code, down) {
+                kbd.onkeyevent = function(keysym, code) {
```
CendioOssman: These probably don't need to change now that you tweaked the settings for lint.

core/rfb.js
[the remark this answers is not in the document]
CendioOssman: there's apparently a `checkLoops` setting for exactly this. So I suggest turning on that option.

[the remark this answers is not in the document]
juanjoDiaz: Endless loops are dangerous. That's why there shouldn't be used unless strictly necessary.
That's why the `no-constant-condition` and `checkLoops` rule is enable by default. Disabling it will not indicate endless loop anywhere in the code.
That's why I think that using a comment is good to indicate that this particular endless loop is an exception and not the norm.

If you still disagree and prefer you to completely disable the rule. Let me know and I can do it.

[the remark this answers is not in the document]
CendioOssman: My vote is to disable this rule since I consider this particular pattern to be a common and useful one.

What does everyone else think?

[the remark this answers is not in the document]
samhed: Agreed, +1 for disabling the rule

juanjoDiaz:
Ok. All the comments have been fixed.
Please give me the green light to go with the next commit.

CendioOssman:
LGTM, please go ahead.

samhed:
The "eslint-disable no-fallthrough"-comments are a bit annoying in my eyes, but I can live with that. Other than that, it looks nice.

juanjoDiaz:
Ok. Prefering const/let over var is in.

There is an issue with firefox where is doesn't treat variables in for..in and for..of loops as being declared on every loop (as they are according the spec). That doesn't happen on my firefox so I assume that the reason is that the tests are done in a very old version of Firefox (I already noticed a similar thing caused by a very old version of chrome).

Options are:
* Bump the version of Firefox used for testing
* Transpile the code for Firefox
* Replace the for..in/for..of loops with imperative loops
* Add ugly comments (that you guys hate :)) before each for..in/for..of to allow using let instead of const.

What would you prefer? I'd rather bumping the testing browsers version.

Once we fix that and I get the greenlight. I'll push the arrow functions commit.

app/localization.js
samhed: This still has to be fixed!

juanjoDiaz: Fixed!

app/error-handler.js
```diff
@@ -2,6 +2,7 @@
 // native support in the browsers, so that our error handler
 // can catch script-loading errors.
 
+/* eslint-disable no-var */
```
CendioOssman: Even though this file cannot get the full ES6 treatment, I think `let` and `const` are fine. We only had problem with really old Safari in that regard.

.eslintrc
```diff
@@ -9,6 +9,8 @@
   "extends": "eslint:recommended",
   "rules": {
       "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
-      "no-constant-condition": ["error", { "checkLoops": false }]
+      "no-constant-condition": ["error", { "checkLoops": false }],
+      "no-var": "error",
+      "prefer-const": "error"
```
CendioOssman: As I mentioned earlier, I do not agree with this rule. IMO `const` should mark that things are conceptually constant, not just happened to never be reassigned.

app/localization.js
```diff
@@ -21,25 +21,21 @@ export function Localizer() {
...
-        for (var i = 0;i < userLanguages.length;i++) {
-            var userLang = userLanguages[i];
-            userLang = userLang.toLowerCase();
-            userLang = userLang.replace("_", "-");
-            userLang = userLang.split("-");
+        for (let i = 0;i < userLanguages.length;i++) {
+            const userLang = userLanguages[i]
+                .toLowerCase()
+                .replace("_", "-")
+                .split("-");
```
CendioOssman: Perhaps we should move this to a helper function instead? Avoids this churn, and makes the `const` declaration more clear.

app/localization.js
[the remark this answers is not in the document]
CendioOssman: My comment here stands. It is not required here to get `const` to work, so the existing code should be untouched.

app/localization.js
```diff
@@ -84,40 +79,32 @@ Localizer.prototype = {
...
                 node.data = str;
             }
 
             if (elem.hasAttribute("translate")) {
-                if (isAnyOf(elem.getAttribute("translate"), ["", "yes"])) {
-                    enabled = true;
-                } else if (isAnyOf(elem.getAttribute("translate"), ["no"])) {
-                    enabled = false;
-                }
+                enabled = isAnyOf(elem.getAttribute("translate"), ["", "yes"]);
```
CendioOssman: This is a behavioural change. Does this still follow the spec?

app/localization.js
```diff
@@ -140,7 +140,7 @@ Localizer.prototype = {
                 }
             }
 
-            for (let i = 0; i < elem.childes.length; i++) {
+            for (let i = 0; i < elem.childNodes.length; i++) {
```
CendioOssman: This belonged in the previous commit.

app/ui.js
```diff
@@ -1089,7 +1084,7 @@ var UI = {
     },
 
     disconnectFinished: function (e) {
-        let wasConnected = UI.connected;
+        const wasConnected = UI.connected;
```
CendioOssman: Unnecessary change.

app/ui.js
CendioOssman: Alright, I'm going to stop reviewing here. You need to check the existing feedback we've given you and fix those things.

.eslintrc
juanjoDiaz: That's impossible to enforce.
Should it be just everything let then?

app/localization.js
juanjoDiaz: I can't see any difference between
```
if (var1 = true) {
  var2 = true
} else {
  var2 = false
}
```

or `var1 = var2`

unless you want to be really strict with typing, in which case I'd still do `var1 = !!var2`

juanjoDiaz:
I've checked all the comments and reverted the changes except for the const/let replacements.

Open questions are:
* *const vs let* It is a common best practice to use const whenever possible. Makes it clear when a variable never changes and makes code more solid. Also, it can be enforced by eslint. I guess we could also. I think that we could also enforce "always let", but no other arbitrary system is enforceable AFAIK.
* *test browsers versions* For the tests to pass, firefox version needs to be bumped (and I'd bump all other browser's versions as well). Other alternatives are listed in my previous comment.

anything else missing?

juanjoDiaz:
ping ^

app/ui.js
JakeDoe: const would be better, then you are using ```const ``` you tell other developers what it is an immutable entity.

samhed:
> const vs let It is a common best practice to use const whenever possible. Makes it clear when a variable never changes and makes code more solid. Also, it can be enforced by eslint. I guess we could also. I think that we could also enforce "always let", but no other arbitrary system is enforceable AFAIK.

I'm fine with enforcing this.

> test browsers versions For the tests to pass, firefox version needs to be bumped (and I'd bump all other browser's versions as well). Other alternatives are listed in my previous comment.

Did you check if it actually is possible? See my previous comment on this subject

CendioOssman:
Sorry, things have been hectic here. I'll check the latest things now.

CendioOssman:
> * const vs let It is a common best practice to use const whenever possible. Makes it clear when a variable never changes and makes code more solid. Also, it can be enforced by eslint. I guess we could also. I think that we could also enforce "always let", but no other arbitrary system is enforceable AFAIK.

We should only enforce a ban on `var`, nothing else at this point. `const` vs `let` is something a human needs to decide.
> * test browsers versions For the tests to pass, firefox version needs to be bumped (and I'd bump all other browser's versions as well). Other alternatives are listed in my previous comment.

What versions? Does Firefox ESR still work?

app/localization.js
```diff
@@ -21,25 +21,24 @@ export function Localizer() {
...
-
         this.language = 'en'; // Default: US English
 
         /*
          * Navigator.languages only available in Chrome (32+) and FireFox (32+)
          * Fall back to navigator.language for other browsers
          */
-        if (typeof window.navigator.languages == 'object') {
+        let userLanguages;
+         if (typeof window.navigator.languages == 'object') {
```
CendioOssman: Something is off with the indentation here.

app/ui.js
```diff
@@ -1652,7 +1645,7 @@ var UI = {
 };
 
 // Set up translations
-var LINGUAS = ["de", "el", "es", "nl", "pl", "sv", "tr", "zh"];
+const LINGUAS = ["de", "el", "es", "nl", "pl", "sv", "zh"];
```
CendioOssman: This is still broken.

app/webutil.js
```diff
@@ -211,18 +221,16 @@ export function injectParamIfMissing (path, param, value) {
...
     // NB: IE11 doesn't support JSON as a responseType
-    var req = new XMLHttpRequest();
+    const req = new XMLHttpRequest();
     req.open('GET', path);
 
     req.onload = function () {
         if (req.status === 200) {
             try {
-                var resObj = JSON.parse(req.responseText);
+                resolve(JSON.parse(req.responseText));
```
CendioOssman: This is a change of behaviour as it will grab any exceptions from the resolve handler as well. Please keep it the previous way.

core/base64.js
CendioOssman: The extra indirection might cause performance hiccups though. But we'll have to see during testing.

core/des.js
```diff
@@ -79,80 +79,76 @@ export default function DES(passwd) {
     "use strict";
 
     // Tables, permutations, S-boxes, etc.
-    var PC2 = [13,16,10,23, 0, 4, 2,27,14, 5,20, 9,22,18,11, 3,
+    const PC2 = [13,16,10,23, 0, 4, 2,27,14, 5,20, 9,22,18,11, 3,
                25, 7,15, 6,26,19,12, 1,40,51,30,36,46,54,29,39,
```
CendioOssman: The alignment of these and those below are off now.

core/input/keyboard.js
```diff
@@ -306,25 +304,23 @@ Keyboard.prototype = {
             return;
         }
 
-        let target = this._target;
-        let downList = this._keyDownList;
         ['AltLeft', 'AltRight'].forEach(function (code) {
-            if (!(code in downList)) {
+            if (!(code in this._keyDownList)) {
```
CendioOssman: `this` points to something different here, so this does not work.

core/input/mouse.js
```diff
@@ -225,6 +225,7 @@ Mouse.prototype = {
         } else {
             x = e.clientX - bounds.left;
         }
+        let y;
```
CendioOssman: If we are just declaring it then let's keep `x` and `y` together like before.

core/rfb.js
```diff
@@ -967,13 +960,13 @@ RFB.prototype = {
         if (!this._rfb_credentials.username ||
             !this._rfb_credentials.password ||
             !this._rfb_credentials.target) {
-            var event = new CustomEvent("credentialsrequired",
-                                        { detail: { types: ["username", "password", "target"] } });
-            this.dispatchEvent(event);
+            this.dispatchEvent(new CustomEvent(
+                "credentialsrequired",
```
CendioOssman: These line breaks seem unnecessary. Are the lines really that long?

karma.conf.js
```diff
@@ -1,29 +1,32 @@
 // Karma configuration
 
 module.exports = function(config) {
-  var customLaunchers = {};
-  var browsers = [];
-  var useSauce = false;
+  const customLaunchers = {};
+  let browsers = [];
+  let useSauce = false;
+  let transpile = false;
```
CendioOssman: We are still transpiling modules, so perhaps this should have a more specific name for clarity.

CendioOssman:
IE 11 does not currently work either with Karma nor by accessing the page directly. It complains about "Const must be initialized", so the conversion doesn't seem to be working properly.

app/webutil.js
juanjoDiaz: `resolve(JSON.parse(req.responseText));` is functionally equivalent to:
```
const json = JSON.parse(req.responseText);
resolve(json);
```
since the parse method is executed first and it's result passed to the resolve method call.
If `JSON.parse` throws an exception, the resolve method is never called and the try-catch with capture the exception.

Do you still prefer the second version with the extra variable?

juanjoDiaz:
> Getting there. :)
> Just some minor things. Also, the transpiling changes should probably be an early commit so that each commit can be tested on every browser.
Good to hear that this is getting somewhere :)

The transpiling are included on the 

> The extra indirection might cause performance hiccups though. But we'll have to see during testing.
JS engines are pretty smart nowadays. So I don't think that it really makes a difference. (See https://jsperf.com/scope-resolution-cached-not-cached/3)

> IE 11 does not currently work either with Karma nor by accessing the page directly. It complains about "Const must be initialized", so the conversion doesn't seem to be working properly.
How can I reproduce the IE11 issue?
I can't find it in Travis and I can't find any unitialize const in the code (`const MY_VAR;`)
Does it tell you the file or the line where the error comes from?

> I'd prefer bumping the version used in the travis tests, if it's possible. It looks like firefox 45 was used on the Linux tests, but looking at our travis configuration, we aren't specifying the version. If we are unlucky, firefox 45 is the latest that is available here.
You can specify the travis version in travis using https://docs.travis-ci.com/user/firefox
However, it seems to me that the browser version is not given by Travis but by Sauce labs. Isn't that the case, @samhed ?
Unfortunatelly, it looks like Sauce labs only support Firefox up to 45 on linux (https://saucelabs.com/platforms) even though it supports much newer versions in all other OSs.
So for now, I'll add transpilation for firefox.
In the future, I recommend that noVNC uses `browser`, `module` & `main` fields in the package.json so the user can decide to use the transpile version or the original. I can do a PR for that once we get this whole ES6 upgrade done :)

core/input/keyboard.js
juanjoDiaz: True. That won't be a problem when moving to arrow functions but it's now.

core/rfb.js
samhed: I don't think we need a consistent line-break pattern for each dispatchEvent call. It's fine if the line-breaks are on a case-to-case basis.

samhed:
> However, it seems to me that the browser version is not given by Travis but by Sauce labs. Isn't that the case, @samhed ?

It seems like that indeed.

> Unfortunatelly, it looks like Sauce labs only support Firefox up to 45 on linux (https://saucelabs.com/platforms) even though it supports much newer versions in all other OSs. So for now, I'll add transpilation for firefox.

Ok, that sucks. Perhaps we should try to contact them and ask what's going on. @DirectXMan12 what do you think?

core/rfb.js
juanjoDiaz: Most dispatchEvents are the same length.
We just had very long lines and not-so-long ones.

I introduce the change because some of you guys asked me to put the CustomEvent inside of the dispatchEvent in all cases (before it was mixed) and that made most dispatch to be on the long size.

I think that it looks cleaner and better as it is now tbh.
Would you rather me to rever all of them?

juanjoDiaz:
Ok.

There are a couple minor questions to @CendioOssman and @samhed .

Apart from that, I'm waiting for the LGTM and I'll push the Classes and arrow functions commit :)

app/webutil.js
samhed: As @CendioOssman is hinting, we should try avoid any changes in behavior. Imagine troubleshooting or changing something related to this code in the future, and jumping between versions and checking blame, this commit could easily become confusing.

juanjoDiaz:
I've reverted the `resolve(JSON.parse(req.responseText));` statement. Still, change because it was `var` and now it's not. But reduce the change to a minimum.

Anything else pending?

samhed:
You haven't fixed all the things we asked you to change.

CendioOssman:
> How can I reproduce the IE11 issue?
> I can't find it in Travis and I can't find any unitialize const in the code (const MY_VAR;)

I simply opened `vnc.html` or the karma page in IE on a Windows 2008 R2 machine here. Do you have access to a Windows machine?

> Does it tell you the file or the line where the error comes from?

For `vnc.html`:
```
Const must be initialized
  Instantiating http://ossman.lkpg.cendio.se:6080/core/rfb.js
  Loading http://ossman.lkpg.cendio.se:6080/app/ui.js
  Loading http://ossman.lkpg.cendio.se:6080/app/ui.js
```

And for karma:
```
IE 11.0.0 (Windows 7 0.0.0) ERROR
  Const must be initialized
  at http://ossman.lkpg.cendio.se:9876tests/test.keyboard.js:43
```

CendioOssman:
> > Unfortunatelly, it looks like Sauce labs only support Firefox up to 45 on linux (https://saucelabs.com/platforms) even though it supports much newer versions in all other OSs. So for now, I'll add transpilation for firefox.
> 
> Ok, that sucks. Perhaps we should try to contact them and ask what's going on. @DirectXMan12 what do you think?

It looks like both Chrome and Firefox are ancient. I don't think we have such a pressing need to test on Linux specifically? So I suggesting removing the Linux tests with a FIXME until Sauce gets their act together.

.eslintrc
CendioOssman: Again, `no-var` is fine and good but `prefer-const` should not be enabled.

app/webutil.js
```diff
@@ -211,18 +221,17 @@ export function injectParamIfMissing (path, param, value) {
...
-    var req = new XMLHttpRequest();
+    const req = new XMLHttpRequest();
     req.open('GET', path);
 
     req.onload = function () {
         if (req.status === 200) {
             try {
-                var resObj = JSON.parse(req.responseText);
+                const resObj = JSON.parse(req.responseText);
+                resolve(resObj);
```
CendioOssman: Still wrong. The `resolve()` call needs to stay outside of the `try {}`.

karma.conf.js
```diff
@@ -1,29 +1,28 @@
...
 
 module.exports = function(config) {
-  var customLaunchers = {};
-  var browsers = [];
-  var useSauce = false;
+  const customLaunchers = {};
+  let browsers = [];
+  let useSauce = false;
+  // New firefox versions don't need transpilation but Sauce Labs only support Firefox 45
+  let transpileToES5 = ['internet explorer', 'firefox'].includes(process.env.TEST_BROWSER_NAME);
```
CendioOssman: I think it's more important to test a current Firefox without transpilation than to test that old version. So let's comment out the Linux tests with a FIXME that we're waiting for Sauce to upgrade.

samhed:
We commented on `dispatchEvent` indentation, but other than that I think it looks good.

samhed:
> It looks like both Chrome and Firefox are ancient. I don't think we have such a pressing need to test on Linux specifically? So I suggesting removing the Linux tests with a FIXME until Sauce gets their act together.

+1

juanjoDiaz:
Sorry, @CendioOssman , I missed the `Again, no-var is fine and good but prefer-const should not be enabled.`
@samhed said 12 days ago that he was happy with enforcing that rule when I explained the reason.

> const vs let It is a common best practice to use const whenever possible. Makes it clear when a variable never changes and makes the code more solid. Also, it can be enforced by eslint. We could enforce "always let", but no other arbitrary system is enforceable AFAIK.

 So now I'm not completely clear about what to do. Just have a chat between you guys and let me know what you want me to do. Maybe @DirectXMan12  or others have also an opinion :)

Btw, is there any reason that you are using a custom module `use_require.js` for transpilation instead of just using, for example, rollup (which is already a dependency anyway)?

samhed:
Regarding `prefer-const`, I think @CendioOssman is right here. I agree that there are situations where it's more important to signal the intent rather than the actual reality.

CendioOssman:
> Regarding the issues mentioned by @CendioOssman , I've realized that I've added the transpilation for the karma tests but not for the actual release code.
> 
> I'm not sure of how are you currently building and releasing so I'm not 100% sure of the best way here so any help is welcome.
> My understanding is that you use use_require.js so I just have to add the preset to https://github.com/novnc/noVNC/blob/master/utils/use_require.js#L152
> 
> Is that correct?

Sounds correct, yes. You also need to have a look in `vendor/browser-es-module-loader/src/babel-worker.js`.

We do transpiling in three places:

 1. `babel-worker.js`: This is done "live" in the browser so that noVNC works out-of-box
 2. `karma.conf.js`: Because karma doesn't support ES6 modules yet
 3. `use_require.js`: For people that want to use tools that require CommonJS, or avoid the slow conversion in the browser

> Btw, is there any reason that you are using a custom module use_require.js for transpilation instead of just using, for example, rollup (which is already a dependency anyway)?

Might just be for historical reasons. But can rollup do everything we need? E.g. there is some manipulation of `vnc.html` in there as well.

juanjoDiaz:
Alright!

I've removed the `prefer-const` rule.
And I added the transpilation in the 3 places mentioned by @CendioOssman (let's leave the build simplification for another PR maybe 🙂)

Anything else left?

samhed:
Did you manage to reproduce the IE 11 issue?

juanjoDiaz:
I don't have access to any machine with IE 11.

The problem described was just that I wasn't transpiling and `const` is not 100% supported as it should in IE11.

However, all the tests pass meaning that the problem is gone with transpilation and I added the same transpilation in the other 2 places.

CendioOssman:
Alright, so IE is still broken. From what I can tell the construct it doesn't support is `for (const foo in ...)`. It however works fine with `let`. So I think we need to be careful with those for now.

I don't understand why the tests aren't failing in the Travis run though. One guess is that IE 11 on Windows 10 is really Edge in disguise. Could you try adding Windows 7 as a test OS to see if that breaks things?

CendioOssman:
We should also really have Edge in the tests, so feel free to add that as well. :)

CendioOssman:
`forEach()` unfortunately requires creating a new function scope, which is very annoying. So right now I think `for..in` is the lesser evil. We would just have to use `let` rather than `const` to keep IE happy.

juanjoDiaz:
Alright. Rolled back to for..in and for..of

I still think that's the most unsafe way, but oh well... we have already removed the const rule anyway...

samhed:
Looks good now.

I'm having major issues displaying this pull-request now.. Doesn't matter if I use Chrome or Firefox or which machine I'm using, I guess it's a github problem.

CendioOssman:
Alright, but we still need to use `let` in `for..in` constructs for IE to work. I've tweaked the travis settings on `master` to make sure this is properly tested.

CendioOssman:
Seems to be these:

```
$ git grep for.*\(const.*in
core/input/keyboard.js:        for (const code in this._keyDownList) {
core/rfb.js:        for (const authType in clientSupportedTypes) {
core/util/polyfill.js:                    for (const nextKey in nextSource) {
tests/fake.websocket.js:        for (const prop in props) {
tests/test.keyboard.js:        for (const key in KeyboardEventInit) {
tests/test.mouse.js:        for (const key in MouseEventInit) {
tests/test.rfb.js:                for (const opt in opts) {
```

juanjoDiaz:
It seems like I forgot to push :/

Should be good now :)

CendioOssman:
Go ahead.

juanjoDiaz:
Any idea of what might be wrong?

It seems that IE11 doesn't like something. However, the error is not very meaningful and points to `core/util/polyfill.js:20` which hasn't been changed and `vendor/pako/lib/zlib/inflate.js` which is a vendor library and hasn't been changed.

@CendioOssman can you check if you can run it locally and get a more meaningful error?

CendioOssman:
The polyfill is probably the only real error. It's inside the `Object.assign()` polyfill, so I guess that is being called incorrectly somewhere.

The inflate error is just a bogus error cause by karma shutting down on the initial error, which then means IE fails to load the remaining js files.

CendioOssman:
Could you split this in to two commits? One of the arrow functions and one for the classes. The first should require a fairly mild review, but the latter is a bigger change.

And you need to keep ES6 stuff out of `error-handler.js` as it is loaded without transpiling.

juanjoDiaz:
After rebasing, something weird seems to be going on with Edge. Any idea?

core/rfb.js
```diff
@@ -953,16 +959,15 @@ RFB.prototype = {
...
-    _negotiate_xvp_auth: function () {
+    _negotiate_xvp_auth() {
         if (!this._rfb_credentials.username ||
             !this._rfb_credentials.password ||
             !this._rfb_credentials.target) {
-            this.dispatchEvent(new CustomEvent(
-                "credentialsrequired",
-                { detail: { types: ["username", "password", "target"] } }));
+            this.dispatchEvent(new CustomEvent("credentialsrequired",
+                                               { detail: { types: ["username", "password", "target"] } }));
```
CendioOssman: Unrelated change. (there are a few like this further down)

core/rfb.js
```diff
@@ -1675,7 +1680,7 @@ RFB.messages = {
         buff[offset + 2] = 0; // padding
         buff[offset + 3] = 0; // padding
 
-        const length = text.length;
+        let length = text.length;
```
CendioOssman: Unrelated

app/ui.js
```diff
@@ -383,7 +381,7 @@ var UI = {
         document.documentElement.classList.remove("noVNC_disconnecting");
         document.documentElement.classList.remove("noVNC_reconnecting");
 
-        let transition_elem = document.getElementById("noVNC_transition_text");
+        const transition_elem = document.getElementById("noVNC_transition_text");
```
CendioOssman: There's a whole bunch of these unnecessary changes from `let` to `const`.

karma.conf.js
```diff
@@ -99,6 +99,7 @@ module.exports = function(config) {
       'tests/test.*.js': ['babel'],
       'tests/fake.*.js': ['babel'],
       'tests/assertions.js': ['babel'],
+      'tests/payback*.js': ['babel'],
```
CendioOssman: Typo. But these aren't tested, so no need to convert them.

tests/test.rfb.js
```diff
@@ -1239,19 +1239,18 @@ describe('Remote Frame Buffer Protocol Client', function() {
...
 
                 function send_num_str_pairs(pairs, client) {
-                    const pairs_len = pairs.length;
                     const data = [];
-                    push32(data, pairs_len);
+                    push32(data, pairs.length);
 
-                    for (let i = 0; i < pairs_len; i++) {
-                        push32(data, pairs[i][0]);
+                    pairs.forEach((pair) => {
```
CendioOssman: Unrelated.

tests/test.rfb.js
```diff
@@ -1625,22 +1624,25 @@ describe('Remote Frame Buffer Protocol Client', function() {
...
 
                 // just enough to set FBU.rects
                 client._sock._websocket._receive_data(new Uint8Array([0, 0, 0, 3]));
                 expect(client._sock._websocket._get_sent_data()).to.have.length(0);
 
-                client._framebufferUpdate = function () { this._sock.rQskip8(); return true; };  // we magically have enough data
+                client._framebufferUpdate = function () {
+                    this._sock.rQskip8();
+                    return true;
+                };  // we magically have enough data
```
CendioOssman: Unrelated.

utils/genkeysymdef.js
```diff
@@ -108,7 +108,7 @@ out +=
 "};\n" +
 "\n" +
 "export default {\n" +
-"    lookup : function(u) {\n" +
+"    lookup(u) {\n" +
```
CendioOssman: Unrelated.

utils/use_require.js
```diff
@@ -39,15 +39,14 @@ const no_transform_files = new Set([
...
 
 // util.promisify requires Node.js 8.x, so we have our own
 function promisify(original) {
-    return function () {
-        const obj = this;
+    return () => {
         const args = Array.prototype.slice.call(arguments);
         return new Promise((resolve, reject) => {
-            original.apply(obj, args.concat((err, value) => {
+            original.apply(this, args.concat((err, value) => {
```
CendioOssman: This does not have the same behaviour as the previous code. I don't think you can use an arrow function here.

utils/use_require.js
```diff
@@ -70,29 +69,26 @@ const babelTransformFile = promisify(babel.transformFile);
 
 // walkDir *recursively* walks directories trees,
 // calling the callback for all normal files found.
-const walkDir = function (base_path, cb, filter) {
+function walkDir(base_path, cb, filter) {
     return readdir(base_path)
-    .then(files => {
+    .then((files) => {
```
CendioOssman: Not needed, right?

utils/use_require.js
```diff
@@ -157,9 +153,7 @@ const make_lib_files = function (import_format, source_maps, with_app_dir, only_
     });
 
     // No point in duplicate files without the app, so force only converted files
-    if (!with_app_dir) {
-        only_legacy = true;
-    }
+    const only_legacy = !with_app_dir;
```
CendioOssman: This is not the same thing.

.eslintrc
```diff
@@ -8,8 +8,13 @@
   },
   "extends": "eslint:recommended",
   "rules": {
-      "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
-      "no-constant-condition": ["error", { "checkLoops": false }],
-      "no-var": "error"
+    "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
```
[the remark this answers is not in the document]
juanjoDiaz: The whole file is indented to 2 spaces except for this 3 lines.

I'll re-indent the whole file to 4 spaces

app/localization.js
```diff
@@ -78,33 +78,32 @@ Localizer.prototype = {
...
-        const self = this;
+    translateDOM() {
         function process(elem, enabled) {
             function isAnyOf(searchElement, items) {
                 return items.indexOf(searchElement) !== -1;
             }
 
             function translateAttribute(elem, attr) {
-                const str = self.get(elem.getAttribute(attr));
+                const str = this.get(elem.getAttribute(attr));
```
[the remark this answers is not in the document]
juanjoDiaz: Why?

It was doing `const self = this;` at the beginning of the function but we are not changing the bindings or calling the function in any scope so `this` will never change.

core/base64.js
```diff
@@ -11,8 +11,7 @@ export default {
     toBase64Table : 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/='.split(''),
     base64Pad     : '=',
 
-    encode: function (data) {
-        "use strict";
+    encode(data) {
```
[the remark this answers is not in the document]
juanjoDiaz: Yeah but strict mode is automatically enabled for all ES6 modules and we are in one here, aren't we?

http://www.ecma-international.org/ecma-262/6.0/#sec-strict-mode-code

app/ui.js
juanjoDiaz: I thought that the let/const commit was approved already... I haven't touched that

We already discussed the let vs const stuff above. The const enforcing is removed.

Do you want me now to go back and change the const to let?

utils/use_require.js
juanjoDiaz: Why not?

As in the other comment above, `const obj = this;` and we are not doing any binding or calling the function on any other object so obj will always be equals to this.

Also, what do you mean about not being able to use an arrow function?

utils/use_require.js
juanjoDiaz: It's the convention that parenthesis are skip on 1-lines but used when using braces.

`myArg => myFunc(myArgs)`

vs

```
(myArg) => {
  return myFunc(myArgs);
}
```

Would you prefer something else?

juanjoDiaz:
I've fixed that issues that you commented and were clear issues and I've commented on the ones that I think that is correct.

Let's agree on what to do about the latter and try to get this merged. :)

I won't be doing any more changes as part of this PR. If I decide to do anything else (ES6 related or not) I'll do it as separate smaller PRs.

juanjoDiaz:
@CendioOssman can you clarify why " transpiling of classes isn't working"??

I run `npm run prepare` and the transpiled code in the lib folder doesn't contain classes at all.

CendioOssman:
Unfortunately github's system are crashing when trying to load this PR. 
I've filed a support request, but I'm afraid we cannot continue until 
they've resolved the issue. :/


-- 
Pierre Ossman           Software Development
Cendio AB               https://cendio.com
Teknikringen 8          https://twitter.com/ThinLinc
583 30 Linköping        https://facebook.com/ThinLinc
Phone: +46-13-214600    https://plus.google.com/+CendioThinLinc

A: Because it messes up the order in which people normally read text.
Q: Why is top-posting such a bad thing?

juanjoDiaz:
@CendioOssman Would it be a solution to revert this to the `const/let` commit that we were all happy with, get that merged and then submit I can submit a separate PR for `classes/arrow functions`?

Otherwise, this might take forever, more conflicts will appear, etc.

samhed:
Den Sat, 19 May 2018 13:57:22 -0700
skrev Re: [novnc/noVNC] Add eslint and update noVNC to ES6 (#1013):

> @CendioOssman Would it be a solution to revert this to the
> `const/let` commit that we were all happy with, get that merged and
> then submit I can submit a separate PR for `classes/arrow functions`?
> 
> Otherwise, this might take forever, more conflicts will appear, etc.

Sounds good to me.

-- 
Samuel Mannehed         Software Development
Cendio AB               https://cendio.com
Teknikringen 8          https://twitter.com/ThinLinc
583 30 Linköping        https://facebook.com/ThinLinc
Phone: +46-13-214600    https://plus.google.com/+CendioThinLinc

juanjoDiaz:
Ok. Now we just have the 3 commits that were already approved. Let's get those in and I'll do a separate PR for classes and arrow functions.

Btw, I can access the PR just fine and I don't think that we are too far from each other... No clue of what might be the problem that you are having...

juanjoDiaz:
The fail test is Sauce lab being flaky as usual. the changes are exactly the same as have been pushed a million times before and passed :)

PR #1016, opened 2018-02-05
Add Spanish translation

Kirbo92:
Add Spanish translation
captura de pantalla de 2018-02-05 08-48-06


juanjoDiaz:
Just for my own knowledge. Aren't this JSON files added by the maintainers right before a release?
That's what I understood from the documentation.

Kirbo92:
Yes, but if you look in app / locale, the Spanish language is missing, and when you enter the client with the browser in Spanish, the error jumps.

CendioOssman:
The idea was to add the JSON whenever we got a new `.po`, but I forgot to do so here. The bigger issue though is the crash. If the file is missing then it should have just continued without a translation.

CendioOssman:
Should be fixed in 7c332ad9303d209dbb32840f5be6eb3de4c8c67e and 37b4d13db81e0e80e117c07b86ff98714c7b6b1a.

PR #1021, opened 2018-02-18
Add support to token

linnil1:
Hi.
I found that noVNC can redirect to any machine by using token with  [websockify](https://github.com/novnc/websockify/issues/3). ([wiki](https://github.com/novnc/websockify/wiki/Token-based-target-selection))

However, it doesn't work on recent version.
After reading code in `app/ui.js`, I found that `path` didn't include query section in URL.
So, just add it easily by directly add  `document.location.search`

Also, the path is always point to `ws://ip:port/websocky`, which make bug when I deploy it, such as I add prefixStrip in the path like `127.0.0.1:8080/mypath/vnc.html`.
The solution is to add `window.location.pathname` in path.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1013,7 +1013,7 @@ var UI = {
         if(port) {
             url += ':' + port;
         }
-        url += '/' + path;
+        url += '/' + path + document.location.search;
 
         UI.rfb = new RFB(document.getElementById('noVNC_container'), url,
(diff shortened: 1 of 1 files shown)


CendioOssman:
noVNC has the setting `path` that allows you to handle both of these things. E.g. `http://example.com/mypath/vnc.html?path=%2Fmypath%2Fwebsockify%3Ftoken%3Dabc123`

Issue #1023, opened 2018-02-21
graphic glitch for noVNC


leslie-qiwa:
We're using noVNC in our application. Sometimes, we found some [graphic glitch](https://ibb.co/enrE3H) which doesn't have in native VNC player. Does anyone have any clue on it?

![screen capture](http://image.ibb.co/iQdu3H/Screen_Shot_2018_02_21_at_2_13_46_PM.png)

samhed:
What version of noVNC are you using?

leslie-qiwa:
thanks for reply. It is from master : a4d51bd22048df40963c999903e5156f343df293

CendioOssman:
Which browser and VNC server?

leslie-qiwa:
Safari - Version 11.0.2 (13604.4.7.1.3)
Firefox - 57.0.2 
Chrome - Version 64.0.3282.167 (Official Build) (64-bit)

We are using this [vncproxy](https://github.com/amitbet/vncproxy)

DirectXMan12:
Can you send us a recording with the issue?

leslie-qiwa:
Thanks for prompt reply. I'm trying to identify if problem is at [vncproxy](https://github.com/amitbet/vncproxy) side. Once it is done, will update you.

leslie-qiwa:
Just one correction, we are using https://github.com/libvirt/libvirt-console-proxy as proxy. 

For more information. Our server is ESX. Per this issue: https://github.com/novnc/noVNC/issues/44, seems like ESX VNC is not officially supported. It works in our side if using above proxy, then we saw above glitch sometimes.

leslie-qiwa:
After more investigation, I found out this issues is at our implementation. So I'll just close it. Seems like VMWare ESXi uses tight_png encoding, so it would be great if https://github.com/novnc/noVNC/issues/51 can be solved.

CendioOssman:
#51 is already solved, so I'm not sure what you are waiting for in that case? And #44 is about extra stuff, not the basic things like working image.

leslie-qiwa:
Thanks for the confirmation. Seems like `function encodingName(num)` at `core/encodings.js` need add support for `pseudoEncodingTightPNG` as well.

It is interesting that above graphic glitch problem can happen if I just change `encodingTight` to -260. But once I enable `pseudoEncodingTightPNG`, it will display correctly.

leslie-qiwa:
#1026 is created. It works in our ESX environment.

Issue #25, opened 2010-09-01
Consider encodings using zlib (zlib, tight, ZRLE)


kanaka:
Standard tight encoding and ZRLE will require zlib decompression in Javascript. This will probably be pretty slow (based on some initial testing), but having them as (non-default) options might be worthwhile in some cases.

leslie-qiwa:
@kanaka did you merge tight_png part? 
@quicklyfrozen seems like rfb.js has changed a lot. so would you mind to update your tight_png implementation with upstream?

Background, VMware ESXi VNC uses tight_png encoding. It can decode if forcing to use old tight, it can work, but have graphic glitches sometimes. So I'm thinking it would be great if someone can add the support.

CendioOssman:
TightPNG has been in noVNC since c0c20581f5598a7e0b864d20591e7f65f17c3445 and the protocol has not changed since then AFAIK. There might be bugs though since very few servers use that encoding.

kanaka:
@leslie-qiwa Can you point to any documentation showing that VMware ESXi supports tightPNG? I'm not able to find anything with some quick googling but I would be interested to see some evidence if you have it. I developed tightPNG (along with Corentin Chary) and last I knew it was only support by QEMU and libvncserver.

leslie-qiwa:
No document shows that. I found it because noVNC reports error "unknown encoding -260". 

I use https://github.com/amitbet/vncproxy to dump all messages. It shows ESXi returns both Tight and TightPNG.

Please refer this issue https://github.com/novnc/noVNC/issues/1023

leslie-qiwa:
https://github.com/novnc/noVNC/pull/1026 is created. It works in our ESX environment.

CendioOssman:
Ah, I see the problem. I've opened #1027 so we can handle this discussion better.

Neustradamus:
Note: ZRLE support has been added:
- https://github.com/novnc/noVNC/issues/545
- https://github.com/novnc/noVNC/pull/1589

PR #1026, opened 2018-02-23
add encoding support for TightPNG

leslie-qiwa:
Refer discussion on https://github.com/novnc/noVNC/issues/25, https://github.com/novnc/noVNC/issues/1023


CendioOssman:
Looks good so far, but I think we might want to revert more of 3e8b26ab58e7ef7457a324304306981d1ab0f528 so that we get the checks for zlib based filters back.

PR #1032, opened 2018-03-01
Render cursor manually when needed

### LINKED ISSUE #572: Local cursor not supported in IE or Edge
Normally we use a data uri to store the cursor, we then use this URI as a value for style.cursor on the canvas. However, this isn't supported in any version of Internet Explorer or Microsoft Edge.

http://caniuse.com/#search=cursor
http://caniuse.com/#feat=datauri

Can anyone think of another way to support local cursor in noVNC?

CendioOssman:
Render the cursor on top of the canvas when we cannot rely on the browser to handle the cursor. This means that local cursor can now be used for touch devices and on IE/Edge.

Fixes #572.

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -498,18 +498,6 @@ Display.prototype = {
         this._damage(x, y, img.width, img.height);
     },
 
-    changeCursor: function (pixels, mask, hotx, hoty, w, h) {
-        Display.changeCursor(this._target, pixels, mask, hotx, hoty, w, h);
-    },
-
-    defaultCursor: function () {
-        this._target.style.cursor = "default";
-    },
-
-    disableLocalCursor: function () {
-        this._target.style.cursor = "none";
-    },
-
     autoscale: function (containerWidth, containerHeight) {
         var vp = this._viewportLoc;
         var targetAspectRatio = containerWidth / containerHeight;
@@ -654,45 +642,4 @@ Display.prototype = {
             this.onflush();
         }
     },
-};
-
-// Class Methods
-Display.changeCursor = function (target, pixels, mask, hotx, hoty, w, h) {
-    if ((w === 0) || (h === 0)) {
-        target.style.cursor = 'none';
-        return;
-    }
-
-    var cur = []
-    var y, x;
-    for (y = 0; y < h; y++) {
-        for (x = 0; x < w; x++) {
-            var idx = y * Math.ceil(w / 8) + Math.floor(x / 8);
-            var alpha = (mask[idx] << (x % 8)) & 0x80 ? 255 : 0;
-            idx = ((w * y) + x) * 4;
-            cur.push(pixels[idx + 2]); // red
-            cur.push(pixels[idx + 1]); // green
-            cur.push(pixels[idx]);     // blue
-            cur.push(alpha);           // alpha
-        }
-    }
-
-    var canvas = document.createElement('canvas');
-    var ctx = canvas.getContext('2d');
-
-    canvas.width = w;
-    canvas.height = h;
-
-    var img;
-    if (SUPPORTS_IMAGEDATA_CONSTRUCTOR) {
-        img = new ImageData(new Uint8ClampedArray(cur), w, h);
-    } else {
-        img = ctx.createImageData(w, h);
-        img.data.set(new Uint8ClampedArray(cur));
-    }
-    ctx.clearRect(0, 0, w, h);
-    ctx.putImageData(img, 0, 0);
-
-    var url = canvas.toDataURL();
-    target.style.cursor = 'url(' + url + ')' + hotx + ' ' + hoty + ', default';
 };
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -17,6 +17,7 @@ import EventTargetMixin from './util/eventtarget.js';
 import Display from "./display.js";
 import Keyboard from "./input/keyboard.js";
 import Mouse from "./input/mouse.js";
+import Cursor from "./util/cursor.js";
 import Websock from "./websock.js";
 import DES from "./des.js";
 import KeyTable from "./input/keysym.js";
@@ -163,6 +164,8 @@ export default function RFB(target, url, options) {
     this._canvas.height = 0;
     this._canvas.tabIndex = -1;
     this._screen.appendChild(this._canvas);
+
+    this._cursor = new Cursor();
 
     // populate encHandlers with bound versions
     this._encHandlers[encodings.encodingRaw] = RFB.encodingHandlers.RAW.bind(this);
@@ -413,6 +416,8 @@ RFB.prototype = {
         // Make our elements part of the page
         this._target.appendChild(this._screen);
 
+        this._cursor.attach(this._canvas);
+
         // Monitor size changes of the screen
         // FIXME: Use ResizeObserver, or hidden overflow
         window.addEventListener('resize', this._eventHandlers.windowResize);
@@ -426,6 +431,7 @@ RFB.prototype = {
 
     _disconnect: function () {
         Log.Debug(">> RFB.disconnect");
+        this._cursor.detach();
         this._canvas.removeEventListener("mousedown", this._eventHandlers.focusCanvas);
         this._canvas.removeEventListener("touchstart", this._eventHandlers.focusCanvas);
         window.removeEventListener('resize', this._eventHandlers.windowResize);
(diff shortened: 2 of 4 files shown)


core/display.js
```diff
@@ -258,41 +282,43 @@ Display.prototype = {
                 'type': 'flip'
             });
         } else {
-            var x, y, vx, vy, w, h;
-
-            x = this._damageBounds.left;
-            y = this._damageBounds.top;
-            w = this._damageBounds.right - x;
-            h = this._damageBounds.bottom - y;
+            let damage = { x: this._damageBounds.left,
```
DirectXMan12: avoid creating temporary objects like this -- it's a big performance drain on fast-path functions (because some GCs don't handle it well)

core/display.js
```diff
@@ -258,41 +282,43 @@ Display.prototype = {
...
                 //        as well (see copyImage()), but we haven't
                 //        noticed any problem yet.
                 this._targetCtx.drawImage(this._backbuffer,
-                                          x, y, w, h,
-                                          vx, vy, w, h);
+                                          rect.x, rect.y, rect.w, rect.h,
+                                          vx, vy, rect.w, rect.h);
+
+                if (this._fallbackCursor) {
+                    let cursor = { x: this._fallbackCursorPos.x,
```
DirectXMan12: ditto here

CendioOssman:
> I'm also not certain I'm a fan of this approach -- honestly, I feel like the "fake cursor using transparent canvas" approach described in #572 (comment) is likely to be a bit cleaner

I tried that first and was actually more complex. I also didn't get it to work. IE kept changing the bounding box of things when that canvas was moved around, so scroll bars got screwed up.

core/display.js
CendioOssman: Any GC we care about though? A decent Javascript engine is pretty much a pre-requisite for noVNC anyway.

Any suggestions? We need to return multiple values from `intersect()`, so I'm not seeing a clear way to get rid of temporary objects.

DirectXMan12:
Does this not work: https://github.com/novnc/noVNC/commit/97fbc2bee3a09ec1fac281516435d65369f773e1 ?

If it seems to do what you intended, I'll create a PR for it (it also seems to be a bit simpler, IMO).

CendioOssman:
Actually it is very buggy when I'm testing here. But it does not have the scroll issue at least. However that might be because it has `position: fixed;`. I don't think that is something we can generally use as we would have to update the position on every movement of the main canvas element, including scrolling. That would be very complex, if not impossible, to track reliably.

I've pushed my attempt here:

https://github.com/CendioOssman/noVNC/tree/cursor_dom

It uses a different DOM structure to make sure we always have coordinates relative to the same base as the main canvas.

samhed:
I think the draw-on-canvas approach looks good, it feels like a natural place to have a fallback for this sort of thing. Why do you think a DOM element for the cursor feels cleaner?

DirectXMan12:
> It uses a different DOM structure to make sure we always have coordinates relative to the same base as the main canvas.

That looks a lot better to me, although I'd prefer the logic be factored out into a helper class, so that the main rfb code doesn't have a lot of cursor-displaying-related logic in it (i.e. similarly to how it's done in my rough prototype above).

> I think the draw-on-canvas approach looks good, it feels like a natural place to have a fallback for this sort of thing. Why do you think a DOM element for the cursor feels cleaner?

I find the code a *lot* easier to read at a glance, and it allows us to more easily keep the fallback code logically separated.

CendioOssman:
> That looks a lot better to me, although I'd prefer the logic be factored out into a helper class, so that the main rfb code doesn't have a lot of cursor-displaying-related logic in it (i.e. similarly to how it's done in my rough prototype above).

That code needs to find a solution to misbehaving scroll bars in IE though, and I don't have any good ideas for that right now. :/

I'm wondering if there is an alternative approach to this. Right now we've been focusing on a rather tight integration with RFB, basing things on pointer events there. What if we instead see it as a general cursor emulation for any HTML element and design it based on that. That will probably make the position calculations more straight forward, and `position: fixed` probably makes more sense in that model. But it suffers from the issue of tracking DOM layout changes.

I'm happy with whatever works, but right now the DOM model looks to be more complex and brittle. So I'd prefer to merge this, and revisit this in the future when hopefully the browsers have fewer issues we need to deal with.

DirectXMan12:
> That code needs to find a solution to misbehaving scroll bars in IE though, and I don't have any good ideas for that right now. :/

Let me poke around a bit.  What version of IE were you seeing this in?

DirectXMan12:
gah!  IE 11 on Windows 10 does not reproduce.  IE 11 on windows 8 does.

So, the key thing I'm looking for here it that it doesn't make the display code change, because I don't want to make that have more cases, and it feels like mixing implementation details.

That leaves two options, IMO.

1. Somehow make my solution work with canvas movement.  I'm not overly concerned with scrolling, since it shouldn't be possible to scroll the page while focus is on the canvas.  Similarly, if you move the canvas otherwise (flying canvas?), I'd expect the cursor to stay in the same spot relative to the viewport, not relative to the canvas (the downside there is that in case of flying canvases, the remote machine doesn't get mouse move events from the flying canvas, but I can live with that).  Have I missed something here?

2. Render on a display sized canvas (like your original solution), but on a *different* canvas (transparent, no pointer events) just for cursors, and let the browser composite things together (this was an IE-fueled-rage idea).

CendioOssman:
Well, the cursor is shown even without focus so there are possible cases.

I can have another look and see what's possible with a more general approach.

DirectXMan12:
> Well, the cursor is shown even without focus so there are possible cases.

Yeah, we should probably hide the cursor on blur, if we're doing a pseudo-cursor, to mimic the behavior of normal local cursor.

CendioOssman:
What the?! I added a comment here March 15, but it seems to have been lost to the void. No wonder I never got a reply from you @DirectXMan12. :)

Anyway, I did a new attempt at the floating canvas thing here:

https://github.com/CendioOssman/noVNC/tree/cursor3

It seems to work well. Have a look.

DirectXMan12:
I'll try and take a peek this week.  Fair warning, I'll be at a conference (KubeCon, in Copenhagen) next week, so my review might be delayed.

DirectXMan12:
@CendioOssman code in the new version looks reasonable, and pretty clean too.

samhed:
Is it this code https://github.com/CendioOssman/noVNC/tree/cursor3 you had a look at @DirectXMan12? Should we make a new PR for that one?

DirectXMan12:
> Should we make a new PR for that one?

Yeah.

CendioOssman:
No need for a new PR. It's the same thing we're trying to do, just a bit different. So I've pushed the other code here. Good to merge?

CendioOssman:
That looks like some npm bug. And it works fine here. Let's restart the job and see if it succeeds.

CendioOssman:
Didn't help. :/

Something is wonky with travis' npm setup. Any ideas on how to fix it? I can not reproduce the issue here. I'll try nuking the travis cache and see if that helps...

CendioOssman:
Yay! that worked!

CendioOssman:
@DirectXMan12 , you still with us? :)
If I don't hear anything I'll go ahead and merge this.

CendioOssman:
Crap. It seems I got too focused on IE, so I didn't actually test with a real touch device. Something is off there...

CendioOssman:
Updated commit which gets everything working for all the major browsers. Will merge next week.

DirectXMan12:
Oh, yeah, I'm still alive.  Lots of travel and a mad dash to get some stuff done for a deadline ;-).  Sorry about losing track of this.

sdellutri:
@CendioOssman - I have been trying to follow this issue and use the enhancements that you have done here, but I am not successful and hoping you can help me.

I am using the noVNC library inside of my application and using it on a touch screen windows system with chrome browser.   I cannot get the local cursor working.  I have updated to the latest master and still no luck.   Is there anything I need to do ? (settings or API calls) to make it working properly?
Thanks in advance for your help.

sdellutri:
Thanks so much for your fast reply :) !
I am using TightVNC server on windows as the server and using websockify  via NodeJS 
When connecting from non touch screen system it works ok.. but with touch screen its only remote cursor.   Sometimes I am having very highly latent connections so the remote cursor is very difficult to work with while the local one works great.   Do you have any suggestions or ideas on how I can troubleshoot this?

CendioOssman:
Please open a new issue for this with details of your setup and we can continue the discussion there. :)

sdellutri:
Thank you!
Opened #1132 

https://github.com/novnc/noVNC/issues/1132#issue-354714036

Issue #1004, opened 2018-01-25
Keyboard input stops working after pressing ALT (Gr) with Firefox Quantum


sonicnkt:
Hi,
i noticed that keyboard input (letters and numbers only, backspace and enter still work) after i press the Right Alt key. i have an @ symbol in my password and with a german keyboard layout you have to press ALT-GR + Q to creat this symbol.
VNC Password works fine btw, im talking about entering this symbol after a connection is established (windows password in my case)

But like i said after i the symbol is added cant input anything else until i reload the complete page and connect again.

Cant test it with another browser atm because chromium refuses to load my novnc page at all with my current test setup. Maybe this is a firefox related problem but i thought i ask if someone else experiences this. And I just cloned the latest repository to make sure i have the latest version.

Nils

CendioOssman:
Which version or commit of noVNC are you testing with?

And what does the server setup look like?

sonicnkt:
Just updated my setup 2-3 Hours ago, so it should be 3729976.

This a basic testing setup with node.js running websockify and no custom settings beside a self signed certificate, everything running on windows 10 LTSB.  VNC server is UltraVNC, input with other viewers works fine and noVNC works great as well besides the described problem with the input.

CendioOssman:
I don't have a UltraVNC server to test with here, but I don't see any issues with a TigerVNC server and Firefox.

Can you enable debug in the noVNC settings and see what you get in the browser console when you press AltGr+Q and any presses after that?

sonicnkt:
I had some very poor performance results with tigervnc compared to ultravnc and it has less config options. I can try it with another server as well, but i think its not related to it.

I did what you ask and i noticed something, it seems to be that after i press the right-alt key that the ctrl key seems to be locked, even tho i never touched that one. I just noticed it because while typing my debug test in notepad.exe the "Bearbeiten" menu item open after the text input stopped working (press B), that would be the normal behavior for CTRL+B.

Here is the Debug log:
`
>> Keyboard.allKeysUp
keyboard.js:297:8
<< Keyboard.allKeysUp
keyboard.js:302:8
>> Keyboard.allKeysUp
keyboard.js:297:8
<< Keyboard.allKeysUp
keyboard.js:302:8
>> set_cursor
rfb.js:2506:8
<< set_cursor
rfb.js:2525:8
onmousebutton down, x: 354, y: 319, bmask: 1
mouse.js:101:8
onmousebutton up, x: 354, y: 319, bmask: 1
[...]

PR #1035, opened 2018-03-09
Better detection of AltGr on Windows

CendioOssman:
Try to properly detect the fake CtrlL+AltR sequence Windows sends when pressing AltGr. This allows us to send more accurate key events over to the server.

This mainly makes things better when using QEMU key events, but other cases will improve as well.

The remaining corner case is Ctrl+AltGr which only works if the browser has special handling for it (only Chrome so far). On most browsers we'll get Ctrl+Alt instead. But before this patch we got nothing at all, so it is not necessarily worse.

Based on the same work in TigerVNC:

https://github.com/TigerVNC/tigervnc/commit/5124978b10641ca25607480f3bfbe9c29b5b6cc4

### DIFF
--- a/core/input/keyboard.js
+++ b/core/input/keyboard.js
@@ -21,6 +21,7 @@ export default function Keyboard(target) {
     this._keyDownList = {};         // List of depressed keys
                                     // (even if they are happy)
     this._pendingKey = null;        // Key waiting for keypress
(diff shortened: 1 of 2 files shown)


CendioOssman:
Tested on Chrome, IE and Firefox. Works well (with workaround for #1004).

PR #1036, opened 2018-03-09
Work around broken Alt keyup in Firefox

CendioOssman:
Firefox no longer sends keyup events properly for the Alt keys. Try to sniff out the state of the Alt key by monitoring other events that include its state.

Relevant Firefox bugs:

https://bugzilla.mozilla.org/show_bug.cgi?id=1335347
https://bugzilla.mozilla.org/show_bug.cgi?id=1435717

### DIFF
--- a/core/input/keyboard.js
+++ b/core/input/keyboard.js
@@ -27,7 +27,8 @@ export default function Keyboard(target) {
         'keyup': this._handleKeyUp.bind(this),
         'keydown': this._handleKeyDown.bind(this),
         'keypress': this._handleKeyPress.bind(this),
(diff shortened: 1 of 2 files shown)


core/input/keyboard.js
```diff
@@ -281,6 +282,25 @@ Keyboard.prototype = {
         Log.Debug("<< Keyboard.allKeysUp");
     },
 
+    _checkAlt: function (e) {
```
DirectXMan12: please leave a comment here about why this is necessary (I see you go into more detail below, but a brief

```js
// work around broken Alt key handling on FF for Windows (see below)
```

would be good here too

PR #1038, opened 2018-03-12
fix bug for compatible with webkit-3.0

xiaolou86:


CendioOssman:
The principle is sound, but we also grab `contextmenu` so we need to check which event type it is.

samhed:
How old is webkit-3.0 and where do you use it? I can't find any information about webkit versions. The only browser that uses webkit now-a-days is safari right?

xiaolou86:
@CendioOssman Yes. It should be more careful.
@samhed  Its api docs are here: https://lazka.github.io/pgi-docs/#WebKit-3.0

DirectXMan12:
what version of webkit are we on for safari?  I'm not sure how much I want to support some super-old version that's not relevant for any real browsers.

CendioOssman:
I'd say somewhere around version 10 of Safari, which wikipedia claims is webkit 602. Don't know how that relates to webkit-3 versions. I would assume that Safari 10 supports event constructors though, so I'd say webkit-3 is older.

DirectXMan12:
Agreed, feel free to close

xiaolou86:
Sorry to replay so late.

PR #1048, opened 2018-04-07
Add translation in zh_CN

ghostplant:
Add translation in zh_CN.

Signed-off-by: CUI Wei <ghostplant@qq.com>

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1651,7 +1651,7 @@ var UI = {
 };
 
 // Set up translations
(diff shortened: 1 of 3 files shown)


samhed:
Thank you!

I'll update po/Makefile as well

samhed:
> I'll update po/Makefile as well

nvm, you already did that :)

ghostplant:
yes, but still need po2js to generate a js file which is not done yet~

samhed:
Yes, that will be done before the next release

ghostplant:
sounds good~

PR #1050, opened 2018-04-12
add get-only connectionState property and dispatch CustomEvent when state is connecting

jussisaurio:
**This is a:** feature

**What does it change:**
- `RFB.prototype.connectionState` can be used to get, but not set, `_rfb_connection_state`
- A `CustomEvent` is dispatched when state is updated to `connecting`.
- Adds tests for both

**Why?**

Getter: through public properties/methods, there does not seem to be a direct way of telling what the state of the connection is at any given moment. Being able to get the value on request enables custom UI behavior based on the precise state without hacking into the internal API.

"Connecting" event: as the 'disconnect' and 'connect' events are surely perceived to be useful by most contributors/users of this library, surely 'connecting' is valuable as well, as that event can be used to display custom progress indicators and the like.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -312,6 +312,9 @@ RFB.prototype = {
         }
     },
 
+    get connectionState() { return this._rfb_connection_state; },
+    set connectionState(val) { Log.Error("Cannot assign value " + val + " to get-only property connectionState"); },
+
     // ===== PUBLIC METHODS =====
 
     disconnect: function () {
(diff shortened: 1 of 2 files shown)


samhed:
We had this kind of API up until recently, it was removed with this PR:

https://github.com/novnc/noVNC/pull/950

There is no scenario where the RFB object would unexpectedly end up in a `connecting` state, since as @CendioOssman mentioned, it will start in that state, and you can't reconnect using the same RFB object. The interface can easily implement for example a throbber inbetween creating the RFB object and getting the `connected`-event.

For user-initated client-side disconnects and `disconnecting`, the case is the same as for `connecting`, can't happen unexpectedly. You can add a disconnecting-throbber inbetween the button-press and the `disconnected`-event.

For server-side disconnects there is nothing that can be done since `RFB.connectionState` will always go directly to `disconnected`.

Feel free to reopen if you have a different use-case and we have missed something.

PR #1055, opened 2018-04-18
Optionally read path from cookie

ryanlovett:
Permit a server to set a fallback path to the websocket without altering the URL via query string or hash fragment. The cookie takes least precedence and is named following the convention of the input id in vnc.html.

### DIFF
--- a/vnc_lite.html
+++ b/vnc_lite.html
@@ -187,7 +187,11 @@
         }
 
         var password = WebUtil.getConfigVar('password', '');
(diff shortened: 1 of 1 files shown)


CendioOssman:
`vnc_lite.html` is supposed to be a small example of how to use the library, so I'm afraid this is a step in the wrong direction.

It might be something for the normal UI though, but it would probably be at a general settings level in that case.

PR #1056, opened 2018-04-19

ryanlovett:
If noVNC is not served at the root path, prefix the websocket path with the server path.  This makes the path to the websocket consistent with the path to all the other web assets. This is convenient so that users do not need to manually change the path setting, though they still can override it.

For example if the websockify server is at /some/sub/dir, with this patch noVNC will connect to /some/sub/dir/websockify rather than /websockify.

### DIFF
--- a/vnc_lite.html
+++ b/vnc_lite.html
@@ -201,7 +201,7 @@
                 defaultPath = pathName + '/websockify';
             }
         }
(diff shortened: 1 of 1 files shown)


CendioOssman:
`vnc_lite.html` is supposed to be a small example of how to use the library, so I'm afraid this is a step in the wrong direction.

It might be a good idea to put in the normal UI though.

PR #1066, opened 2018-04-24
Update the internationalization module

colin-zhou:
1.  Fix specification in zh_cn.po
2.  Make it work with lower version of gettext(0.18.2)


CendioOssman:
Hang on, `--language=Python`? These are Javascript files, not Python.

colin-zhou:
Thanks for your correction. It is not necessary to specify the language, I remove this now.

https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html

chnhyg:
Hello Zhou,
"Connect" is "连接" instead of "链接" and you need to fix it.

samhed:
@chnhyg I fixed that in 7449170cc89ab7a8cb6aa829f2b213a6fd936be6

PR #1067, opened 2018-04-26
Correct the URL command line to open a specific VNC host and port

petitlouis:


CendioOssman:
There must be some misunderstanding here. The host and port is for the WebSocket proxy, not the VNC server. If your VNC server has built in support for the WebSocket protocol then launch.sh is not needed.

PR #1073, opened 2018-05-04
Fix/ie11 numpad5 compatibility

vlastoun:
This should fix numpad 5 compatibility with internet explorer 11.

### DIFF
--- a/core/input/domkeytable.js
+++ b/core/input/domkeytable.js
@@ -75,7 +75,7 @@ addNumpad("PageUp", KeyTable.XK_Prior, KeyTable.XK_KP_Prior);
 // 2.5. Editing Keys
 
 addStandard("Backspace", KeyTable.XK_BackSpace);
-addStandard("Clear", KeyTable.XK_Clear);
+addNumpad("Clear", KeyTable.XK_Clear, KeyTable.XK_KP_Begin);
 addStandard("Copy", KeyTable.XF86XK_Copy);
 // - CrSel
 addStandard("Cut", KeyTable.XF86XK_Cut);
--- a/core/input/vkeys.js
+++ b/core/input/vkeys.js
@@ -58,6 +58,7 @@ export default {
     0x63: 'Numpad3',
     0x64: 'Numpad4',
     0x65: 'Numpad5',
+    0x0c: 'Numpad5', // IE11 sends evt.keyCode: 12 when numlock is off
     0x66: 'Numpad6',
     0x67: 'Numpad7',
     0x68: 'Numpad8',
--- a/core/base64.js
+++ b/core/base64.js
@@ -100,7 +100,7 @@ export default {
 
         // If there are any bits left, the base64 string was corrupted
         if (leftbits) {
-            err = new Error('Corrupted base64 string');
+            var err = new Error('Corrupted base64 string');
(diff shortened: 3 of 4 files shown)


CendioOssman:
Under what circumstances does the 5 key misbehave?

core/input/domkeytable.js
```diff
@@ -75,7 +75,7 @@ addNumpad("PageUp", KeyTable.XK_Prior, KeyTable.XK_KP_Prior);
 // 2.5. Editing Keys
 
 addStandard("Backspace", KeyTable.XK_BackSpace);
-addStandard("Clear", KeyTable.XK_Clear);
+addNumpad("Clear", KeyTable.XK_Clear, KeyTable.XK_KP_Clear);
```
CendioOssman: Chrome seems to do the same thing, even on Linux. So this seems correct.

core/input/keysym.js
```diff
@@ -94,6 +94,7 @@ export default {
     XK_KP_Next:                     0xff9b,
     XK_KP_Page_Down:                0xff9b,
     XK_KP_End:                      0xff9c,
+    XK_KP_Clear:                    0xff9d,
```
CendioOssman: However this is wrong. There is no such keysym. Use `XK_KP_Begin`, which is what a Unix system generates for that key.

core/input/vkeys.js
```diff
@@ -12,7 +12,7 @@
 export default {
     0x08: 'Backspace',
     0x09: 'Tab',
-    0x0a: 'NumpadClear',
```
CendioOssman: Keep this. It is for an actual, separate `[Clear]` button.

core/input/vkeys.js
```diff
@@ -12,7 +12,7 @@
...
+    0x0c: 'NumpadClear',
```
CendioOssman: The physical key is still `Numpad5`, so add that and add a comment why there is a special one here.

CendioOssman:
> Is there any way to check state of numLock key in connected machine?

Not actively, no. But there is an extension to keep things in sync. Unfortunately it's rather difficult to implement in noVNC as we cannot ask the browser what the current state is.

core/input/vkeys.js
```diff
@@ -58,6 +58,7 @@ export default {
     0x63: 'Numpad3',
     0x64: 'Numpad4',
     0x65: 'Numpad5',
+    0x0c: 'Numpad5', // ie11 sends evt.keyCode 12
```
CendioOssman: I was thinking more along the lines of when this happens. :)
E.g. "this is sent when numlock is off".

core/input/util.js
```diff
@@ -52,6 +52,7 @@ export function getKeycode(evt){
                 case 'PageDown': return 'Numpad3';
                 case 'ArrowLeft': return 'Numpad4';
                 case 'ArrowRight': return 'Numpad6';
+                case 'NumpadClear': return 'Numpad5';
```
CendioOssman: This can be removed now that `vkeys.js` is updated.

CendioOssman:
Dammit. Why doesn't github notify properly when new commits are pushed? I'll try to have a look at this today.

Issue #1065, opened 2018-04-24
clipboard error


StefanHoutzager:
I get this error after pasting a large text in the noVNC clipboard and copy from that clipboard. noVNC clone from about a week ago.

![novnc-error](https://user-images.githubusercontent.com/6817215/39175319-4c03172c-47aa-11e8-9ad0-523a60476dcd.png)

samhed:
I can reproduce this issue both in Chrome 65 and in Firefox 59 on Linux.

samhed:
The size of the send-buffer is `10240` and in `_encode_message()` we create a new view of the send buffer. The last two arguments to the Uint8Array constructor is the range of the send buffer which we want to use. Trying to send something larger than 10240 bytes won't work with the current code. We need to add some check in the code here.

samhed:
Could you please verify that #1074 fixes your issue?

StefanHoutzager:
Wow, great! I will test next week and let you know, thanks.
Good weekend, Stefan.

      Van: Samuel Mannehed <notifications@github.com>
 Aan: novnc/noVNC <noVNC@noreply.github.com> 
Cc: Stefan Houtzager <agent_008_nl@yahoo.com>; Author <author@noreply.github.com>
 Verzonden: vrijdag 4 mei 16:41 2018
 Onderwerp: Re: [novnc/noVNC] clipboard error (#1065)
   
Could you please verify that #1074 fixes your issue?—

samhed:
#1074 has been merged now, feel free to test the current master.

StefanHoutzager:
I tested the clipboard error fix today, but get the error again.
![clipboard](https://user-images.githubusercontent.com/6817215/40538208-544ca02c-6011-11e8-852e-ab8b526d6f42.png)

samhed:
That's a new error, how much are you trying to paste?

samhed:
I can't reproduce your error, even with an 8MB large paste. Could you upload the file somewhere so I can try with yours? Could you also, in detail, describe the steps you take to get this error?

samhed:
Also, which VNC server do you use?

StefanHoutzager:
I will find some time next monday for answers. Thanks for your interest!

      Van: Samuel Mannehed <notifications@github.com>
 Aan: novnc/noVNC <noVNC@noreply.github.com> 
Cc: Stefan Houtzager <agent_008_nl@yahoo.com>; Author <author@noreply.github.com>
[...]

PR #1074, opened 2018-05-04
Handle sending large clipboards

samhed:
Pasting clipboard texts that were larger than 10240 bytes didnt work and caused a crash in noVNC. This commit fixes the crash and adds handling for sending large clipboard texts. Fixes issue #1065.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1,7 +1,7 @@
 /*
  * noVNC: HTML5 VNC client
  * Copyright (C) 2012 Joel Martin
- * Copyright (C) 2017 Samuel Mannehed for Cendio AB
+ * Copyright (C) 2018 Samuel Mannehed for Cendio AB
  * Licensed under MPL 2.0 (see LICENSE.txt)
  *
  * See README.md for usage and integration instructions.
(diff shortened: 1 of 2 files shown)


core/rfb.js
```diff
@@ -1688,19 +1688,42 @@ RFB.messages = {
...
-        for (var i = 0; i < n; i++) {
-            buff[offset + 8 + i] =  text.charCodeAt(i);
-        }
+        sock._sQlen += 8;
 
-        sock._sQlen += 8 + n;
-        sock.flush();
+        // We can only flush a max of sQbufferSize bytes at a time and we have
+        // already used 8 bytes for the msg-type, padding and length.
+        let remaining = length;
```
CendioOssman: The comment seems a bit odd here. Remnant from earlier version of this code?

core/rfb.js
```diff
@@ -1688,19 +1688,42 @@ RFB.messages = {
...
+
+        // We have to keep track of from where in the text we begin creating the
+        // buffer for the flush in the next iteration.
+        let textOffset = 0;
+
+        while (remaining > 0) {
+
+            let flushSize = Math.min(remaining, (sock._sQbufferSize - sock._sQlen));
+            if (flushSize <= 0) {
+                break;
```
CendioOssman: Shouldn't we have a `fail()` in that case? We've screwed up the protocol at this point.

core/rfb.js
samhed: right, that one should be removed

core/rfb.js
samhed: yeah, I'll add that

samhed:
Thank you for reviewing @CendioOssman. I have fixed the two things you pointed out.

PR #1080, opened 2018-05-21
Add issue templates

CendioOssman:
We often have to ask for the same information for every new issue. Try to avoid this by using github's templates for issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,34 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Client (please complete the following information):**
+ - OS: [e.g. iOS]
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,17 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+
+---
+
+**Is your feature request related to a problem? Please describe.**
```
samhed: "Description of the problem that the feature solves"

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,17 @@
...
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
```
samhed: "Description of my solution"

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,17 @@
...
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
```
samhed: "Alternatives that has been considered"

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+**Describe the bug**
```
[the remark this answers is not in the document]
CendioOssman: Probably, but I think this is more fool proof and I don't mind if the result reads more like a dialogue rather than a document.

Do you think it is very bad to keep it this way?

PR #1083, opened 2018-05-25
Use es6 classes and arrow functs

juanjoDiaz:
Continuation of #1013

Great to see that stuff is finally going in.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,15 +1,17 @@
 {
-  "env": {
-    "browser": true,
-    "es6": true
-  },
-  "parserOptions": {
-    "sourceType": "module"
-  },
-  "extends": "eslint:recommended",
-  "rules": {
-      "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
-      "no-constant-condition": ["error", { "checkLoops": false }],
-      "no-var": "error"
-  }
+    "env": {
+        "browser": true,
+        "es6": true
+    },
+    "parserOptions": {
+        "sourceType": "module"
+    },
+    "extends": "eslint:recommended",
+    "rules": {
+        "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
+        "no-constant-condition": ["error", { "checkLoops": false }],
+        "no-var": "error",
+        "no-useless-constructor": "error",
+        "object-shorthand": ["error", "methods", { "avoidQuotes": true }],
+    }
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,5 +13,10 @@
         "no-var": "error",
         "no-useless-constructor": "error",
         "object-shorthand": ["error", "methods", { "avoidQuotes": true }],
+        "prefer-arrow-callback": "error",
+        "arrow-body-style": ["error", "as-needed", { "requireReturnForObjectLiteral": false } ],
+        "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
+        "arrow-spacing": ["error"],
+        "no-confusing-arrow": ["error", { "allowParens": true }],
     }
 }
--- a/app/localization.js
+++ b/app/localization.js
@@ -10,17 +10,17 @@
  * Localization Utilities
  */
 
-export function Localizer() {
-    // Currently configured language
-    this.language = 'en';
+export class Localizer {
+    constructor() {
+        // Currently configured language
+        this.language = 'en';
 
-    // Current dictionary of translations
-    this.dictionary = undefined;
-}
+        // Current dictionary of translations
+        this.dictionary = undefined;
+    }
 
-Localizer.prototype = {
     // Configure suitable language based on user preferences
-    setup: function (supportedLanguages) {
+    setup(supportedLanguages) {
         this.language = 'en'; // Default: US English
 
         /*
@@ -78,21 +78,22 @@ Localizer.prototype = {
                 return;
             }
         }
-    },
+    }
 
     // Retrieve localised text
-    get: function (id) {
+    get(id) {
         if (typeof this.dictionary !== 'undefined' && this.dictionary[id]) {
             return this.dictionary[id];
         } else {
             return id;
         }
-    },
+    }
 
     // Traverses the DOM and translates relevant fields
     // See https://html.spec.whatwg.org/multipage/dom.html#attr-translate
-    translateDOM: function () {
+    translateDOM() {
         const self = this;
+
         function process(elem, enabled) {
             function isAnyOf(searchElement, items) {
                 return items.indexOf(searchElement) !== -1;
@@ -160,8 +161,8 @@ Localizer.prototype = {
         }
 
         process(document.body, true);
-    },
-};
+    }
+}
 
 export const l10n = new Localizer();
 export default l10n.get.bind(l10n);
--- a/app/ui.js
+++ b/app/ui.js
@@ -41,7 +41,7 @@ const UI = {
     reconnect_callback: null,
     reconnect_password: null,
 
-    prime: function(callback) {
+    prime(callback) {
         if (document.readyState === "interactive" || document.readyState === "complete") {
             UI.load(callback);
         } else {
@@ -51,12 +51,12 @@ const UI = {
 
     // Setup rfb object, load settings from browser storage, then call
     // UI.init to setup the UI/menus
-    load: function(callback) {
+    load(callback) {
         WebUtil.initSettings(UI.start, callback);
     },
 
     // Render default UI and initialize settings menu
-    start: function(callback) {
+    start(callback) {
 
         // Setup global variables first
         UI.isSafari = (navigator.userAgent.indexOf('Safari') !== -1 &&
@@ -116,7 +116,7 @@ const UI = {
         }
     },
 
-    initFullscreen: function() {
+    initFullscreen() {
         // Only show the button if fullscreen is properly supported
         // * Safari doesn't support alphanumerical input while in fullscreen
         if (!UI.isSafari &&
@@ -130,7 +130,7 @@ const UI = {
         }
     },
 
-    initSettings: function() {
+    initSettings() {
         // Logging selection dropdown
         const llevels = ['error', 'warn', 'info', 'debug'];
         for (let i = 0; i < llevels.length; i += 1) {
@@ -169,7 +169,7 @@ const UI = {
         UI.setupSettingLabels();
     },
     // Adds a link to the label elements on the corresponding input elements
-    setupSettingLabels: function() {
+    setupSettingLabels() {
         const labels = document.getElementsByTagName('LABEL');
         for (let i = 0; i < labels.length; i++) {
             const htmlFor = labels[i].htmlFor;
@@ -195,7 +195,7 @@ const UI = {
 * EVENT HANDLERS
 * ------v------*/
 
-    addControlbarHandlers: function() {
+    addControlbarHandlers() {
         document.getElementById("noVNC_control_bar")
             .addEventListener('mousemove', UI.activateControlbar);
         document.getElementById("noVNC_control_bar")
@@ -228,7 +228,7 @@ const UI = {
         }
     },
 
-    addTouchSpecificHandlers: function() {
+    addTouchSpecificHandlers() {
         document.getElementById("noVNC_mouse_button0")
             .addEventListener('click', function () { UI.setMouseButton(1); });
         document.getElementById("noVNC_mouse_button1")
@@ -277,7 +277,7 @@ const UI = {
             .addEventListener('touchmove', UI.dragControlbarHandle);
     },
 
-    addExtraKeysHandlers: function() {
+    addExtraKeysHandlers() {
         document.getElementById("noVNC_toggle_extra_keys_button")
             .addEventListener('click', UI.toggleExtraKeys);
         document.getElementById("noVNC_toggle_ctrl_button")
@@ -292,7 +292,7 @@ const UI = {
             .addEventListener('click', UI.sendCtrlAltDel);
     },
 
-    addMachineHandlers: function() {
+    addMachineHandlers() {
         document.getElementById("noVNC_shutdown_button")
             .addEventListener('click', function() { UI.rfb.machineShutdown(); });
         document.getElementById("noVNC_reboot_button")
@@ -303,7 +303,7 @@ const UI = {
             .addEventListener('click', UI.togglePowerPanel);
     },
 
-    addConnectionControlHandlers: function() {
+    addConnectionControlHandlers() {
         document.getElementById("noVNC_disconnect_button")
             .addEventListener('click', UI.disconnect);
         document.getElementById("noVNC_connect_button")
@@ -315,7 +315,7 @@ const UI = {
             .addEventListener('click', UI.setPassword);
     },
 
-    addClipboardHandlers: function() {
+    addClipboardHandlers() {
         document.getElementById("noVNC_clipboard_button")
             .addEventListener('click', UI.toggleClipboardPanel);
         document.getElementById("noVNC_clipboard_text")
@@ -326,7 +326,7 @@ const UI = {
 
     // Add a call to save settings when the element changes,
     // unless the optional parameter changeFunc is used instead.
-    addSettingChangeHandler: function(name, changeFunc) {
+    addSettingChangeHandler(name, changeFunc) {
         const settingElem = document.getElementById("noVNC_setting_" + name);
         if (changeFunc === undefined) {
             changeFunc = function () { UI.saveSetting(name); };
@@ -334,7 +334,7 @@ const UI = {
         settingElem.addEventListener('change', changeFunc);
     },
 
-    addSettingsHandlers: function() {
+    addSettingsHandlers() {
         document.getElementById("noVNC_settings_button")
             .addEventListener('click', UI.toggleSettingsPanel);
 
@@ -357,7 +357,7 @@ const UI = {
         UI.addSettingChangeHandler('reconnect_delay');
     },
 
-    addFullscreenHandlers: function() {
+    addFullscreenHandlers() {
         document.getElementById("noVNC_fullscreen_button")
             .addEventListener('click', UI.toggleFullscreen);
 
@@ -374,7 +374,7 @@ const UI = {
  * ------v------*/
 
     // Disable/enable controls depending on connection state
-    updateVisualState: function(state) {
+    updateVisualState(state) {
 
         document.documentElement.classList.remove("noVNC_connecting");
         document.documentElement.classList.remove("noVNC_connected");
@@ -441,7 +441,7 @@ const UI = {
             .classList.remove('noVNC_open');
     },
 
-    showStatus: function(text, status_type, time) {
+    showStatus(text, status_type, time) {
         const statusElem = document.getElementById('noVNC_status');
 
         clearTimeout(UI.statusTimeout);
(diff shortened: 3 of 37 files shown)


juanjoDiaz:
Did anyone had the chance of taking a look at this? 🙂

CendioOssman:
Not yet. But hopefully soon. :)

app/localization.js
```diff
@@ -78,33 +78,32 @@ Localizer.prototype = {
...
-        const self = this;
+    translateDOM() {
         function process(elem, enabled) {
             function isAnyOf(searchElement, items) {
                 return items.indexOf(searchElement) !== -1;
             }
 
             function translateAttribute(elem, attr) {
-                const str = self.get(elem.getAttribute(attr));
+                const str = this.get(elem.getAttribute(attr));
```
CendioOssman: These are broken. `this` will be `undefined` in these functions.

core/base64.js
```diff
@@ -11,8 +11,7 @@ export default {
     toBase64Table : 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/='.split(''),
     base64Pad     : '=',
 
-    encode: function (data) {
-        "use strict";
+    encode(data) {
```
CendioOssman: This is a separate cleanup, so could we have that in another commit?

core/display.js
```diff
@@ -10,109 +10,112 @@
...
+
+        // ===== PROPERTIES =====
+
+        this._scale = 1.0;
+        this._clipViewport = false;
+        this.logo = null;
+
+        // ===== EVENT HANDLERS =====
+
+        this.onflush = () => {}; // A flush request has finished
```
CendioOssman: I was thinking even further up, so we can avoiding splitting up the two parts of the properties. :)
(this is fine as well though)

app/ui.js
```diff
@@ -552,8 +552,7 @@ const UI = {
         const barDisplayStyle = window.getComputedStyle(bar).display;
         if (barDisplayStyle !== 'none') {
             bar.style.transitionDuration = '0s';
-            bar.addEventListener('transitionend', function () {
-                this.style.transitionDuration = ""; });
+            bar.addEventListener('transitionend', () => this.style.transitionDuration = '');
```
CendioOssman: I don't think this will work as `this` will not be properly rebound to `bar`.

app/webutil.js
```diff
@@ -119,7 +119,7 @@ export function initSettings (callback /*, ...callbackArgs */) {
     "use strict";
     const callbackArgs = Array.prototype.slice.call(arguments, 1);
     if (window.chrome && window.chrome.storage) {
-        window.chrome.storage.sync.get(function (cfg) {
+        window.chrome.storage.sync.get((cfg) => {
             settings = cfg;
             if (callback) {
                 callback.apply(this, callbackArgs);
```
CendioOssman: `this` will change value here. Have you checked the callers and made sure they are unaffected?

karma.conf.js
```diff
@@ -99,6 +99,7 @@ module.exports = function(config) {
       'tests/test.*.js': ['babel'],
       'tests/fake.*.js': ['babel'],
       'tests/assertions.js': ['babel'],
+      'tests/payback*.js': ['babel'],
```
CendioOssman: Not part of tests, so should not be included.

tests/test.rfb.js
```diff
@@ -23,18 +23,18 @@ import sinon from '../vendor/sinon.js';
     window.UIEvent = UIEvent;
 })();
 
-const push8 = function (arr, num) {
+const push8 = (arr, num) => {
```
CendioOssman: Wouldn't it be better to turn these in to normal functions instead?

utils/genkeysymdef.js
```diff
@@ -108,7 +108,7 @@ out +=
 "};\n" +
 "\n" +
 "export default {\n" +
-"    lookup : function(u) {\n" +
+"    lookup(u) {\n" +
```
CendioOssman: Group this with other other function declaration cleanups in to a separate commit.

utils/use_require.js
```diff
@@ -39,15 +39,14 @@ const no_transform_files = new Set([
...
 
 // util.promisify requires Node.js 8.x, so we have our own
 function promisify(original) {
-    return function () {
-        const obj = this;
+    return () => {
         const args = Array.prototype.slice.call(arguments);
         return new Promise((resolve, reject) => {
-            original.apply(obj, args.concat((err, value) => {
+            original.apply(this, args.concat((err, value) => {
```
CendioOssman: This will be incorrect. `this` needs to be from the call to the unnamed function. With an arrow function you'll instead get `this` from the `promisify()` call.

utils/use_require.js
```diff
@@ -70,29 +69,26 @@ const babelTransformFile = promisify(babel.transformFile);
...
-            .then(stats => {
-                if (filter !== undefined && !filter(filepath, stats)) return;
-
-                if (stats.isSymbolicLink()) return;
-                if (stats.isFile()) return cb(filepath);
-                if (stats.isDirectory()) return walkDir(filepath, cb, filter);
-            });
-        }));
+        return Promise.all(paths.map(filepath => lstat(filepath) .then((stats) => {
+            if (filter !== undefined && !filter(filepath, stats)) return;
```
CendioOssman: It is usually good form to keep every `.then` on a new line for readability.

utils/use_require_helpers.js
```diff
@@ -4,11 +4,10 @@ const path = require('path');
 
 // util.promisify requires Node.js 8.x, so we have our own
 function promisify(original) {
-    return function () {
-        const obj = this;
+    return () => {
         const args = Array.prototype.slice.call(arguments);
         return new Promise((resolve, reject) => {
-            original.apply(obj, args.concat((err, value) => {
+            original.apply(this, args.concat((err, value) => {
```
CendioOssman: See above.

core/display.js
juanjoDiaz: I'm not sure if I understand. Where were you planning to put these? Merged with the properties?

tests/test.rfb.js
juanjoDiaz: What do you mean?

`push8(arr, num)` // Not possible since it's not in a function
`const push8 = function (arr, num) {` Zero advantages since it´s a pure function.
`function push8(arr, num) {` I prefer this notation but I tried to minimize the changes.

utils/genkeysymdef.js
juanjoDiaz: What do you mean?
There are 2 commits:
1.- Change the objects to classes
2.- Change all the functions to arrow functions

(this change is part of the second one)

app/webutil.js
juanjoDiaz: Yes. The context is never used in the callbacks use through the application

This change also makes the scope of the callback consistent regardless of whether chrome storage is used or not. Before it was different.

juanjoDiaz:
I think that everything is fixed except for the couple question that I've made above.

The failing CI seems to be one of those weird Travis issues unrelated to the code.

core/display.js
CendioOssman: Basically something like this:

```Javascript
        this._tile16x16 = this._drawCtx.createImageData(16, 16);
        Log.Debug("<< Display.constructor");

        // ===== EVENT HANDLERS =====

        this.onflush = () => {}; // A flush request has finished

        // ===== PROPERTIES =====

        this._scale = 1.0;
        this._clipViewport = false;
        this.logo = null;
    }

    // ===== PROPERTIES (cont.) =====

    get scale() { return this._scale; }
    set scale(scale) {
        this._rescale(scale);
    }

```

But those old school event handlers will eventually be removed, so it is perhaps nothing to bother with...

.eslintrc
```diff
@@ -11,6 +11,10 @@
         "no-unused-vars": ["error", { "vars": "all", "args": "none", "ignoreRestSiblings": true }],
         "no-constant-condition": ["error", { "checkLoops": false }],
         "no-var": "error",
-        "no-useless-constructor": "error"
+        "no-useless-constructor": "error",
+        "prefer-arrow-callback": ["error", { "allowNamedFunctions": false, "allowUnboundThis": true }],
```
CendioOssman: These are the default options. Can't we just omit them in that case?

.eslintrc
```diff
@@ -11,6 +11,10 @@
...
+        "arrow-body-style": ["error", "as-needed", { "requireReturnForObjectLiteral": false }],
+        "arrow-parens": ["error", "as-needed", { "requireForBlockBody": true }],
+        "arrow-spacing": ["error", { "before": true, "after": true }], "no-confusing-arrow": ["error", { "allowParens": true }],
```
CendioOssman: Need a newline here for the last rule

tests/test.rfb.js
CendioOssman: The last one. You already had a bunch of such transformations, so we've already started down that path.

utils/genkeysymdef.js
CendioOssman: I'd like a third one:

3.- Consistent function declarations (`foo: function()` => `function foo()`, `const foo = function()` => `function foo()`, etc.)

utils/use_require.js
```diff
@@ -39,15 +39,14 @@ const no_transform_files = new Set([
...
 ]);
 
-no_copy_files.forEach((file) => no_transform_files.add(file));
+no_copy_files.forEach(file => no_transform_files.add(file));
 
 // util.promisify requires Node.js 8.x, so we have our own
 function promisify(original) {
-    return function () {
-        const obj = this;
+        return function () {
```
CendioOssman: The indentation is off here.

CendioOssman:
> The failing CI seems to be one of those weird Travis issues unrelated to the code.

Looks like it, yes. You can usually restart the job in those cases and get it to pass.

utils/genkeysymdef.js
juanjoDiaz: I see. Currently:
1.- The shorthand notation is used if the function is a property of an object or class `{ foo() { ... } }` or `class bar { foo() { ... } }`
2.- The fat arrow notation `const foo = () => { ... };` for callback and any other function declared.
3.- The calssic `function foo() { ... }` for top level functions or functions that depend on the scope.

I prefer `const foo = () => { ... };` over `function foo() { ... }` for callbacks and functions that are not declared at the top level since it ensures that `foo` can't be overriden and automatically binds the function scope.

Would you like me to do a 3rd commit replacing all the `const foo = () => { ... };` by `function foo() { ... }` regardless of where they are?

.eslintrc
juanjoDiaz: Do you mean that the line is too long?

.eslintrc
juanjoDiaz: I guess that we could remove the options yeah.
I just copy-pasted them from air-bnb style config
I guess that it's more explicit and less prone to errors when updating in case eslint changes defaults.

But I can remove them if you prefer that :)

utils/genkeysymdef.js
CendioOssman: > So following this logic, lookup is correct as a shorthand function. It wouldn't even be a classic function unless we move it out of the object.

Fair enough. Might want to note that in the commit message in that case so that people aren't confused that the commit modifies more than just classes.

> I generally prefer const foo = () => { ... }; over function foo() { ... } for callbacks and functions that are not declared at the top level since it ensures that foo can't be overriden and automatically binds the function scope.
> 

What do you mean when you say it binds the function scope? Is there some practical detail I'm overlooking here?

I'm just concerned that `const foo = ()` is very similar to the `foo: function()` style we want to move away from. Also, coming from other languages, `const foo` reads more like "reference to an existing function" than a function definition, so that is tripping me up a bit.

.eslintrc
CendioOssman: Very subjective I guess, but personally I would read such things as deviations from the defaults. :)

utils/genkeysymdef.js
juanjoDiaz: I meant that
`const foo = () => { ... };`
is equivalent to
```
function foo() { ... }
foo.bind(this)
```
so you don't need to do stuff like 
```
const self = this;
function foo() { ... <use self> ... }
```

Other difference is that const is safer to accidental overriding.
```
const foo = () => { ... };
foo = 2; // Throws an error saying that you can't override foo
foo();
```
while 
```
function foo() { ... }
foo = 2;
foo(); // Throws an error because foo is not a function
```

So nothing major. I think that my 3 points above summarize well my point of view :)

juanjoDiaz:
I've removed the defaults from the eslint config, added a third commit normalizing the function declarations and added the rationale for that in the commit message as you suggested.

Is there anything else left?

CendioOssman:
Almost. Some details were a bit off in the last commit:

a) The rationale should be split out to the relevant commit. The three parts seem to correspond to the three commits, unless I'm missing something?

b) Most of the changes in commit 3 touch the same functions as in the earlier two commits. Let's avoid the initial steps and just change them once.

c) The final result of the encoding helper functions is broken as we now got an incorrect binding of `this`.

juanjoDiaz:
Alright.

a) Removed the unneeded changes from the first to commits that were overridden by the third commit
b) Fixed the commit message
c) Can you clarify? Do you mean the `encode` method in `Base64`? That shouldn't be broken since it was an old style function and not it's a shorthand function which is not automatically bound like fat arrow functions. Or you were talking about something else?

CendioOssman:
It's part of the Tight decoder. E.g. `indexedToRGBX2Color()`.

tests/test.webutil.js
```diff
@@ -117,9 +117,9 @@ describe('WebUtil', function() {
                 window.chrome = {
                     storage: {
                         sync: {
-                            get: function(cb){ cb(settings); },
-                            set: function(){},
-                            remove: function() {}
+                            get(cb){ cb(settings); },
+                            set(){},
+                            remove() {}
```
samhed: Inconsistent spacing between () and {}. I prefer a space like you did for remove() {}.

core/input/keyboard.js
```diff
@@ -307,7 +307,7 @@ export default class Keyboard {
 
         const target = this._target;
         const downList = this._keyDownList;
-        ['AltLeft', 'AltRight'].forEach(function (code) {
+        ['AltLeft', 'AltRight'].forEach((code) => {
```
samhed: In the other places you removed the parentheses around the parameter:

    code => { ... }

However, I think I prefer leaving the parentheses in like this, but it should at least be consistent.

core/rfb.js
```diff
@@ -199,17 +199,17 @@ export default class RFB extends EventTargetMixin {
...
                 Log.Debug("Starting VNC handshake");
             } else {
                 this._fail("Unexpected server connection while " +
-                        this._rfb_connection_state);
+                           this._rfb_connection_state);
             }
-        }.bind(this));
-        this._sock.on('close', function (e) {
+        });
+        this._sock.on('close', (e) => {
```
samhed: another inconsistency with (e) => { ... } vs e => { ... }

core/rfb.js
```diff
@@ -454,18 +452,15 @@ export default class RFB extends EventTargetMixin {
         const stats = this._encStats;
 
         Log.Info("Encoding stats for this connection:");
-        Object.keys(stats).forEach(function (key) {
+        Object.keys(stats).forEach((key) => {
```
samhed: dito

core/util/events.js
```diff
@@ -120,7 +120,7 @@ export function releaseCapture () {
 
         // There might be events already queued, so we need to wait for
         // them to flush. E.g. contextmenu in Microsoft Edge
-        window.setTimeout(function(expected) {
+        window.setTimeout((expected) => {
```
samhed: inconsistency

core/util/eventtarget.js
```diff
@@ -32,7 +32,7 @@ export default class EventTargetMixin {
       if (!this._listeners || !this._listeners.has(event.type)) {
          return true;
       }
-      this._listeners.get(event.type).forEach(function (callback) {
+      this._listeners.get(event.type).forEach((callback) => {
```
samhed: dito

core/util/logging.js
```diff
@@ -24,7 +24,7 @@ export function init_logging (level) {
         _log_level = level;
     }
 
-    Debug = Info = Warn = Error = function (msg) {};
+    Debug = Info = Warn = Error = () => {};
```
samhed: did you mean to remove the 'msg' parameter?

core/websock.js
```diff
@@ -225,7 +223,7 @@ export default class Websock {
                 this._websocket.close();
             }
 
-            this._websocket.onmessage = function (e) { return; };
+            this._websocket.onmessage = () => {};
```
samhed: 'e' went missing?

core/websock.js
```diff
@@ -196,25 +194,25 @@ export default class Websock {
...
             if (this._websocket.protocol) {
                 Log.Info("Server choose sub-protocol: " + this._websocket.protocol);
             }
 
             this._eventHandlers.open();
             Log.Debug("<< WebSock.onopen");
-        }).bind(this);
-        this._websocket.onclose = (function (e) {
+        };
+        this._websocket.onclose = (e) => {
```
samhed: inconsistency

core/websock.js
```diff
@@ -196,25 +194,25 @@ export default class Websock {
...
             Log.Debug(">> WebSock.onclose");
             this._eventHandlers.close(e);
             Log.Debug("<< WebSock.onclose");
-        }).bind(this);
-        this._websocket.onerror = (function (e) {
+        };
+        this._websocket.onerror = (e) => {
```
samhed: inconsistency

karma.conf.js
```diff
@@ -1,6 +1,6 @@
 // Karma configuration
 
-module.exports = function(config) {
+module.exports = (config) => {
```
samhed: inconsistency

po/xgettext-html
```diff
@@ -89,7 +89,7 @@ for (let i = 0; i < opt.argv.length; i++) {
     const dom = new jsdom.JSDOM(file, { includeNodeLocations: true });
     const body = dom.window.document.body;
 
-    const locator = function (elem) {
+    const locator = (elem) => {
```
samhed: dito

tests/playback.js
```diff
@@ -96,9 +96,9 @@ export default class RecordingPlayer {
 
     // _enablePlaybackMode mocks out things not required for running playback
     _enablePlaybackMode() {
-        this._rfb._sock.send = function (arr) {};
-        this._rfb._sock.close = function () {};
-        this._rfb._sock.flush = function () {};
+        this._rfb._sock.send = () => {};
```
samhed: 'arr' went missing?

CendioOssman:
The parenthesis were discussed here https://github.com/novnc/noVNC/pull/1013#discussion-diff-186998484L74 (open in private mode and it should load)

juanjoDiaz:
Fix the issue with the rfb functions scope.

Following my own definition, those are functions that should be bound and are passed around so they are good as `const foo = (...) => { ... }`


Regarding @samhed comment. I have a nother PR fixing all the spacing inconsistencies (currently there are 11013 in the code 🙂). It was part of #1013, the mega-PR. Since now I'm  slowly pouring those in small PRs, I'll just create a next PR fixing all those once this one gets in if that's ok.

samhed:
ok

juanjoDiaz:
I just rebased on top of master and fix the conflicts once again.

@CendioOssman anything else pending?

CendioOssman:
I don't think so. So let's merge this. Thanks for your hard work. :)

CendioOssman:
Btw, you don't generally need to rebase to fix conflicts unless there is something major. We can sort out the details in most cases when we merge.

CendioOssman:
Hang on, something is broken with IE. Need to investigate.

CendioOssman:
False alarm (sort-of). We have an existing issue where if you run `npm test` by hand you won't get the transformations necessary to run things in IE.

PR #1087, opened 2018-06-07
[optional] Use relative path for socket

zvs001:
In out case we run multiple novnc instances and this is looks like:
`server/pc1/vnc.html`
`server/pc2/vnc.html`

And i had a problem yesterday that socket trying to connect to `server/websock` instead of `server/pc1/websock`, that is correct in my case.

I see that you already have issue like this. This code should handle both cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ recordings
 *.swp
 *~
 noVNC-*.tgz
+.idea/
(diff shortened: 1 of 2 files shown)


CendioOssman:
We already have #1058 working towards this. Did you look at that?

zvs001:
I did clone yesterday. I got this bug and fixed. Now i see that you have discussion about same problem. I just think to propose you something, it is all.

Issue #1054, opened 2018-04-16
Pinch to zoom


thomasphilibert:
hi all, 

Simple question : it's possible to enable the "pinch to zoom" on NoVNC ? 
If no, can i do implement a js module like [this](https://github.com/manuelstofer/pinchzoom) to enable the pinch to zoom ? 

Thank you !

samhed:
No, it's not possible at the moment. It would be very nice to have!

That library looks promising! I have personally tried getting gestures to work with hammer.js but that code was filled with bugs and the maintainers seems to have abandoned the project. I ended up giving up. Would be awesome if we could get something working! Tell me if you need help with implementation or testing.

thomasphilibert:
The library implementation is not the best solution.

But I have an idea: Do you think if it's possible to add an overlay transparent div  (in position absolute and hight z-index) who are displayed only when two finger are on the touchscreen ?

samhed:
What problems have you encountered?

thomasphilibert:
Canvas return a black screen.

It is possible to send a click like an trigger() in Jquery on the display?

samhed:
Does using this library could cause the canvas to draw black? That sounds strange.

Yes, you can create a MouseEvent and use dispatchEvent to trigger it.

thomasphilibert:
Yes but I did not look in detail. I try again soon (i am not an js developper).

Thanks ! the dispatchEvent must be applied directly on the canvas ? Why a simple trigger on canvas no work ? Exemple : 

```
var e = $.Event( "mouseup", { pageX: e.pageX, pageY: e.pageY } );
$('canvas').trigger(e);
```

thomasphilibert:
In vnc-lite.html, i try to add an event on rfb in the main function : 

```
        var element = document.getElementById('customListener');
        element.addEventListener('click', simulateClick);
        simulateClick();

        function simulateClick() {
            var e = new Event( "mousedown", { pageX: 300, pageY: 300 } );
            console.log(e);
            rfb.dispatchEvent(e);
            var e = new Event( "mouseup", { pageX: 300, pageY: 300 } );
            rfb.dispatchEvent(e);
        }
```

But, no work; seem no dispatched on the canvas. my method is not good?

samhed:
What do you mean access? You can use document.getElementById() just like you have done with your customContainer

CendioOssman:
Well, you shouldn't really. The canvas is internal to the RFB object. Anything with a `_` prefix is off limits.

[...]

PR #1093, opened 2018-06-19
Pinch to scroll

lucasraby:
Hi everyone,

For my company i added the pinch to zoom event handler in the mouse module, it work well as far as i know.

If you're intrested into mergin it, feel free to do it.

#1054


CendioOssman:
Ehm... looking at the code this seems to be some gesture to emulate mouse scroll wheel, not pinch zoom?

lucasraby:
yes, indeed.. it emulate a scroll wheel on a pinch gesture, as you point out this may be too specific, i use novnc to stream applications that handle zoom with scroll wheel events.

samhed:
@lucasraby see the feature merged by PR #1414

PR #1098, opened 2018-07-01
Fixes #1075 by removing dependencies on external tools

### LINKED ISSUE #1075: netstat is deprecated, use ss instead
noVNC appears to require the `netstat` command from the `net-tools` package. This package is considered deprecated. `man netstat` will tell you that `This program is mostly obsolete.  Replacement for netstat is ss.`. 

Newer Linux distros don't ship with `net-tools` anymore, instead using the newer and still actively-supported `iproute2` package which includes the `ss` command to replace `netstat`. noVNC should consider replacing its use of `netstat` with `ss`.

hvindin:
Fixes #1075 

Small issue but if you're going to do something you might as well avoid using deprecated tools :do_not_litter: 

Rather than trying to pick a utility, we should be able to just use bash to check if a port is available or not.

We can probably assume bash is available due to the shebang declaring it.

I've tested this on Red Hat 7, Fedora 27, Ubuntu 16.04 and Debian 9 with the version of bash from the default repos.

I can't test this on a mac, so someone might want to double check that.

### DIFF
--- a/utils/launch.sh
+++ b/utils/launch.sh
@@ -70,11 +70,14 @@ while [ "$*" ]; do
 done
 
 # Sanity checks
-which netstat >/dev/null 2>&1 \
-    || die "Must have netstat installed"
-
-netstat -ltn | grep -qs ":${PORT} .*LISTEN" \
-    && die "Port ${PORT} in use. Try --listen PORT"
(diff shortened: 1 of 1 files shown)


lpreams:
Runs fine on my Mac (although noVNC doesn't work with macOS's VNC server so it doesn't really matter anyway)

CendioOssman:
Thank you for your contribution. :)

PR #1102, opened 2018-07-08
use established webRTC connection to initialise RBF object

NaNuNoo:


NaNuNoo:
Sorry, I forget CI.
I will create a new merge request when I fix it.

PR #1103, opened 2018-07-13
Stop transpiling karma tests

CendioOssman:
This runs our code in the same manner as it would be used if loaded directly in the browser. Includes the same kind of fallback for older browsers.

The only downside I see is that IE now runs a bit slower since it has to do the transpiling in the browser. OTOH every other browser runs faster since they don't have to do any transpiling at all.


DirectXMan12:
We probably want a check that our transpilation settings still work properly.  Otherwise, this seems fine.  I *think* the reason we were vendoring sinon before is that the karma one was pulling in a bad version of some other dep, but if that's not an issue anymore, I'm fine with this.

CendioOssman:
IE still gets the transpiled version, but nothing else. That reflects how it happens for users, so hopefully that is enough?

DirectXMan12:
yeah, that's probably fine, then.

PR #1105, opened 2018-07-13
Es6 consistency improvements

juanjoDiaz:
I've applied a bunch of the styles from the airbnb styleguide.

I've left out the ones that you previously staeted that disagree with them (like `+=` instead of `++`) or don't apply (like no bitwise operator).

In the future it might be good to use the airbnb style and override the rules that we don't want in this project.


juanjoDiaz:
Flaky test!

I don't have permissions on this repo so I can't restart the build AFAIK.

juanjoDiaz:
@samhed this is the standardization PR that follows #1083

samhed:
I reran the failing test, passes now.

I'd prefer 4 spaces indentation over 2 spaces.

app/localization.js
```diff
@@ -11,157 +11,153 @@
...
-        }
+    for (let i = 0; i < userLanguages.length; i++) {
+      const userLang = userLanguages[i]
+        .toLowerCase()
+        .replace('_', '-')
+        .split('-');
+
+      // Built-in default?
+      if ((userLang[0] === 'en')
+                && ((userLang[1] === undefined) || (userLang[1] === 'us'))) {
```
samhed: indentation is wrong here

app/localization.js
```diff
@@ -11,157 +11,153 @@
...
+          enabled = true;
+        } else if (isAnyOf(elem.getAttribute('translate'), ['no'])) {
+          enabled = false;
         }
+      }
 
-        process(document.body, true);
+      if (enabled) {
+        if (elem.hasAttribute('abbr')
+                    && elem.tagName === 'TH') {
```
samhed: indentation

CendioOssman:
Alright, so I guess we've taken care of the more practical stuff and are moving in to the subjective style portion.

Unfortunately I have to start with a hard pass on any principle of following airbnb as some gold standard.

Enforcing our existing style still might be interesting. But we'll have to look at it on a case by case basis. I don't think we should be overly anal about style.

Please try to split things up like in earlier PRs so that we can see what each lint setting affects.

juanjoDiaz:
There are a million rules, I've just scratched the surface so far. You can check https://github.com/airbnb/javascript/tree/7ff6303513b46e8b7dd0bc03327b138a6db7b3b4/packages/eslint-config-airbnb-base/rules for example. I've only followed airbnb very loosely.
It's time-consuming and ineffective to have your own guidelines which is why I'd rather use some guidelines that are actively maintained (which minor tweaks if needed). There is no particular reason to use airbnb guidelines, we could use google's, facebook's or any other. I personally rather have a style that I don't like that having no style at all.
I can keep adding rules here but it will always be incomplete and unmaintained.

Regarding the developer experience. Most of this styles rules are automatically applied just running `npm run lint -- --fix`

I'll try to push smaller changes when I have some time.

DirectXMan12:
Having some sort of enforcable style would be nice.  I'm mostly ambivalent on which style is chosen, but choosing a pre-existing one is probably a decent idea -- it means, among other things, that contributors have a chance of already being familiar.

juanjoDiaz:
Exactly.

Here is a list of popular configs: 
https://github.com/dustinspecker/awesome-eslint#configs
Do you guys think that we would be able to agree on one? (So I don't need to keep adding rules manually)

CendioOssman:
I'm afraid I don't see a value in switching over to someone else's style. Developers have to be used to different styles in different projects anyway. And touching lots of lines makes `git blame` more annoying to use, so there is definitely a down side.

So let's start with rules that enforce the style we already have in our code and see what that means.

DirectXMan12:
yeah, I do dislike all the git-blame noise... Unfortunately, I'm unlikely to be a tie-breaker here -- I could go either way ;-).  How difficult is it to write style rules?

samhed:
I prefer #1135 over these changes.

CendioOssman:
Closing in favour of #1135.

Issue #1097, opened 2018-06-29
Need documentation about proper caching headers


patrakov:
**Describe the bug**

Due to too-aggressive caching in modern browsers, it is impossible to update existing noVNC installations without the users who have connected in the past bumping into the error screen.

**To Reproduce**
1. Install Apache2 on Ubuntu 16.04 or 18.04. Let it serve some domain, e.g. http://novnc.example.com/
2. Add the following lines into the default virtual host:

```
                Alias /noVNC /path/to/noVNC

                <Directory /path/to/noVNC>
                    Require all granted
                </Directory>
```
3. In /path/to/noVNC, have a git checkout of noVNC
4. Checkout revision 5585443 (yes I know, very old)
5. Go to http://novnc.example.com/noVNC/vnc.html?1 using Chromium (tested on version 67.0.3396.87) from a different machine. No need to connect to the VNC screen, just load the page.
6. Checkout revision f90c2a6
7. Go to http://novnc.example.com/noVNC/vnc.html?2 using Chromium
8. See the error
9. Press Ctrl+F5, and the error is gone.

(instead of ?1 and ?2 there were actually paths with different websockify tokens)

**Expected behavior**
No error

**Screenshots**

noVNC encountered an error:

Uncaught ReferenceError: Util is not defined
http://novnc.example.com/noVNC/app/ui.js:62:5
ReferenceError: Util is not defined
    at http://novnc.example.com/noVNC/app/ui.js:62:5
    at http://novnc.example.com/noVNC/app/ui.js:1688:3

**Client (please complete the following information):**
 - OS: Linux
 - Browser: Chromium
 - Browser version: 67

**Server (please complete the following information):**
 - noVNC version: see git commit IDs above
 - VNC server: not relevant
 - WebSocket proxy: not relevant

**Additional context**
The bug happens because the browser loads only vnc.html (due to a query string being different from what has been used before), but skips refreshing the javascripts. The web server has to be configured to prevent this (and the required configuration mentioned in the docs), but I don't really know how to do it properly, so no pull request.

I understand that we cannot change the past and thus cannot fix the error completely. However, something has to be done so that it does not reappear in the future.

And, I would be somewhat more happy if the error dialog contained something like this text, in order to help users:

```
A common cause of errors is browser cache.
Try pressing Ctrl+F5 to bypass it.
```

patrakov:
[...]

PR #1107, opened 2018-07-18
Document browser cache issue and add a workaround

patrakov:
This is a minimal documentation-only solution for issue #1097.

### DIFF
--- a/docs/EMBEDDING.md
+++ b/docs/EMBEDDING.md
@@ -81,3 +81,36 @@ load times. To do this please follow these steps:
 
 This will produce a `build/` directory that includes everything needed to run
 the noVNC application.
+
+## HTTP Serving Considerations
+### Browser Cache Issue
+
+If you serve noVNC files using a web server that provides an ETag header, and
+include any options in the query string, a nasty browser cache issue can bite
+you on upgrade, resulting in a red error box. The issue is caused by a mismatch
+between the new vnc.html (which is reloaded because the user has used it with
+new query string after the upgrade) and the old javascript files (that the
+browser reuses from its cache). To avoid this issue, the browser must be told
+to always revalidate cached files using conditional requests. The correct
+semantics are achieved via the (confusingly named) `Cache-Control: no-cache`
+header that needs to be provided in the web server responses.
+
+### Example Server Configurations
+
+Apache:
+
+```
+    # In the main configuration file
+    # (Debian/Ubuntu users: use "a2enmod headers" instead)
+    LoadModule headers_module modules/mod_headers.so
+
+    # In the <Directory> or <Location> block related to noVNC
+    Header set Cache-Control "no-cache"
+```
+
+Nginx:
+
+```
+    # In the location block related to noVNC
+    add_header Cache-Control no-cache;
+```


vnc.html
```diff
@@ -91,6 +91,8 @@
             <div>noVNC encountered an error:</div>
             <br>
             <div id="noVNC_fallback_errormsg"></div>
+            <div>A common cause of errors is browser cache.</div>
```
CendioOssman: I'm sceptical about this. We've seen this for various other things. And if we start recommending deployments with stricter cache control then it should cease to be an issue.

vnc.html
```diff
@@ -91,6 +91,8 @@
...
+            <div>Try pressing Ctrl+F5 to bypass it.</div>
```
CendioOssman: This is not even remotely standardised to give us a cache flush.

docs/EMBEDDING.md
```diff
@@ -81,3 +81,42 @@ load times. To do this please follow these steps:
...
+between the new vnc.html (which is reloaded because the user has used it with
+new query string after the upgrade) and the old javascript files (that the
+browser reuses from its cache). The industry-standard solution to this kind of
+problems is a build step that replaces all links to all javascript files to
+include some token that is derived from the content of the said files. With
+this approach, the browser is free to cache the javascript files forever,
+because, when they change, they get a new URL that is guaranteed not to be in
+the browser cache. Unfortunately, this approach is not implemented in noVNC.
+
+The issue is tracked [on GitHub](https://github.com/novnc/noVNC/issues/1097).
```
CendioOssman: I don't think we should list known bugs/limitations in the documentation. Otherwise we start getting in to a very slippery slope of which issues to include or not.

A generic text about the issue is sufficient.

docs/EMBEDDING.md
```diff
@@ -81,3 +81,42 @@ load times. To do this please follow these steps:
...
+
+Until somebody submits a patch that implements this approach, you have to
+configure your web server to provide headers that prohibit browsers from
+caching noVNC files without revalidation. Example configuration snippets are
+provided below. Both of them result in a `Cache-Control: max-age=0` header.
```
CendioOssman: `max-age` does not guarantee the semantics we want. You need to either also have `must-validate`, or just replace it with `no-cache` which at that point is equivalent.

samhed:
you can force push to the same branch

patrakov:
Thanks for the advice. Let me do it in a few minutes.

patrakov:
Done. Note that I have some more text locally related to serving noVNC, which is unrelated to the cache bug. That's why the generic "HTTP Serving Considerations" heading. I will open a new pull request for it once we sort out the wordings related to the cache issue.

DirectXMan12:
Looks good to me.

DirectXMan12:
What's up with the safari tests?

CendioOssman:
No idea. It worked fine here: https://travis-ci.org/novnc/noVNC/builds/405251529

But this PR looks good, so lets merge. Thanks for your contribution!

0xC4N1:
Thank you for explaining this issue and providing a solution to it. 🖤

I was confused about the red error box because it didn't appear until the update was in production. Disabling the developer tools turned on caching so I had an idea where the issue was coming from. Searching the repository for `cache` made me stumble on 599588f.

Seems like it would be worth mentioning this new requirement in the change logs.

Issue #674, opened 2016-09-30
Copyright format at the top of files


samhed:
As brought up by @DirectXMan12:

Currently we have a list at the top of each file listing contributors each on a separate line. These lists are currently quite incomplete and this style won't work well as the lists grow.

Suggestions? Thoughts @kanaka?

samhed:
Suggested by @DirectXMan12 in @653:

> Copyright (C) the noVNC authors and then have an authors file elsewhere?

kanaka:
I like that plan. In terms of author list we could just expand/complete the last section of the README. If we go with a separate authors file then we should change the README to refer to it (so we just have one copy). I would probably lean slightly towards just updating the README since that's a bit more prominent (and the number of major authors is fairly small). However, I don't have strong feelings about it so I'll go along with whatever the the person who actually does the work decides to do :-)

DirectXMan12:
A separate authors file can be nice if the list ever grows longer.  Either way works for me, though

DirectXMan12:
I'll put up a PR, since I've got a few minutes now :-)

Issue #527, opened 2015-09-03
ASM.js optimizations + web worker support


robertsoakes:
Last year, I had a brief conversation via IRC with Joel about optimizing portions of the code base by re-implementing some of the code in ASM.js and making use of web workers. It's been a while, but I'm interested in re-engaging that conversation.

The company I work for (Guru Labs) uses noVNC in a training/collaboration application and it generally works very well. We have one view, though, where it's possible that 40 or 50 simultaneous sessions might be displayed at once. If all of the machines on the page become active at once, though, it can get really laggy and we've started to get some customer complaints about the interface being "slow".

Seems like a good time to move forward on those ideas, and I've got an engineer who is interested in taking this on as a project. 

As a first step, I would like to re-implement some of the routines as a drop-in ASM.js module and then compare how much of a speedup we get. From there, the next step might be to start looking methods for putting that processing in a web worker.

Which I guess brings me to the point: do you have any feedback on where we might start to experiment? We can always spend quality time with the profiler, but having some candidates on portions of the code that might stand for a rework using ASM.js would be greatly appreciated.

kanaka:
Profiling is never a bad idea. You're definitely going to want to get tests/vnc_perf.html working again. @DirectXMan12 updated tests/vnc_playback.html to get it working and the changes to get tests/vnc_perf.html to work too are probably pretty similar. The tests/vnc_perf.html page is similar but provides quite a bit more detail timing information (per encoding for one thing). It's super useful for performance related work. Both vnc_playback.html and vnc_perf.html playback RFB recordings made with websockify. vnc_playback is more oriented towards demo and vnc_perf towards testing.

In terms areas of code to target for asm.js conversion. The two places that come to mind as probably having the biggest real-world impact are all of `include/jsunzip.js` and `display_tight` in `include/rfb.js`. In addition to likely being in the hot path for most users of noVNC, those pieces of code have the additional benefit of being highly numeric and likely to hugely benefit from asm.js optimizations.

Regarding Web Workers. Now that firefox 38 support websockets in web workers, this is definitely the time to begin exploring that. At a high-level, I think much of the functionality in include/rfb.js, include/websock.js, and the routines in include/display.js that call createImageData and putImageData should run within a web worker. You'll definitely want to look into structured cloning and transferable objects (both forms of zero copy designed for web workers but with varying browser support) for getting the imageData from the web worker to the main thread for display.

[...]

PR #1108, opened 2018-07-20
Move decoders to separate classes

CendioOssman:
Makes things a lot clearer by letting each encoding handle its own details and state.

I've just done the initial shuffle and basic tests right now. Need to do some profiling as well, and an attempt at handling alpha cursors (which requires this change).

If anyone can test TightPNG, then that would be helpful. @DirectXMan12?

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -23,7 +23,7 @@ import DES from "./des.js";
 import KeyTable from "./input/keysym.js";
 import XtScancode from "./input/xtscancodes.js";
 import Inflator from "./inflator.js";
-import { encodings, encodingName } from "./encodings.js";
+import { encodings } from "./encodings.js";
 import "./util/polyfill.js";
 
 // How many seconds to wait for a disconnect to finish
@@ -90,9 +90,8 @@ export default class RFB extends EventTargetMixin {
         this._disconnTimer = null;      // disconnection timer
         this._resizeTimeout = null;     // resize rate limiting
 
-        // Decoder states and stats
+        // Decoder states
         this._encHandlers = {};
-        this._encStats = {};
 
         this._FBU = {
             rects: 0,
@@ -118,19 +117,6 @@ export default class RFB extends EventTargetMixin {
         this._paletteBuff = new Uint8Array(1024);  // 256 * 4 (max palette size * max bytes-per-pixel)
 
         this._rre_chunk_sz = 100;
-
-        this._timing = {
-            last_fbu: 0,
-            fbu_total: 0,
-            fbu_total_cnt: 0,
-            full_fbu_total: 0,
-            full_fbu_cnt: 0,
-
-            fbu_rt_start: 0,
-            fbu_rt_total: 0,
-            fbu_rt_cnt: 0,
-            pixels: 0
-        };
 
         // Mouse state
         this._mouse_buttonMask = 0;
@@ -439,7 +425,6 @@ export default class RFB extends EventTargetMixin {
         this._keyboard.ungrab();
         this._mouse.ungrab();
         this._sock.close();
-        this._print_stats();
         try {
             this._target.removeChild(this._screen);
         } catch (e) {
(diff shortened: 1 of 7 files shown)


core/rfb.js
```diff
@@ -2,12 +2,11 @@
  * noVNC: HTML5 VNC client
  * Copyright (C) 2012 Joel Martin
  * Copyright (C) 2018 Samuel Mannehed for Cendio AB
+ * Copyright (C) 2018 Pierre Ossman for Cendio AB
```
DirectXMan12: If people are ok with it, I think we should just switch this over to `The noVNC authors` -- less copyright churn, among other things.  A number of other opensource projects do this.

core/rfb.js
```diff
@@ -1517,61 +1490,209 @@ export default class RFB extends EventTargetMixin {
...
+                     + msg);
+        } else {
+            this._resize(this._FBU.width, this._FBU.height);
+        }
+
+        return true;
+    }
+
+    _handleDataRect() {
+        this._timing.last_fbu = (new Date()).getTime();
```
DirectXMan12: let's just drop this timing code -- it's not really adding much, IMO.  We can always instrument other ways (esp using the built-in browser dev tools)

DirectXMan12: same with the enc stats.

DirectXMan12:
are the unit tests for tight insufficient?  If so, we should have better unit tests for tight.

core/rfb.js
kanaka: I'm fine with that. When I was the main author with a few people sending bug fixes it was fine just to have my name, but at this point it's a team project so we should reflect that. In the U.S. at least, an author has copyright regardless of whether they label it as such so it's more for documentation than any legal reasons.

core/rfb.js
kanaka: It was really helpful at one point, and gave information that was difficult to get out of profilers (although they are admittedly much better now). If we are considering doing any potential performance impacting changes in the near future it might still be useful to have this. For example, WebAssembly is now supported at a level that it could support a good portion of noVNC users. Anybody inspired to attack that project?

DirectXMan12: WebAssembly in noVNC has been on my radar for a while now, just need to find time.  Maybe this weekend...

I *think* the profilers should be good enough to get most of this now. /me shrugs

kanaka: Well, if you do start playing with WebAssembly let me know. I would be at least interested in seeing what you are working on and might be able to contribute next week to some degree. I might need something to procrastinate on. :-)

core/rfb.js
samhed: #674

core/rfb.js
DirectXMan12: I've started playing a bit.  The problem that I found is that since most of our heavy lifting is copying data around, and webassembly operates on a fixed chunk of memory, we kind-of need the chain of "consume from websocket" to "render to screen" to be written in the source that translates to webassembly.

Started playing around with that for fun (using Rust as the source language), but I also did some profiling (chrome's profiling tools are pretty good).

I recorded `ls -alR` on a decent-sized file tree (a checkout of https://github.com/openshift/origin), coming from a qemu vnc server (which likes to sends lots of tiny TIGHT updates).  TigerVNC takes around a minute to receive/render the frames in realtime.  noVNC takes similar in realtime playback.  In full-speed playback, it takes noVNC about 23 seconds.

Looking at a performance profile (plus some playing around), 14ish of those seconds are spent calling `ctx.putImageData` (a native browser method).  Another 2-5ish seconds are spent doing our actual computations in the TIGHT method (palette stuff, deflate, etc).  AFAICT (doing some small tests), putImageData is actually fairly performance compared to a basic copy written in webassemly (that test was pretty loosely performed, so I can go back and retest more expansively if needed), so I'm not sure how much we gain from conversion to webassembly.  If we're *super-heavily* receiving compressed data, is might help there, but I'd expect more modern servers to more aggressively send normal image data, which we can lean on the browser for heavily.

If people have recordings that seem like they're not performant, I'd love to take a poke at them, but it seems like we do pretty decently.

kanaka: @DirectXMan12 I also did some prototyping/experiments over the weekend (long flight from Japan). I'm going to keep discussing this in the existing asm.js issue (I renamed it to reflect WebAssembly): https://github.com/novnc/noVNC/issues/527

CendioOssman:
@DirectXMan12, this is the state of the Tight tests: :smile:

```Javascript
                it.skip('should handle the TIGHT encoding', function () {
                    // TODO(directxman12): test this
                });

                it.skip('should handle the TIGHT_PNG encoding', function () {
                    // TODO(directxman12): test this
                });
```

Unfortunately it is trivial to make some test cases with the whole zlib thing going on.

CendioOssman:
Updated patch set that completely removes timing and stats.

I've profiled this using `vnc_playback.html` and I cannot measure any performance difference between this code and `master`. So from my POV this is good to go.

CendioOssman:
@DirectXMan12, any objections to merging this?

samhed:
I had another look, looks good to me.

CendioOssman:
Ping @DirectXMan12.

DirectXMan12:
LGTM, sorry for the delay

PR #1113, opened 2018-07-26
Handle fractional screen sizes

### LINKED ISSUE #1111: Scrollbars are sometimes shown in "Local scale" mode
**Describe the bug**
The "Local Scale" mode is supposed to scale the remote screen to fit the browser window. Unfortunately, sometimes it doesn't fit by a fraction of a pixel, and scrollbars appear.

**To Reproduce**
1. Go to your favorite noVNC installation
2. Set the browser zoom to something different than 100%
3. Set the scaling mode to "Local Scale"
4. Connect
5. Play with the browser zoom and window size
6. See a scrollbar

**Expected behavior**
No scrollbar.

**Screenshots**
Well, no screenshots, but some copy-paste from the JS console just to explain what happens here :)

```
> var c = document.getElementsByTagName('canvas')[0];
> c.getBoundingClientRect()
DOMRect {x: 93.1971206665039, y: 0, width: 1768.9903564453125, height: 995, top: 0, …}
> c.offsetHeight
995
> var p = c.parentElement;
> p.getBoundingClientRect()
DOMRect {x: 0, y: 0, width: 1970, height: 994.6154174804688, top: 0, …}
> p.offsetHeight
995
```
See - _screenSize() would think we have a 1970x995 screen, but in fact we have 1970x994 + some fraction of a logical pixel, and it should either think in terms of fractional pixels, or return 1970x994. In other words, the problem is that offsetHeight sometimes rounds up.

 - OS: Arch Linux
 - Browser: Chromium
 - Browser version: Version 67.0.3396.99 (Official Build) Arch Linux (64-bit)

Also reproducible in Firefox 61.0.1 with 300% zoom

 - noVNC version: f90c2a6 with some irrelevant local modifications
 - VNC server: any (tested with x11vnc)
 - WebSocket proxy: websockify

**Additional context**
I was able to fix this problem in Chromium like this, but did not test in anything except Firefox and Chromium:

```
index 3150082..9dc32b2 100644
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -549,8 +549,9 @@ RFB.prototype = {
 
     // Gets the the size of the available screen
     _screenSize: function () {
-        return { w: this._screen.offsetWidth,
-                 h: this._screen.offsetHeight };
+        const r = this._screen.getBoundingClientRect();
+        return { w: Math.floor(r.width),
+                 h: Math.floor(r.height) };
     },
 
     _fixScrollbars: function () {
```

CendioOssman:
With high DPI systems we can end up with a container with a size that
is not an integer number of CSS pixels. Make sure we can handle those
cases by allowing a fractional size for the output canvas. Framebuffer
size and viewport coordinates are still restricted to integer dimensions
though.

Based on initial patch by Alexander E. Patrakov.

Fixes #1111.

More testing is needed. I don't have access to a full set of browsers right now.

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -168,6 +168,9 @@ export default class Display {
             height = this._fb_height;
         }
 
+        width = Math.floor(width);
+        height = Math.floor(height);
+
         if (width > this._fb_width) {
             width = this._fb_width;
         }
@@ -524,8 +527,8 @@ export default class Display {
         //                   style width to a number, the canvas is cleared.
         //                   However, if you set the style width to a string
         //                   ('NNNpx'), the canvas is scaled without clearing.
-        const width = Math.round(factor * vp.w) + 'px';
-        const height = Math.round(factor * vp.h) + 'px';
+        const width = factor * vp.w + 'px';
+        const height = factor * vp.h + 'px';
 
         if ((this._target.style.width !== width) ||
             (this._target.style.height !== height)) {
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -546,7 +546,8 @@ export default class RFB extends EventTargetMixin {
         }
 
         const size = this._screenSize();
-        RFB.messages.setDesktopSize(this._sock, size.w, size.h,
+        RFB.messages.setDesktopSize(this._sock,
+                                    Math.floor(size.w), Math.floor(size.h),
                                     this._screen_id, this._screen_flags);
 
         Log.Debug('Requested new desktop size: ' +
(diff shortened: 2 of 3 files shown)


DirectXMan12:
I'm concerned about passing fractional pixels, since browsers can sometimes be unpredictable with those.

DirectXMan12:
I don't have a high-dpi system screen to play around with, though

CendioOssman:
If you zoom then that will behave just like if the system was a hidpi one.

We will get fractional pixels whether we like it or not as we don't have that kind of control over the container (that's what triggered #1111). So we might as well get this fixed properly.

samhed:
whoops, pushed to the wrong branch.

CendioOssman:
Great. Should test some mobile devices as well.

CendioOssman:
I've now tested on Safari on macOS, Safari on iOS and Chrome and Firefox on Android. I don't see any issues.

PR #1117, opened 2018-07-29
Fix test playback

DirectXMan12:
This fixes playback.js, which got broken somewhere along the way during the refactors.  It also adds the ability to pass `--record` through `utils/launch.sh` to websockify, for easier recording for test purposes.

### DIFF
--- a/tests/vnc_playback.html
+++ b/tests/vnc_playback.html
@@ -2,8 +2,22 @@
 <html>
     <head>
         <title>VNC Playback</title>
-        <script src="/vendor/browser-es-module-loader/dist/browser-es-module-loader.js"></script>
-        <script type="module" src="./playback.js"></script>
+        <!-- promise polyfills promises for IE11 -->
+        <script src="../vendor/promise.js"></script>
+        <!-- ES2015/ES6 modules polyfill -->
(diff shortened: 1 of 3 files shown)


CendioOssman:
The fix looks correct, but something has seriously gone wrong. I'm seeing a massive performance regression with Firefox here. A test file I have that used to run in 6-8 seconds now takes 15 seconds.

tests/vnc_playback.html
```diff
@@ -2,8 +2,22 @@
...
+        <script src="../vendor/promise.js"></script>
+        <!-- ES2015/ES6 modules polyfill -->
+        <script type="module">
+            window._noVNC_has_module_support = true;
+        </script>
+        <script>
+            window.addEventListener("load", function() {
+                if (window._noVNC_has_module_support) return;
+                var loader = document.createElement("script");
+                loader.src = "vendor/browser-es-module-loader/dist/browser-es-module-loader.js";
```
CendioOssman: I retract my previous statement. This line is wrong. An extra `../` is needed here.

CendioOssman:
False alarm on the performance. I was just running things on a much slower machine.

PR #1119, opened 2018-07-30
Show dot when there otherwise would be no visible cursor

### LINKED ISSUE #1031: Cursor not showing when using MobileVNC server
Is there a way to force the cursor on or to get noVNC to properly recognize my server as a server that has a cursor?

The VNC cursor is not showing up in noVNC when I connect to my touch screen/tablet (HMI). The HMI is running on Windows CE and it is running smartlab.at MobileVNC server (http://www.smartlab.at/MobileVNC/). I am using noVNC 1.0.0. I tried Firefox and Chrome with the same results. The cursor shows up on the HMI but not in the browser. I tried RealVNC and the cursor shows up as a small black square in the client and shows up on the HMI.
Here is my canvas information:
`<canvas style="margin: auto; outline: medium none currentcolor; flex-shrink: 0; width: 800px; height: 480px; cursor: none;" width="800" height="480" tabindex="-1"></canvas>` 
(see `cursor:none`)

I was able to get the cursor to show in the browser by "hacking" the `disableLocalCursor` to be `default` instead of `none`:
``` javascript
    disableLocalCursor: function () {
        this._target.style.cursor = "default";
    },
```

patrakov:
Fixes #1031.

The dot is currently enabled unconditionally, I can hide it behind a preference if you want.

I thought about disabling this by default on touch devices, but had to say "no" because my Lenovo Yoga 2 Pro laptop is detected under Linux as a touch device but I still want the dot to be shown there. The laptop has a touchscreen but for me it is an unused extra.


patrakov:
> I'd also like to see it implemented in RFB rather than Cursor

Not sure about a clean interface then. Currently the code in Cursor converts the pixels and mask array into an array of RGBA values. So this is a natural place to see whether any pixel is non-transparent. If we do it in RFB, then we should test the mask directly there, not sure if it is a good idea. Anyway I will try later :)

CendioOssman:
Right, you would need to examine the mask. You could also look at pushing the conversion up a layer. I need to do that in order to support other cursor extensions anyway.

patrakov:
> You could also look at pushing the conversion up a layer.

OK, I understand now what needs to be done.

samhed:
Thanks for working on this, it will be nice to have!

core/rfb.js
```diff
@@ -1603,6 +1616,33 @@ export default class RFB extends EventTargetMixin {
...
+            x: x, y: y, w: w, h: h,
+        }
+        this._refreshCursor();
+    }
+
+    _shouldShowDot() {
+        if (!this._showDot) { return false; }
+        for (let y = 0; y < this._cursorImage.h; y++) {
+            for (let x = 0; x < this._cursorImage.w; x++) {
+                let idx = 4 * (x + this._cursorImage.w * y) + 3;
```
DirectXMan12: why not just do 

```js
for (let i = 3; i < this._cursorImage.length; i += 4) {
   if (this_cursorImage.rgbaPixels[idx]) { return false }
}
```

DirectXMan12: also, this needs comments as to why it's doing what it's doing

core/rfb.js
```diff
@@ -2534,9 +2574,24 @@ RFB.encodingHandlers = {
         this._FBU.bytes = pixelslength + masklength;
         if (this._sock.rQwait("cursor encoding", this._FBU.bytes)) { return false; }
 
-        this._cursor.change(this._sock.rQshiftBytes(pixelslength),
-                            this._sock.rQshiftBytes(masklength),
-                            x, y, w, h);
+        // Decode from pixels + mask to RGBA
+        let rgba = [];
```
DirectXMan12: might want to use a typedarraybuffer here instead.

core/rfb.js
```diff
@@ -2534,9 +2574,24 @@ RFB.encodingHandlers = {
...
+        const pixels = this._sock.rQshiftBytes(pixelslength);
+        const mask = this._sock.rQshiftBytes(masklength);
+
+        for (let y = 0; y < h; y++) {
+            for (let x = 0; x < w; x++) {
+                let idx = y * Math.ceil(w / 8) + Math.floor(x / 8);
+                let alpha = (mask[idx] << (x % 8)) & 0x80 ? 255 : 0;
+                idx = ((w * y) + x) * 4;
```
DirectXMan12: don't re-use `idx` like this.  Make separate variables for `pixelsIdx` vs `maskIdx`.

core/rfb.js
patrakov: I agree

core/rfb.js
patrakov: This was already in the old code that I just moved from one place to the other.

core/rfb.js
patrakov: Ditto

DirectXMan12:
Sorry about stuff that was in the old code before, but we should fix it as we move it :-)

patrakov:
Would you prefer me to amend my commits so that they fix stuff at the same time as moving it, or is it sufficient to fix the typedarraybuffer and reuse-of-idx issue in separate commits?

DirectXMan12:
I'd prefer to fix them as they move it, if it's not too big of an issue.

core/rfb.js
```diff
@@ -165,7 +165,15 @@ export default class RFB extends EventTargetMixin {
         this._canvas.tabIndex = -1;
         this._screen.appendChild(this._canvas);
 
-    this._cursor = new Cursor();
+        // Cursor
+        this._cursor = new Cursor();
+        this._cursorImage = {
+            rgbaPixels: [],
+            x: 0,
+            y: 0,
```
CendioOssman: Could we call these `hotX` and `hotY` so no one thinks it might be a position?

core/rfb.js
```diff
@@ -47,6 +47,7 @@ export default class RFB extends EventTargetMixin {
         this._rfb_credentials = options.credentials || {};
         this._shared = 'shared' in options ? !!options.shared : true;
         this._repeaterID = options.repeaterID || '';
+        this._showDot = options.showDot || true;  // temporarily for debugging
```
CendioOssman: Try to not push your debugging code as it might be overlooked and actually committed. :)

core/rfb.js
```diff
@@ -323,6 +318,12 @@ export default class RFB extends EventTargetMixin {
         }
     }
 
+    get showDot() { return this._showDot; }
```
CendioOssman: I would prefer if "cursor" is in here somewhere to make it clear what it refers to. E.g. `showCursorDot`.

core/rfb.js
```diff
@@ -424,7 +425,11 @@ export default class RFB extends EventTargetMixin {
         // Make our elements part of the page
         this._target.appendChild(this._screen);
 
+        // We don't want to reuse the cursor from the old connection
+        // but maybe want to show a dot instead
```
CendioOssman: I don't understand this comment. RFB objects only exist for a single connection.

core/rfb.js
```diff
@@ -1619,12 +1624,22 @@ export default class RFB extends EventTargetMixin {
         this._refreshCursor();
     }
 
+    _shouldShowDot() {
+        if (!this._showDot) { return false; }
+        for (let y = 0; y < this._cursorImage.h; y++) {
+            for (let x = 0; x < this._cursorImage.w; x++) {
+                let idx = 4 * (x + this._cursorImage.w * y) + 3;
+                if (this._cursorImage.rgbaPixels[idx]) { return false; }
```
CendioOssman: You should probably be prepared for cursors with colourful transparent pixels. (i.e. only check the alpha channel).

core/util/cursor.js
```diff
@@ -213,3 +215,23 @@ export default class Cursor {
         this._canvas.style.top = this._position.y + "px";
     }
 }
+
+Cursor.none = {
+    rgbaPixels: [],
```
CendioOssman: Have you checked that an empty cursor works with all browsers? I'm thinking it might be better to use `cursor: none;` in those cases.

core/util/cursor.js
```diff
@@ -213,3 +215,23 @@ export default class Cursor {
...
+    rgbaPixels: Base64.decode(
+        "/////wAAAP//////" +
+        "AAAA/wAAAAAAAAD/" +
+        "/////wAAAP//////"
+    ),
+    w: 3,
+    h: 3,
+    x: 1,
+    y: 1,
+};
```
CendioOssman: These should be in `rfb.js` IMO as they are not a fundamental type for the `Cursor` class.

vnc.html
```diff
@@ -206,6 +206,9 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br />VNC</h1>
                     <li>
                         <label><input id="noVNC_setting_view_only" type="checkbox" /> View Only</label>
                     </li>
+                    <li>
+                        <label><input id="noVNC_setting_show_dot" type="checkbox" /> Pointer Always Visible</label>
```
CendioOssman: Bike shedding, but I wonder if people might think this will show a normal arrow pointer. For reference RealVNC (and derived software) displays "Show dot when no cursor".

CendioOssman: I think this should be under "Advanced" since it is only for corner cases.

docs/EMBEDDING.md
```diff
@@ -61,6 +61,9 @@ query string. Currently the following options are available:
 * `resize` - How to resize the remote session if it is not the same size as
   the browser window. Can be one of `off`, `scale` and `remote`.
 
+* `show_dot` - If a dot cursor should be shown when the remote server provides
+  no local cursor, or provides a fully-transparent (invisible) cursor.
+
```
CendioOssman: `API.md` also needs to be updated.

And these additions should probably be in the same commit as adds the feature in the code.

vnc.html
DirectXMan12: agreed on that point

core/util/cursor.js
patrakov: That's irrelevant. This "cursor" is a placeholder for initialization purposes only, it is never set. There is already a check in cursor.js that replaces an empty cursor with `cursor: none`.

core/rfb.js
patrakov: But I do exactly this! see `4 * ... + 3` formula.

core/rfb.js
patrakov: OK, so I misunderstood the logic, the comment is wrong, and the code is suboptimal, because the cursor is initialized twice. Will rework.

core/rfb.js
patrakov: OK

core/util/cursor.js
patrakov: OK

patrakov:
Unfortunately i am busy with other things at this time, will work on polishing this pull request later.

patrakov:
Oops, was working from an outdated checkout, please ignore.

CendioOssman:
Is everything updated so we should review this again?

patrakov:
Yes please. I took all review comments into account where I could.

core/rfb.js
```diff
@@ -167,13 +169,13 @@ export default class RFB extends EventTargetMixin {
...
-            w: 0,
-            h: 0,
-        };
+
+        // TightVNC 2.8.11 sends no cursor at all until Windows changes it, so
+        // let's show something visible by default. An arrow is what a user
+        // would expect, but it is too intrusive for servers that always draw
+        // the cursor in the remote framebuffer image. So let's show a small
+        // dot as a compromise.
+        this._cursorImage = RFB.cursors.dot;
```
CendioOssman: I'm not a fan of this. It would punish well behaving servers because of a bug in TightVNC. I think it is better that users will have to use this new setting for such buggy servers.

core/rfb.js
```diff
@@ -2600,3 +2630,17 @@ RFB.encodingHandlers = {
         }
     }
 }
+
+RFB.cursors = {
+    dot: {
+        rgbaPixels: new Uint8Array(Base64.decode(
+            "/////wAAAP//////" +
+            "AAAA/wAAAAAAAAD/" +
+            "/////wAAAP//////"
```
CendioOssman: Nitpick, but do we really need to pull in Base64 just for this? It's just 3x3 pixels. The array can't be that large?

core/util/cursor.js
```diff
@@ -80,6 +80,7 @@ export default class Cursor {
     }
 
     change(rgba, hotx, hoty, w, h) {
+        if (!this._target) { return; }
```
CendioOssman: Hmm... What's the corner case that needed this? And won't it result in the cursor not being properly updated?

core/rfb.js
```diff
@@ -2600,3 +2630,17 @@ RFB.encodingHandlers = {
...
+        )),
+        w: 3,
+        h: 3,
+        x: 1,
+        y: 1,
```
CendioOssman: These should also be changed to `hotX`/`hotY`.

core/rfb.js
CendioOssman: Unfortunately the spec isn't clear on this point, so we'll have to go with observed behaviour. Servers that do server-side rendering of the cursor don't always send an empty local cursor (probably because it is an extension, i.e. optional).

core/util/cursor.js
CendioOssman: Hmm... But if we're not connected then there is no `RFB` object. Or did you manage to find a window whilst it was connecting?

core/rfb.js
patrakov: Given the unclear spec, we cannot call TightVNC behavior a bug. And I agree regarding the servers that do server-side rendering of the cursor. So definitely a case of conflicting requirements, OK, let's resolve it in favor of servers that do server-side rendering of the cursor and thus want the empty initial cursor.

core/util/cursor.js
patrakov: I am currently in a meeting, will recheck later.

patrakov: OK, you seem to be right, this is an artifact from manual console-based testing.

patrakov:
Pushed an updated version, please review again.

patrakov:
Oops. GitHub has reacted badly to my attempt to rebase on the latest code. Not sure what to do.

patrakov:
Repushed.

CendioOssman:
Looks good and seems to work well. Thank you for your hard work. :)

@DirectXMan12, any objections to merging?

vnc.html
```diff
@@ -221,6 +221,9 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br />VNC</h1>
                                 <label for="noVNC_setting_repeaterID">Repeater ID:</label>
                                 <input id="noVNC_setting_repeaterID" type="input" value="" />
                             </li>
+                            <li>
+                                <label><input id="noVNC_setting_show_dot" type="checkbox" /> Show Dot when No Cursor</label>
+                            </li>
```
samhed: I would prefer seeing this placed below the websocket-settings. Either in it's own 'section' with a \<hr\> separating stuff, or  together with the logging-setting.

patrakov: I would prefer it in its own section. Will redo the pull request.

samhed: Sounds great, thank you!

DirectXMan12:
modulo the UI issue, I've got no objections.

patrakov:
The UI issue is resolved now

PR #1124, opened 2018-08-05
Passive  listeners, Touch devices

al3x609:
The touch move events need to block the call to preventDefault (), for a better touch response, the delay in scrolling could improve up to 300 [ms], the test was done on Ipad Air 2 with the audit function of Chrome.

With `{passive: true}` touch and wheel listeners, the Developers promises not to call preventDefault () to disable scrolling.

[more info Passive Listeners](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md)

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -257,27 +257,28 @@ var UI = {
 
         document.documentElement
             .addEventListener('mousedown', UI.keepVirtualKeyboard, true);
-
+        
+        // passive event listeners won't call preventDefault() to disable scrolling
         document.getElementById("noVNC_control_bar")
-            .addEventListener('touchstart', UI.activateControlbar);
+            .addEventListener('touchstart', UI.activateControlbar, {passive: true});
         document.getElementById("noVNC_control_bar")
-            .addEventListener('touchmove', UI.activateControlbar);
+            .addEventListener('touchmove', UI.activateControlbar, {passive: true});
         document.getElementById("noVNC_control_bar")
             .addEventListener('touchend', UI.activateControlbar);
         document.getElementById("noVNC_control_bar")
             .addEventListener('input', UI.activateControlbar);
 
         document.getElementById("noVNC_control_bar")
-            .addEventListener('touchstart', UI.keepControlbar);
+            .addEventListener('touchstart', UI.keepControlbar, {passive: true});
         document.getElementById("noVNC_control_bar")
             .addEventListener('input', UI.keepControlbar);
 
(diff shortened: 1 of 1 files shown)


app/ui.js
```diff
@@ -257,27 +257,28 @@ var UI = {
...
 
         document.getElementById("noVNC_control_bar")
-            .addEventListener('touchstart', UI.keepControlbar);
+            .addEventListener('touchstart', UI.keepControlbar, {passive: true});
         document.getElementById("noVNC_control_bar")
             .addEventListener('input', UI.keepControlbar);
 
         document.getElementById("noVNC_control_bar_handle")
-            .addEventListener('touchstart', UI.controlbarHandleMouseDown);
+            .addEventListener('touchstart', UI.controlbarHandleMouseDown, {passive: true});
```
samhed: controlbarHandleMouseDown() calls preventDefault()

app/ui.js
```diff
@@ -257,27 +257,28 @@ var UI = {
...
         document.getElementById("noVNC_control_bar_handle")
             .addEventListener('touchend', UI.controlbarHandleMouseUp);
         document.getElementById("noVNC_control_bar_handle")
-            .addEventListener('touchmove', UI.dragControlbarHandle);
+            .addEventListener('touchmove', UI.dragControlbarHandle, {passive: true});
```
samhed: dragControlbarHandle() calls preventDefault()

app/ui.js
```diff
@@ -257,27 +257,28 @@ var UI = {
 
         document.documentElement
             .addEventListener('mousedown', UI.keepVirtualKeyboard, true);
-
+        
+        // passive event listeners won't call preventDefault() to disable scrolling
         document.getElementById("noVNC_control_bar")
-            .addEventListener('touchstart', UI.activateControlbar);
+            .addEventListener('touchstart', UI.activateControlbar, {passive: true});
```
CendioOssman: IE doesn't support options so this needs some support code.

CendioOssman:
How is this causing a problem in practice though? `vnc.html` is designed to not use any scrolling.

CendioOssman:
Those will not be passive listeners since we want to consume the event, so we can't really improve things there.

(Also, it shouldn't delay anything there as the reason for the delay is that the browser needs to wait for the Javascript before doing the scrolling. And since the Javascript is the important part in that case we'll always have to wait for it.)

samhed:
Ah, I see, nothing to do here in that case.

PR #1126, opened 2018-08-09
Remove machine control buttons from vnc_lite

samhed:
The vnc_lite example is intended to be minimal and these buttons are only useful in special cases.

### DIFF
--- a/vnc_lite.html
+++ b/vnc_lite.html
@@ -5,7 +5,7 @@
     <!--
     noVNC example: lightweight example using minimal UI and features
     Copyright (C) 2012 Joel Martin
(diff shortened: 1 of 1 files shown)


CendioOssman:
:+1: for me. But I'm wondering if it is better with a more complete PR so we can see what the final result of a minimal `vnc_lite.html` would be.

Some nice things I would like to see:

 * Everything in a single file
 * Focus on being a developer example. I.e. clear code with good comments so that people can easily learn how to integrate the `RFB` object in to their own code.

DirectXMan12:
Yeah, I think that would be a nice exercise

samhed:
Closing in favor of #1129

PR #1129, opened 2018-08-16
Cleanup of vnc_lite.html

samhed:
The aim for vnc_lite.html is to be a minimal developers example for how to integrate RFB into their own application. Here's a summary of what these commits does:

* Cleanup and clarification of the code
* Keep everything in one file
* Remove all dependencies to `app/`
* The following features are removed:
    * using the fragment to pass URL params (only query string is available now)
    * machine control buttons (XVP extension)
    * the option to set log level, title, token, encryption, repeater id, shared mode and auto resize

This PR is based on the discussion in #1126 and includes the commit from that PR as well.


kanaka:
lgtm

vnc_lite.html
```diff
@@ -22,156 +20,171 @@
...
+            // Let's create a password input
+            const form = document.createElement('form');
             form.innerHTML = '<label></label>';
             form.innerHTML += '<input type=password size=10 id="password_input">';
             form.onsubmit = setPassword;
+            // And add place it in the status-bar
+            const statusElem = document.getElementById('noVNC_status');
+            statusElem.innerHTML = '';
+            statusElem.appendChild(form);
+            statusElem.querySelector('label').textContent = 'Password Required: ';
```
CendioOssman: Can't we use `window.prompt()`? It would make things simpler.

vnc_lite.html
```diff
@@ -22,156 +20,171 @@
...
-        document.getElementById('machineResetButton').onclick = machineReset;
 
-        WebUtil.init_logging(WebUtil.getConfigVar('logging', 'warn'));
-        document.title = WebUtil.getConfigVar('title', 'noVNC');
+        // Read parameters specified in the URL query string
         // By default, use the host and port of server that served this file
-        var host = WebUtil.getConfigVar('host', window.location.hostname);
-        var port = WebUtil.getConfigVar('port', window.location.port);
+        const host = readQueryVariable('host', window.location.hostname);
+        let port = readQueryVariable('port', window.location.port);
```
CendioOssman: I wonder if we should reduce this to just `websocket_url` and get rid of the URL code.

vnc_lite.html
```diff
@@ -22,24 +22,9 @@
...
-    <link rel="icon" sizes="72x72" type="image/png" href="app/images/icons/novnc-72x72.png">
-    <link rel="icon" sizes="76x76" type="image/png" href="app/images/icons/novnc-76x76.png">
-    <link rel="icon" sizes="96x96" type="image/png" href="app/images/icons/novnc-96x96.png">
-    <link rel="icon" sizes="120x120" type="image/png" href="app/images/icons/novnc-120x120.png">
-    <link rel="icon" sizes="144x144" type="image/png" href="app/images/icons/novnc-144x144.png">
-    <link rel="icon" sizes="152x152" type="image/png" href="app/images/icons/novnc-152x152.png">
-    <link rel="icon" sizes="192x192" type="image/png" href="app/images/icons/novnc-192x192.png">
-    <!-- Firefox currently mishandles SVG, see #1419039
+    <!-- Icon -->
     <link rel="icon" sizes="any" type="image/svg+xml" href="app/images/icons/novnc-icon.svg">
```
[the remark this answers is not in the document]
samhed: very true, I didn't think of that

vnc_lite.html
```diff
@@ -216,9 +203,7 @@
 
             // Creating a new RFB object will start a new connection
             rfb = new RFB(document.body, url,
-                          { repeaterID: WebUtil.getConfigVar('repeaterID', ''),
-                            shared: WebUtil.getConfigVar('shared', true),
```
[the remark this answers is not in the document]
samhed: hm, ok. I didn't get it to work with x11vnc or tigervnc.. but nvm

vnc_lite.html
samhed: I had a look at that as well, but it doesn't allow us to hide the input as the user is typing it, as is normal for password inputs

vnc_lite.html
```diff
@@ -22,156 +20,171 @@
...
+
+        .noVNC_status_normal {
+            background: linear-gradient(#b2bdcd 0%,#899cb3 49%,#7e93af 51%,#6e84a3 100%);
+        }
+        .noVNC_status_error {
+            background: linear-gradient(#c83737 0%,#899cb3 49%,#7e93af 51%,#6e84a3 100%);
+        }
+        .noVNC_status_warn {
+            background: linear-gradient(#b4b41e 0%,#899cb3 49%,#7e93af 51%,#6e84a3 100%);
+        }
```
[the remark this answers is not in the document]
samhed: I'm fine with that

vnc_lite.html
```diff
@@ -22,156 +20,171 @@
...
-    <link rel="apple-touch-icon" sizes="152x152" type="image/png" href="app/images/icons/novnc-152x152.png">
+    <style type="text/css">
+
+        body {
+            margin:0;
+            background-color:#313131;
+            border-bottom-right-radius: 800px 600px;
+            height:100%;
+            display: flex;
+            flex-direction: column;
```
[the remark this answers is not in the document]
samhed: most of this layout stuff is to get the status text to center properly, if we don't feel that's necessary we can remove a lot

vnc_lite.html
```diff
@@ -162,12 +158,30 @@
...
         }
 
+        // This function extracts the value of one variable from the
+        // query string. If the variable isn't defined in the URL
+        // it returns the default value instead.
+        function readQueryVariable(name, defaultValue) {
+            // A URL with a query parameter can look like this:
+            // https://www.example.com?myqueryparam=myvalue
+            const re = new RegExp('.*[?&]' + name + '=([^&#]*)'),
+                  match = document.location.href.match(re);
```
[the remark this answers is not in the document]
samhed: I'll take a look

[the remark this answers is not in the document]
samhed: Seems like Firefox < 53 had a bug where `search` could be empty even for valid URLs that contained a query string.

All browsers aside from IE support `URLSearchParams` now-a-days. Would be nice to switch to that in the future.

vnc_lite.html
```diff
@@ -84,7 +84,7 @@
     <script>
         window.addEventListener("load", function() {
             if (window._noVNC_has_module_support) return;
-            var loader = document.createElement("script");
+            const loader = document.createElement("script");
```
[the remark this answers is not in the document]
samhed: It seems to be possible using this plugin:

https://www.npmjs.com/package/eslint-plugin-html

I'd be fine skipping that for now however.

vnc_lite.html
samhed: No strong opinions here, but isn't the current approach quite a bit more user friendly?

vnc_lite.html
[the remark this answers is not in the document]
CendioOssman: It's not crucial, it would just be nice now that all other Javascript is being checked.

vnc_lite.html
CendioOssman: Hmmm... that's not ideal. But perhaps it's still worth it to get rid of a bunch of complexity? It is just an example after all.

vnc_lite.html
CendioOssman: Maybe. But it definitely makes the code more difficult to follow.

vnc_lite.html
samhed: Yeah, I think so too

samhed:
Fixed all your comments @CendioOssman aside from eslint for html and the removal of host/port/path.

samhed:
To summarize, this PR now also includes:
* Removal of favicon
* A slightly different look of the top bar with simpler styling
* A simple prompt instead of a password input in the top bar when password is required

vnc_lite.html
```diff
@@ -167,25 +165,14 @@
...
         // Connect
         (function() {
 
             status("Connecting");
 
             // Build the websocket URL used to connect
             var url;
-            if (WebUtil.getConfigVar('encrypt',
-                                     (window.location.protocol === "https:"))) {
+            if (window.location.protocol === "https:") {
```
CendioOssman: I think this change means we can drop the port logic above (the one that sets 80/443).

vnc_lite.html
```diff
@@ -179,16 +179,16 @@
             }
         }
 
-        var password = readQueryVariable('password', '');
-        var path = readQueryVariable('path', 'websockify');
+        const password = readQueryVariable('password', '');
+        const path = readQueryVariable('path', 'websockify');
 
         // Connect
         (function() {
```
CendioOssman: You should be able to nuke this thing as well as it was only there to create a scope AFAICT.

vnc_lite.html
samhed: done

CendioOssman:
LGTM. @DirectXMan12? @kanaka?

DirectXMan12:
code in the final result looks good to me.  Maybe add a comment at the top of the file explaining that this is a self-contained file, so it doesn't import webutil or external CSS.

kanaka:
I think @DirectXMan12's suggestion is good that we should have a note at the top with the purpose of the file. Otherwise looks good.

samhed:
That's a good idea @DirectXMan12 , I added that. Thanks for reviewing

PR #1135, opened 2018-09-06
Enforce code style

CendioOssman:
A suggestion for eslint rules to enforce the style we currently have. This is a small subset meant to cover the stuff we really care about.

Replaces #1105.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+**/xtscancodes.js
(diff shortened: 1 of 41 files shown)


CendioOssman:
I also prefer without the braces, but with braces is the dominant style of the project right now. And I think @DirectXMan12 said that was on purpose?

PR #1151, opened 2018-10-10
add ultravnc single click auth support

secureinfo:
add ultravnc single click auth support

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -863,6 +863,8 @@ export default class RFB extends EventTargetMixin {
                 this._rfb_auth_scheme = 16; // Tight
             } else if (includes(2, types)) {
                 this._rfb_auth_scheme = 2; // VNC Auth
+            } else if (includes(104, types)) {
+                this._rfb_auth_scheme = 104; // UltraVNC SCPrompt
             } else {
                 return this._fail("Unsupported security types (types: " + types + ")");
             }
@@ -1069,18 +1071,39 @@ export default class RFB extends EventTargetMixin {
         return this._fail("No supported sub-auth types!");
     }
 
+    _negotiate_ultravnc_scprompt() {
+        if (this._sock.rQwait("UltraVNC Single Click Prompt length", 4)) {return false;}
+        let strlen = this._sock.rQshift32();
+        Log.Info(strlen);
+        strlen = ((strlen & 0xFF) << 24) | ((strlen & 0xFF00) << 8) | ((strlen >> 8) & 0xFF00) | ((strlen >> 24) & 0xFF);
+        Log.Info(strlen);
+
+        if (this._sock.rQwait("UltraVNC Single Click Prompt", strlen, 4)) {return false;}
+        this._sock.rQskipBytes(strlen);
+
+        this._sock.send([1, 0, 0, 0]);
+
+        this._rfb_init_state = 'SecurityResult';
+
+        return true;
+    }
+
+    _negotiate_noauth() {
+        if (this._rfb_version >= 3.8) {
+            this._rfb_init_state = 'SecurityResult';
+            return true;
+        }
+        this._rfb_init_state = 'ClientInitialisation';
+        return this._init_msg();
+    }
+
     _negotiate_authentication() {
         switch (this._rfb_auth_scheme) {
             case 0:  // connection failed
                 return this._handle_security_failure("authentication scheme");
 
             case 1:  // no auth
-                if (this._rfb_version >= 3.8) {
-                    this._rfb_init_state = 'SecurityResult';
-                    return true;
-                }
-                this._rfb_init_state = 'ClientInitialisation';
-                return this._init_msg();
+                return this._negotiate_noauth();
 
             case 22:  // XVP auth
                 return this._negotiate_xvp_auth();
@@ -1090,6 +1113,9 @@ export default class RFB extends EventTargetMixin {
 
             case 16:  // TightVNC Security Type
                 return this._negotiate_tight_auth();
+
+            case 104:  // UltraVNC SCPrompt
+                return this._negotiate_ultravnc_scprompt();
 
             default:
                 return this._fail("Unsupported auth scheme (scheme: " +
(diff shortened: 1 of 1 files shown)


core/rfb.js
```diff
@@ -64,6 +64,7 @@ export default class RFB extends EventTargetMixin {
         this._rfb_max_version = 3.8;
         this._rfb_tightvnc = false;
         this._rfb_xvp_ver = 0;
+        this._rfb_ultravnc_scinfo = '';
```
CendioOssman: I don't see any need to keep this around, so let it be local for the auth function.

core/rfb.js
```diff
@@ -863,6 +864,10 @@ export default class RFB extends EventTargetMixin {
                 this._rfb_auth_scheme = 16; // Tight
             } else if (includes(2, types)) {
                 this._rfb_auth_scheme = 2; // VNC Auth
+            } else if (includes(17, types)) {
+                this._rfb_auth_scheme = 17; // UltraVNC
+            } else if (includes(104, types)) {
+                this._rfb_auth_scheme = 104; // UltraVNC SC
```
CendioOssman: Note that this list sets which authentication forms we prefer. Are you sure these belong last?

CendioOssman: Do you know of some documentation for these extensions? Do you think you could write up a patch for @rfbproto?

core/rfb.js
```diff
@@ -1069,18 +1074,46 @@ export default class RFB extends EventTargetMixin {
         return this._fail("No supported sub-auth types!");
     }
 
+    _negotiate_ultravnc_auth() {
+        if (this._sock.rQwait("UltraVNC auth result", 4)) {return false;}
+        this._sock.rQshift32();
```
CendioOssman: What are in these bytes? Nothing of value?

core/rfb.js
```diff
@@ -1069,18 +1074,46 @@ export default class RFB extends EventTargetMixin {
...
+        this._rfb_init_state = 'Security';
```
CendioOssman: So we're going back to handshaking for authentication? Why can't we just pick the proper one right away?

core/rfb.js
```diff
@@ -1069,18 +1074,46 @@ export default class RFB extends EventTargetMixin {
...
+    }
+
+    _negotiate_ultravnc_sc_auth() {
+        if (this._sock.rQwait("UltraVNC Single Click auth msg size", 4)) {return false;}
+        const strlen = this._sock.rQshift32L()
+
+        if (this._sock.rQwait("UltraVNC Single Click auth msg body", strlen, 4)) {return false;}
+        this._rfb_ultravnc_scinfo = this._sock.rQshiftStr(strlen);
+
+        Log.Info(this._rfb_ultravnc_scinfo)
```
CendioOssman: Some context for this log line would be nice. And what is in this info?

core/rfb.js
```diff
@@ -1069,18 +1074,46 @@ export default class RFB extends EventTargetMixin {
...
+
+        this._sock.send([1,0,0,0]);
+
+        this._rfb_init_state = 'Authentication';
```
CendioOssman: You are already in that state here so you don't need to change it.

core/websock.js
```diff
@@ -95,6 +95,13 @@ export default class Websock {
                this._rQ[this._rQi++];
     }
 
+    rQshift32L() {
```
CendioOssman: I'm a bit sceptical adding more wrappers here. And this would be incomplete as you would also need 16-bit and peek functions. And there is just the one call to this. I think it is better to swap the result in the handler instead (ntohl() style).

core/rfb.js
```diff
@@ -1083,14 +1083,14 @@ export default class RFB extends EventTargetMixin {
...
+        const strlen = this._sock.rQshift32L();
 
         if (this._sock.rQwait("UltraVNC Single Click auth msg body", strlen, 4)) {return false;}
         this._rfb_ultravnc_scinfo = this._sock.rQshiftStr(strlen);
 
-        Log.Info(this._rfb_ultravnc_scinfo)
+        Log.Info(this._rfb_ultravnc_scinfo);
 
-        this._sock.send([1,0,0,0]);
+        this._sock.send([1, 0, 0, 0]);
```
CendioOssman: Please merge these fixes with the original commit.

vnc_lite.html
```diff
@@ -179,7 +180,7 @@
 
         // Creating a new RFB object will start a new connection
         rfb = new RFB(document.getElementById('screen'), url,
-                      { credentials: { password: password } });
+                      { credentials: { password: password }, repeaterID: repeaterID });
```
CendioOssman: We want to keep `vnc_lite.html` minimal, so I'm afraid we don't want to add more features to it.

core/websock.js
secureinfo: I think it's better to get little endian value in high performance websock. If it is not appropriate, i can convert it in the handler.

core/rfb.js
secureinfo: Ok, I will change it in the next commit.

vnc_lite.html
secureinfo: Ok, I will change it in the next commit.

core/rfb.js
secureinfo: This is the authentication process required by UltraVNC's Signle Click mode.

secureinfo: Ok, I will try to write it there.

core/rfb.js
secureinfo: Sorry,I don't know much about github's pull request, so I don't quite understand what you mean.

secureinfo: Does it mean rebase?

CendioOssman: And then do a force push (`git push -f`) to update the branch and automatically update this pull request.

secureinfo: Thank you for your help.

core/rfb.js
CendioOssman: My point was, do you really want the other authentication types to be used instead of single click?

How does single click work from a user's point of view?

core/websock.js
CendioOssman: There is no performance requirement here. So put it in the handler.

core/rfb.js
CendioOssman: I guess not since you removed all of this?

CendioOssman:
Getting there. But I'll await a suggestion for protocol documentation before continuing. It's very difficult to check if this does the right thing without some reference.

samhed:
I see you pushed new commits, did you adress all the concerns by @CendioOssman ? Looks like there's still some unanswered questions

CendioOssman:
For reference, here is the PR for the protocol documentation:

https://github.com/rfbproto/rfbproto/pull/16

CendioOssman:
No response. Closing.

PR #1166, opened 2018-11-24
Simplify EventTargetMixin

juanjoDiaz:


juanjoDiaz:
Broke it into 2 lines so it's shorter 🙂

core/util/eventtarget.js
```diff
@@ -8,33 +8,28 @@
...
         if (!this._listeners.has(type)) {
             this._listeners.set(type, new Set());
         }
         this._listeners.get(type).add(callback);
     }
 
     removeEventListener(type, callback) {
-        if (!this._listeners || !this._listeners.has(type)) {
-            return;
+        if (this._listeners.has(type)) {
```
CendioOssman: Very minor nitpick, but is it worth having symmetry with `dispatchEvent()`?

juanjoDiaz: The difference is that dispatch returns a boolean and this doesn't return anything.

That would mean:
```javascript
        if (!this._listeners.has(type)) {
            return;
        }

        this._listeners.get(type).delete(callback);
```

which feels a bit unnecessary to me.

But I can change it if you would prefer it 🙂

samhed: I like it the way it is, the functions are different enough.

PR #1167, opened 2018-11-24
Add getters/setter to websock

juanjoDiaz:
Add getter/setter instead of methods to some websock properties.
Replace `rQshift8` , `rQshift16` and `rQshift32` by a more generic `rQshift(bytes)`


juanjoDiaz:
Would you be happier with 
```javascript
    rQshift(bits) {
        let res = 0;
        for (let byte = bytes - 8; byte >= 0; byte -= 8) {
            res += this._rQ[this._rQi++] << (bits);
        }
        return res;
    }
```

so you can call it like `rQshift(16)`

The problem is that the user can then pass 15 and screw up things so we would need to check for that.

Would you have a better suggestion for the naming of either `rQshift` and `rQshiftBytes`?
They are indeed confusing since `rQshift` returns an integer and `rQshiftBytes` returns a string.

CendioOssman:
> Would you have a better suggestion for the naming of either `rQshift` and `rQshiftBytes`?
> They are indeed confusing since `rQshift` returns an integer and `rQshiftBytes` returns a string.

Right. And the number was supposed to be the type. I.e. `rQshift16()` shifts a `16`, i.e. a big endian 16-bit integer.

You can argue that the naming could be better, but I think it very much makes sense to have one function per integer type, even if they have very similar implementation.

CendioOssman:
OTOH, websock is supposed to be generic, so more generic naming might make sense. E.g. `rQshiftInt16()` and `rQshiftUint16()` or something along those lines.

CendioOssman:
OTTH (rambling...), if we're doing an API cleanup then perhaps we should do some more sane naming in general. The whole `rQ` thing is not really in line with other JavaScript APIs (except for jQuery's obsession with brevity).

juanjoDiaz:
Would then `rQshiftInt(bytes)` be a better name?
or `rQshiftInt(bits)`?
Or do you really feel that we should keep the 3 functions?

The only thing I have against that is that we would have 3 functions doing exactly the same. So any bugfix, change, etc. must be done in 3 places...


Regarding `rQ`, do you have a better name? 🙂

CendioOssman:
> Would then `rQshiftInt(bytes)` be a better name?
> or `rQshiftInt(bits)`?
> Or do you really feel that we should keep the 3 functions?
> 

I really think we should have 3 functions. People rarely think of integer sizes in arbitrary bytes or bits. :)

> The only thing I have against that is that we would have 3 functions doing exactly the same. So any bugfix, change, etc. must be done in 3 places...
> 

I'm not opposed to some internal function that makes sure we share code. The API should be more explicit though IMO.

> Regarding `rQ`, do you have a better name? :slightly_smiling_face:

Not really. I've just been annoyed with it, but not enough to come up with a real alternative. :)

juanjoDiaz:
Better now?

core/decoders/tight.js
```diff
@@ -78,8 +78,8 @@ export default class TightDecoder {
             return false;
         }
 
-        const rQi = sock.get_rQi();
-        const rQ = sock.rQwhole();
+        const rQi = sock.rQi;
+        const rQ = sock.rQ;
```
CendioOssman: Is this safe? `rQwhole()` gave you a new object, rather than a reference to an existing one.

core/websock.js
```diff
@@ -45,58 +45,58 @@ export default class Websock {
...
     rQshift32() {
-        return (this._rQ[this._rQi++] << 24) +
-               (this._rQ[this._rQi++] << 16) +
-               (this._rQ[this._rQi++] << 8) +
-               this._rQ[this._rQi++];
+        return this.rQshift(4);
+    }
+
+    // TODO(directxman12): test performance with these vs a DataView
+    rQshift(bytes) {
```
CendioOssman: A `_` prefix would be good to indicate that this is an internal function.

core/websock.js
```diff
@@ -45,58 +45,58 @@ export default class Websock {
...
+        let res = 0;
+        for (let byte = bytes - 1; byte >= 0; byte--) {
+            res += this._rQ[this._rQi++] << (byte * 8);
```
CendioOssman: Have you done any performance tests? There is at least some risk that a for loop will be less efficient than the previous versions. And reading data can be done in very hot code paths.

core/websock.js
```diff
@@ -45,58 +45,58 @@ export default class Websock {
...
-        this._rQi += num;
+    rQshift8() {
+        return this.rQshift(1);
     }
 
-    // TODO(directxman12): test performance with these vs a DataView
     rQshift16() {
-        return (this._rQ[this._rQi++] << 8) +
-               this._rQ[this._rQi++];
+        return this.rQshift(2);
```
CendioOssman: These changes should probably be a separate commit as they don't really match the commit message.

core/websock.js
```diff
@@ -107,36 +107,27 @@ export default class Websock {
...
-        return new Uint8Array(this._rQ.buffer, 0, this._rQlen);
-    }
-
-    rQslice(start, end) {
-        if (end) {
-            return new Uint8Array(this._rQ.buffer, this._rQi + start, end - start);
-        } else {
-            return new Uint8Array(this._rQ.buffer, this._rQi + start, this._rQlen - this._rQi - start);
-        }
+    rQslice(start, end = this.rQlen) {
```
CendioOssman: Default parameters are not supported by IE unfortunately. I guess we did not have a test that caught this?

core/websock.js
```diff
@@ -267,7 +258,7 @@ export default class Websock {
             }
         }
 
-        this._rQlen = this._rQlen - this._rQi;
+        this._rQlen = this.rQlen;
```
CendioOssman: This is very difficult to make sense of, and has a high risk of breaking in the future. Keep the existing code.

core/decoders/tight.js
juanjoDiaz: `rQwhole` made a clone of the object.
That's safer but let's performant.

This function is the only place where `rQwhole` was used and it only reads from the variable but never modifies it.

core/websock.js
juanjoDiaz: Do you mean extracting to a common function?

core/websock.js
juanjoDiaz: No I haven't.
There are definitely few extra operations. although, this could be optimized by the browser engine and, in my experience most of this micro-optimizations yield completely different results based on the browser engine.

Any hint on how would you like this to be tested?

core/websock.js
juanjoDiaz: Neither are arrow functions.
I thought that that is why noVNC is transpiled?

core/websock.js
juanjoDiaz: Sure thing!

CendioOssman:
Sorry about the delays here, things have been hectic. I hope to get to this soon. :/

core/websock.js
CendioOssman: We have `tests/vnc_playback.html` that can be used together with the browsers profiler. I can have a look and see how it behaves.

core/websock.js
CendioOssman: Sorry, you're right. We just need to confirm that the current transpile settings catch this then.

PR #1179, opened 2018-12-21
Super key feature

### LINKED ISSUE #1178: Add support for Super key, or add custom key map options
Add support for Super/Windows key.

Something like the CTRL + ALT + DEL Buttons, but for the Super(Windows) key, so we can send that to the VNC host.

ssebs:
Resolve #1178

sample


samhed:
Hi @ssebs, thanks for a good pull request! Very nice with the .gif as well. I think this would be good addition.

Would you mind combining your commits into one? You should be able to force-push to the same branch.

ssebs:
I've made the requested changes.

samhed:
Great, thank you. I'll take a look tomorrow.

CendioOssman:
Looks good. Please merge all three commits in to one and update this PR though, then we can go ahead and merge it.

Issue #1172, opened 2018-11-30
Scrollbars are not displayed


jsrober:
noVNC does not show scrollbars even though the remote desktop is larger than the browser canvas.

**Client (please complete the following information):**
 - OS: Windows 10
 - Browser: Tested with Chrome, IE, and Edge

**Server (please complete the following information):**
 - noVNC version: 18439b0680811f175e39ce381d0faa138f868d3c
 - VNC server: RealVNC Enterprise
 - WebSocket proxy: websockify

CendioOssman:
Hmmm... Do you see the rest of the session if you change the settings to do scaling instead?

And have you checked that you haven't accidentally enabled "Clip to Window"? :)

jsrober:
Hi @CendioOssman Thank you for your help.

ClipToWindow is checked and grayed out. I am not able to un-check it. Sounds like this might be the source of my problem and that it's my error, not a problem with noVNC. Do you think?

RE: Scaling Mode
When I enable LocalScaling, I can see the entire remote desktop. When I enable RemoteResizing I only see part of the remote desktop and scrollbars are not displayed.

CendioOssman:
Ah. It behaves that way on touch devices as they generally don't show meaningful scroll bars. Is that perhaps the case here?

jsrober:
@CendioOssman maybe, but I don't think so. When I run an older version of noVNC (0.6) I see the scrollbars as I expect.

samhed:
> When I enable RemoteResizing I only see part of the remote desktop and scrollbars are not displayed.

Not all servers support this feature, and I guess RealVNC doesn't.

> @CendioOssman maybe, but I don't think so. When I run an older version of noVNC (0.6) I see the scrollbars as I expect.

Well, do you have a touch device or not?

> When I hover the mouse very close to the edge of the browser where the scrollbar should appear, I don't see a scrollbar and the mouse icon does not change to indicate a scrollbar is there.

That's expected. When "Clip to Window" is enabled, there won't be any scrollbars. You will instead have the hand icon in the controlbar which allows click-to-drag panning.

> How can I disable ClipToWindow? Why is that grayed out in the noVNC settings? Seems like that might be the source of this issue.

If you do have a touch device, that is why it is greyed out. As @CendioOssman mentioned - touch devices do generally not show meaningful scroll bars.

jsrober:
@samhed I am working on a Windows 10 laptop. It does have a touch screen.

[...]

PR #1182, opened 2019-01-09
Enable scrollbars for all touch devices aside from Android and iOS

samhed:
Instead of forcing "clip to window" for all touch devices, we only do it for Android and iOS devices instead. This is where the scrollbars are bad.

This allows users on, for example, Windows with touch to still use scroll bars.

This PR should fix issue #1172

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -8,7 +8,8 @@
 
 import * as Log from '../core/util/logging.js';
 import _, { l10n } from './localization.js';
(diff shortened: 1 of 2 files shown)


samhed:
Fixed the tests now.

jsrober:
Hi @samhed I checked out the branch and tried it out, but it does not show scrollbars. What could I do be doing wrong?

PR #1186, opened 2019-01-13
Add Korean translation

Baw-Appie:


CendioOssman:
Thank you for your contribution. Unfortunately those files are generated and the files that need to be updated are those in the `po/` directory. Please see this wiki page for some more details:

https://github.com/novnc/noVNC/wiki/Translations

Could you update your pull request adding a `kr.po` instead?

PR #1187, opened 2019-01-14
Add a callback to web socket open event handler

mykola-mokhnach:
Such callback might be useful in case one wants to add an additional security layer to the WebSocket communication

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -184,14 +184,14 @@ export default class RFB extends EventTargetMixin {
         this._sock.on('message', this._handle_message.bind(this));
         this._sock.on('open', () => {
             if (typeof options.onSocketOpen === "function") {
-                let result;
+                let errorDescription;
                 try {
-                    result = options.onSocketOpen(this._sock._websocket);
+                    errorDescription = options.onSocketOpen(this._sock._websocket);
                 } catch (e) {
                     return this._fail("onSocketOpen callback failed: " + e.stack);
                 }
-                if (result) {
-                    return this._fail("onSocketOpen callback failed: " + result);
+                if (errorDescription) {
+                    return this._fail("onSocketOpen callback failed: " + errorDescription);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


CendioOssman:
Thank you for your contribution. Could you describe your scenario a bit more? Perhaps with an example of how you want to deploy things? We would prefer events over the old `on...` style callbacks, so help us understand what would fit best.

mykola-mokhnach:
Thanks for the swift feedback @CendioOssman 

In our case there is a custom RFB protocol wrapper, which requires a valid session id to be sent by the client on web socket connection establishment before the "normal" communication could start. 

> We would prefer events over the old on... style callbacks

Sure, simply give an example of how you would like the implementation to look like

CendioOssman:
So you basically need to do some communication over the websocket first? And it your case a simple `send()` is enough?

I'm thinking that if we want to make this a general system then it will be too much of a restriction to only allow `send()` and no responses.

Perhaps change the RFB constructor so that it could accept an already connected WebSocket instead of an URL? Do you think that could work for you?

CendioOssman:
See my previous comment, I'm not sure an event would be sufficient for cases very similar to yours. So would a possible alternative for you to be to provide the already authenticated WebSocket to the RFB constructor?

mykola-mokhnach:
@CendioOssman Unfortunately such approach does not really work, since we anyway must to attach event handlers to this web socket inside the Websock wrapper class to be able to fire internal events and handlers. Of course, it is possible to rewrite the whole  Websock implementation to change that, however this would be an overkill for such a simple task.

CendioOssman:
Right, you would have to have another asynchronous step for that approach to work. But that is not too unexpected for a JavaScript API.

Is there some other approach that might work for you? E.g. using a cookie or embedding things in the WebSocket URL?

I'm not keen on adding this event for your use case as I can see that we'll likely get a request for someone wanting to do the same thing, but wait for the server to respond "OK" to the authentication. And then we would end up maintaining two solutions for pretty much the same thing.

mykola-mokhnach:
Yes, I can understand your concern. We'll then merge the change to our local fork.

PR #1189, opened 2019-01-15
Add a public getter for the _screen property

mykola-mokhnach:
The `_screen` property defines a set of hardcoded css properties. By having this accessor we could easily customise them and/or assign new if needed.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -302,6 +302,8 @@ export default class RFB extends EventTargetMixin {
         this._refreshCursor();
     }
 
+    get screen() { return this._screen; }
+
     // ===== PUBLIC METHODS =====
 
     disconnect() {


mykola-mokhnach:
@samhed Any examples/references please? Would it be ok to add a public method like `setScreenProperty(name, value)` instead?

samhed:
No, I'm thinking more in the line of "setBackgroundColor" in that case.

Could you describe what attributes you want to change?

CendioOssman:
Right, we don't want to expose then internal objects we create. Think of RFB as a single HTML element (which it'll probably be eventually using Shadow DOM). So what functions are you missing from it? What is your use case?

mykola-mokhnach:
this makes sense. I'll create another PR to be able to only change the default screen background, so one could set a branding image or a custom color to it if necessary

PR #1190, opened 2019-01-16
Make the screen background customizable

mykola-mokhnach:
Add a pubic accessor to the screen style -> background property, so it can be customised from the client code if needed.

The previous discussion can be found at https://github.com/novnc/noVNC/pull/1189

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -31,6 +31,7 @@ import TightPNGDecoder from "./decoders/tightpng.js";
 
 // How many seconds to wait for a disconnect to finish
 const DISCONNECT_TIMEOUT = 3;
+const DEFAULT_BACKGROUND = 'rgb(40, 40, 40)';
 
 export default class RFB extends EventTargetMixin {
     constructor(target, url, options) {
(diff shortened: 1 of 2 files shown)


CendioOssman:
Looks mostly okay to me. I'd just like to see some tweaks of the terms. We've tried to avoid referring to the screen or canvas in the documentation, and instead present the RFB object as self contained.

So perhaps just call it `background`? And document it as setting the [background](https://developer.mozilla.org/en-US/docs/Web/CSS/background) CSS style for the area not covered by the remote session.

mykola-mokhnach:
@CendioOssman Sure, will do

CendioOssman:
The link got the tags backwards, so it didn't parse correctly. :)

Almost there. Perhaps bike shedding, but I'd really like to avoid mentioning the canvas. The rest of the document refers to the displayed image as the remote session.  So less details about how this is technically implemented and more the visible end result. We don't want people assuming a specific internal architecture as we might change that.

mykola-mokhnach:
Updated the docstring

CendioOssman:
Thanks. Looks good to me as well. Could you merge all the commits in to one and force update this PR?

mykola-mokhnach:
Yes, will do. BTW you could also squash on merge. Github does provide such functionality automatically

samhed:
> BTW you could also squash on merge. Github does provide such functionality automatically

There's one small difference if we do this - we won't get the merge commit.

PR #1204, opened 2019-02-17
Small improvements

juanjoDiaz:
I was looking at some stuff in the code base and here are few small improvements:

* Remove intermediate variable from mouse: so it's the same as in keyboard, consistency)
* Simplify keysymdef fallback
* Remove unnecessary constructor parameter from Cursor
* Remove unnecessary context from eventtarget: Since it uses an fat arrow function, it's prebound and the explicit context has no effect.
* Use default argument for base64: so it's more concise and readable. We discussed in a previous PR that default arguments are OK.
* Convert DES into a class: It was weird to have a function with a bunch of nested functions that returns an object and it's called with `new`. Now it looks simpler to me. Also extracted the static data out of it so it doesn't create a new copy on every instantiation. 
* Move support check from display to browser: For consistency with other browser support checks.
* Move websocket and display to decoder constructors: to simplify the class an avoid passing the objects to each and every method.
* Get rid of callbacks from the app in favour of promises: promises are already polyfilled for IE11 since they are used by noNVC's core.

There is one commit per change.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -39,22 +39,20 @@ const UI = {
     reconnect_callback: null,
     reconnect_password: null,
 
-    prime(callback) {
-        if (document.readyState === "interactive" || document.readyState === "complete") {
-            UI.load(callback);
-        } else {
-            document.addEventListener('DOMContentLoaded', UI.load.bind(UI, callback));
-        }
-    },
+    prime() {
+        return WebUtil.initSettings().then(() => {
+            if (document.readyState === "interactive" || document.readyState === "complete") {
+                return UI.start();
+            }
 
-    // Setup rfb object, load settings from browser storage, then call
-    // UI.init to setup the UI/menus
-    load(callback) {
-        WebUtil.initSettings(UI.start, callback);
+            return new Promise((resolve, reject) => {
+                document.addEventListener('DOMContentLoaded', () => UI.start().then(resolve).catch(reject));
+            });
+        });
     },
 
     // Render default UI and initialize settings menu
-    start(callback) {
+    start() {
 
         UI.initSettings();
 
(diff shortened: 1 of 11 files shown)


app/ui.js
```diff
@@ -39,22 +39,20 @@ const UI = {
     reconnect_callback: null,
     reconnect_password: null,
 
-    prime(callback) {
+    prime() {
         if (document.readyState === "interactive" || document.readyState === "complete") {
-            UI.load(callback);
-        } else {
-            document.addEventListener('DOMContentLoaded', UI.load.bind(UI, callback));
+            return WebUtil.initSettings().then(UI.start);
```
CendioOssman: Some duplication of `WebUtil.initSettings()` that might be possible to chain away. But otherwise it looks very nice. :)

core/base64.js
```diff
@@ -56,9 +56,7 @@ export default {
     ],
     /* eslint-enable comma-spacing */
 
-    decode(data, offset) {
-        offset = typeof(offset) !== 'undefined' ? offset : 0;
-
+    decode(data, offset = 0) {
```
samhed: Default arguments are not supported in internet explorer. I expect that this gives an error in IE

samhed: Perhaps babel handles this?

juanjoDiaz: Yeah. Babel handles this.
There was a similar discussion in some other PR and there are already some default arguments in the codebase.

core/util/browser.js
```diff
@@ -25,29 +25,32 @@ window.addEventListener('touchstart', function onFirstTouch() {
...
-                _cursor_uris_supported = true;
-            } else {
-                Log.Warn("Data URI scheme cursor not supported");
-                _cursor_uris_supported = false;
-            }
-        } catch (exc) {
-            Log.Error("Data URI scheme cursor test exception: " + exc);
-            _cursor_uris_supported = false;
-        }
+let _supportsCursorURIs = false;
```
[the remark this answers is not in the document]
juanjoDiaz: The variable was here before I touch this but I assume that's for performance.
It just caches the result of the check so it's not rechecked every time the function is called.

juanjoDiaz:
All the comments should have been resolved now.

PR #1206, opened 2019-02-17
Add support for custom logger

### LINKED ISSUE #243: Making logging more configurable
At the moment, the logging functionality in NoVNC is in a bit of an odd place, because it's defined in Util, which is sort of a stateless namespacey object, rather than one created from a (parametrized) constructor like `RFB`, for example. But logging does have state (currently just the log level, but I would like to parametrize it so it can log to other objects than `console`, so it would need to store a target as well).

Currently you work around this by calling `init_logging` immediately after defining the function, so while the logger is in theory configurable, it configures itself implicitly as soon as it's defined.

Of course, I could make a hack to work around this in a dozen different ways, but I'm thinking it would make sense if the logger was explicitly instantiated (probably at the top of `ui.js`, before including all the other files. And for clients which dont use `ui.js`, we can instantiate the logger on our own, and pass custom parameters to it)

I'm wondering how you would like such a change to be structured, though.

I can see basically two approaches:
1. Util is turned into a "proper" object structured like, say, `RFB`, with a constructor and everything. So you have to do something like `window.util = new Util()` in `ui.js` to initialize the object and make it globally visible.
2. Util is left as a "namespace", but the logging functionality is moved out into a separate object, which is given the treatment described in (1).

Of course, if I go with option 1, a side question is about name clashes, since the code already refers to `Util` all over the place and expects to get the object, and not the constructor. Should the global variable keep the name `Util`, or be renamed to `util` (which would require a changes all over NoVNC) or should the constructor `Util` be renamed to something else? (`window.Util = new Utility()` perhaps?)

I've experimented with this a bit already, but I don't want to put too much work into it until I know which approach you'd prefer, so my pull request actually has a chance of being accepted. :)

Thoughts?

juanjoDiaz:
Closes #243 

This PR establishes a new `Logger` interface:
```ts
interface Logger {
  debug(...args) {}
  info(...args) {}
  warn(...args) {}
  error(...args) {}
}
```

and create a static setter in the RBF object (`setLogger`) that can receive any instance of such interface and set it as generic logger for all noVNC related classes.

In `logging.js` the are to example implementation of such interface: `NoopLogger` and `ConsoleLogger`.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -7,7 +7,7 @@
  *
  */
 
-import { Log, setLogger } from './util/logging.js';
+import { log, setLogger } from './util/logging.js';
 import { decodeUTF8 } from './util/strings.js';
 import { dragThreshold } from './util/browser.js';
 import EventTargetMixin from './util/eventtarget.js';
@@ -121,7 +121,7 @@ export default class RFB extends EventTargetMixin {
         };
 
         // main setup
-        Log.Debug(">> RFB.constructor");
+        log.debug(">> RFB.constructor");
 
         // Create DOM elements
         this._screen = document.createElement('div');
@@ -168,7 +168,7 @@ export default class RFB extends EventTargetMixin {
         try {
             this._display = new Display(this._canvas);
         } catch (exc) {
-            Log.Error("Display exception: " + exc);
+            log.error("Display exception: " + exc);
             throw exc;
         }
         this._display.onflush = this._onFlush.bind(this);
@@ -189,14 +189,14 @@ export default class RFB extends EventTargetMixin {
             if ((this._rfb_connection_state === 'connecting') &&
                 (this._rfb_init_state === '')) {
                 this._rfb_init_state = 'ProtocolVersion';
-                Log.Debug("Starting VNC handshake");
+                log.debug("Starting VNC handshake");
             } else {
                 this._fail("Unexpected server connection while " +
                            this._rfb_connection_state);
             }
         });
         this._sock.on('close', (e) => {
-            Log.Debug("WebSocket on-close event");
+            log.debug("WebSocket on-close event");
             let msg = "";
             if (e.code) {
                 msg = "(code: " + e.code;
@@ -229,13 +229,13 @@ export default class RFB extends EventTargetMixin {
             }
             this._sock.off('close');
         });
-        this._sock.on('error', e => Log.Warn("WebSocket on-error event"));
+        this._sock.on('error', e => log.warn("WebSocket on-error event"));
 
         // Slight delay of the actual connection so that the caller has
         // time to set up callbacks
         setTimeout(this._updateConnectionState.bind(this, 'connecting'));
 
-        Log.Debug("<< RFB.constructor");
+        log.debug("<< RFB.constructor");
 
         // ===== PROPERTIES =====
 
@@ -328,7 +328,7 @@ export default class RFB extends EventTargetMixin {
 
     sendCtrlAltDel() {
         if (this._rfb_connection_state !== 'connected' || this._viewOnly) { return; }
-        Log.Info("Sending Ctrl-Alt-Del");
+        log.info("Sending Ctrl-Alt-Del");
 
         this.sendKey(KeyTable.XK_Control_L, "ControlLeft", true);
         this.sendKey(KeyTable.XK_Alt_L, "AltLeft", true);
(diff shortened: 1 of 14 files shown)


CendioOssman:
I'm having mixed feelings about this. I'd like to avoiding defining new interfaces unless we have to. And is it really any different that an `RFB` object always logs to the console, the same way a `WebSocket` object always does?

@samhed, thoughts?

samhed:
@juanjoDiaz could you explain the use case for this a bit?

juanjoDiaz:
The use case was described in #243.
Being able to choose how you log.
Maybe you want your logs to be printed at the console, maybe you want to add them to the actual page for the user to see, maybe you want to send them to some log server, etc.

Overall, I think that's a feature that adds configurability without any performance or maintainability penalty.

samhed:
Thanks for giving examples.

I'm fine with merging this.

core/rfb.js
```diff
@@ -310,6 +310,10 @@ export default class RFB extends EventTargetMixin {
 
     // ===== PUBLIC METHODS =====
 
+    static setLogger(logger) {
```
CendioOssman: This method needs to be documented as well.

core/util/logging.js
```diff
@@ -10,47 +10,69 @@
...
+};
+
+export class  NoopLogger {
+    Debug() {}
+    Info() {}
+    Warn() {}
+    Error() {}
+}
+
+export class ConsoleLogger {
```
CendioOssman: I'd rather not invent something new here. I think we should use the already standardised [console API](https://developer.mozilla.org/en-US/docs/Web/API/console) as the interface.

juanjoDiaz: Do you me the interface or this particular implementation?

I agree with the interface (it's just lowercasing the methods). But the console API is large. We probably shouldn't use the table, time, etc, methods.

If you refer to `ConsoleLogger` in particular. The good thing is that `ConsoleLogger` takes a log level and doesn't log stuff below that level. Unfortunately, you can't do that with the standard console API.

samhed: Hm, the downside to merging this would be to effectively restrict what kind of logging functions we can use in the internal modules, since we wouldn't want to break an external API once defined.

If we would use the console API we wouldn't at least limit ourselves functionality-wise compared to what's available today.

juanjoDiaz: Can you clarify a bit?

The interface that is defined is a subset of the console API i.e. it exposes the commonly used methods of the Console API but not all the methods.

```ts
interface Logger {
  debug(...args) {}
  info(...args) {}
  warn(...args) {}
  error(...args) {}
}
```

If you would like to use more methods from the console API (let's say table) you would just need to add ti and wouldn't break anything.


```ts
interface Logger {
  debug(...args) {}
  info(...args) {}
  warn(...args) {}
  error(...args) {}
  table(...args) {}
}
```

or did you mean something else?

samhed: We should be nice to people who include our library in their own projects. I mean let's say we initially say that our API requires `debug`, `info`, `warn` & `error`. We then in a later version want to use `table` in our internal modules - that would break the upgrade for everyone who used their own logging module. It would force them to also support `table`.

juanjoDiaz: I understand your point.

According to MDN, not including non-standard methods we have:
```
assert()
clear()
count()
countReset()
dir()
dirxml()
error()
group()
groupCollapsed()
groupEnd()
info()
log()
table()
time()
timeEnd()
timeLog()
trace()
warn()
```

So we are missing:
```
assert()
clear()
count()
countReset()
dir()
dirxml()
group()
groupCollapsed()
groupEnd()
log()
table()
time()
timeEnd()
timeLog()
trace()
```

Some of those are not widely supported. For example, `countReset()` or `table()` are not supported by IE11, `timeLog()` is only supported in Firefox, etc.

So the question is: Which of those should be supported?
Does it make sense to support every method?
What happens when Console API add new methods?
Current methods (+ trace) easily work with a logging level. But all the others don't. So how would that work?

samhed: Generally I guess we wouldn't consider using a function that wasn't supported by all the major browsers.
But yeah I'm starting to realize that I'm probably changing my mind here. Somewhere this feature will grow to a point where the work and maintenance required exceeds the value. I didn't really research what I was asking for.

Also, given @CendioOssman 's intial scepticism I'm leaning towards saying no to this feature. @CendioOssman is MIA for ~1 month at the moment, so it's not much point in waiting for him to give more input.

PR #1209, opened 2019-02-19
Add Russian translation

shweew:
That's all right now.


shweew:
For some reason did not pass the Node.js test: 6
TEST_BROWSER_NAME=firefox TEST_BROWSER_OS='Windows 10' 
Under Windows 10 and Firefox 65.0.1 64 bit perfectly works for me.

samhed:
Thanks for the PR! I restarted the tests, works fine now. Sometimes a few tests timeout like that.

shweew:
Correction of grammatical errors in ru_RU files

shweew:
PR is not suitable for commit?

po/noVNC.pot
```diff
@@ -6,9 +6,9 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: noVNC 1.1.0\n"
+"Project-Id-Version: noVNC 1.0.0\n"
```
samhed: It should be 1.1.0 since that's the upcoming release.

app/locale/ru_RU.json
```diff
@@ -0,0 +1,73 @@
...
+    "Automatic Reconnect": "Автоматическое переподключение",
+    "Reconnect Delay (ms):": "Задержка переподключения (мс):",
+    "Show Dot when No Cursor": "Показать точку вместо курсора",
+    "Logging:": "Лог:",
+    "Disconnect": "Отключение",
+    "Connect": "Подключение",
+    "Password:": "Пароль:",
+    "Send Password": "Пароль: ",
+    "Cancel": "Выход"
+}
```
samhed: You can skip this commit, updating the json files will be done as part of the release procedure.

po/Makefile
```diff
@@ -1,7 +1,7 @@
 all:
 .PHONY: update-po update-js update-pot
 
-LINGUAS := cs de el es ko nl pl sv tr zh_CN zh_TW
+LINGUAS := ru_RU cs de el es ko nl pl sv tr zh_CN zh_TW
```
samhed: this should simply be "ru" I think

po/noVNC.pot
shweew: OK

app/locale/ru_RU.json
shweew: Miss at merge.

po/Makefile
shweew: When just "ru" that does not work translation in the Russian locale.

po/ru_RU.po
```diff
@@ -0,0 +1,306 @@
...
+msgid "Esc"
+msgstr "Esc"
+
+#: ../vnc.html:143
+msgid "Send Escape"
+msgstr "Передать нажатие Escape"
+
+#: ../vnc.html:146
+msgid "Ctrl+Alt+Del"
+msgstr Ctrl+Alt+Del"
```
samhed: Missing a string marker here

po/Makefile
samhed: I checked out your code and tested here, and I can't reproduce your problem. For me it's the other way around..

Our code is written so that if the browser asks for a specific sub-language like 'ru_RU' and no such specific translation exists, it will fallback to use 'ru'. If we name the translation like 'ru_RU' and the browser asks for 'ru' we will not get a translation.

app/ui.js
```diff
@@ -1644,7 +1644,7 @@ const UI = {
 };
 
 // Set up translations
-const LINGUAS = ["cs", "de", "el", "es", "ko", "nl", "pl", "sv", "tr", "zh_CN", "zh_TW"];
+const LINGUAS = ["ru_RU", "cs", "de", "el", "es", "ko", "nl", "pl", "sv", "tr", "zh_CN", "zh_TW"];
```
samhed: Aside from changing this to 'ru' could you please also place it alphabetically in this list?

po/Makefile
samhed: Could you please also place it alphabetically in this list?

app/ui.js
shweew: Strange, I have everything on the contrary, tested in Firefox and Chrome, if "ru", then does not work.

samhed: If you open a console in your browser and type navigator.language, what do you get?

shweew:
[image: image.png]

пн, 25 февр. 2019 г. в 19:39, Samuel Mannehed <notifications@github.com>:

app/ui.js
shweew: image

shweew: Firefox in Linux:
image

po/ru_RU.po
shweew: Fix

app/ui.js
shweew: image

samhed: Ok. So that explains why the translations are working for you. Your `navigator.language` string is different from mine. But it doesn't explain why only `ru` in the code wouldn't work.

The issue here is that if we run your current code, people running Linux will not be getting russian translations in our browsers, since our `navigator.language` value is simply `ru`. Another issue is that anyone whose browser is configured to use a sub language of russian (for example Moldavian russian, `ru-md`) won't get the fallback to normal russian translations if our code is specified for `ru_RU`.

Are you sure that, when you have tried `ru`, you have made all the necessary changes?

 * rename the po-file to `ru.po`
 * change `ru_RU` to `ru` in `po/Makefile` and `app/ui.js`
 * go to `po/` and run `make update-pot` and then `make update-js`

It should work.

shweew: Make and PR

samhed:
That looks better, I assume it works with `ru` now? Thanks for having patience with me.

Other than that, it seems like a lot of unwanted changes slipped through. Essentially the only files we should be changing here are `po/ru.po`, 'po/Makefile` and `app/ui.js`.

Lastly, could I ask you to squash all of these changes into one commit? Then we should be ready for merge.

shweew:
Make and PR

samhed:
Merged #1211

PR #1218, opened 2019-03-14
Some formatting fixes for HTML files

### LINKED ISSUE #1214: vnc.html badly formatted
**Describe the bug**
Badly formatted vnc.html is causing issues when trying to embed novnc into web applications.

**To Reproduce**
1. Use your favorite text editor and open vnc.html
2. notice all the <link> tags have no closing </link> tag or don't end with />

**Expected behavior**
A properly formatted html document, maybe use a strict dtd reference?

 - OS: ALL
 - Browser: ALL
 - Browser version: ALL

samhed:
Fixes #1214 and other minor issues.

### DIFF
--- a/vnc.html
+++ b/vnc.html
@@ -18,34 +18,34 @@
     <meta charset="utf-8" />
 
     <!-- Icons (see app/images/icons/Makefile for what the sizes are for) -->
-    <link rel="icon" sizes="16x16" type="image/png" href="app/images/icons/novnc-16x16.png">
-    <link rel="icon" sizes="24x24" type="image/png" href="app/images/icons/novnc-24x24.png">
-    <link rel="icon" sizes="32x32" type="image/png" href="app/images/icons/novnc-32x32.png">
-    <link rel="icon" sizes="48x48" type="image/png" href="app/images/icons/novnc-48x48.png">
-    <link rel="icon" sizes="60x60" type="image/png" href="app/images/icons/novnc-60x60.png">
-    <link rel="icon" sizes="64x64" type="image/png" href="app/images/icons/novnc-64x64.png">
-    <link rel="icon" sizes="72x72" type="image/png" href="app/images/icons/novnc-72x72.png">
-    <link rel="icon" sizes="76x76" type="image/png" href="app/images/icons/novnc-76x76.png">
-    <link rel="icon" sizes="96x96" type="image/png" href="app/images/icons/novnc-96x96.png">
-    <link rel="icon" sizes="120x120" type="image/png" href="app/images/icons/novnc-120x120.png">
-    <link rel="icon" sizes="144x144" type="image/png" href="app/images/icons/novnc-144x144.png">
-    <link rel="icon" sizes="152x152" type="image/png" href="app/images/icons/novnc-152x152.png">
-    <link rel="icon" sizes="192x192" type="image/png" href="app/images/icons/novnc-192x192.png">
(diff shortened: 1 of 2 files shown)


vnc.html
```diff
@@ -152,7 +152,7 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br />VNC</h1>
             <div class="noVNC_vcenter">
             <div id="noVNC_power" class="noVNC_panel">
                 <div class="noVNC_heading">
-                    <img src="app/images/power.svg"> Power
+                    <img alt="" src="app/images/power.svg"> Power
```
CendioOssman: That's just silencing the error. We should have a proper text in there. :)

vnc.html
```diff
@@ -319,8 +319,7 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br />VNC</h1>
              on-screen keyboard. Let's hope Chrome implements the ime-mode
              style for example -->
         <textarea id="noVNC_keyboardinput" autocapitalize="off"
-            autocorrect="off" autocomplete="off" spellcheck="false"
-            mozactionhint="Enter" tabindex="-1"></textarea>
+            autocomplete="off" spellcheck="false" tabindex="-1"></textarea>
```
CendioOssman: I assume those attributes served a purpose at one point? Are we sure they're no longer needed.

vnc.html
```diff
@@ -18,34 +18,34 @@
...
-    <link rel="icon" sizes="60x60" type="image/png" href="app/images/icons/novnc-60x60.png">
-    <link rel="icon" sizes="64x64" type="image/png" href="app/images/icons/novnc-64x64.png">
-    <link rel="icon" sizes="72x72" type="image/png" href="app/images/icons/novnc-72x72.png">
-    <link rel="icon" sizes="76x76" type="image/png" href="app/images/icons/novnc-76x76.png">
-    <link rel="icon" sizes="96x96" type="image/png" href="app/images/icons/novnc-96x96.png">
-    <link rel="icon" sizes="120x120" type="image/png" href="app/images/icons/novnc-120x120.png">
-    <link rel="icon" sizes="144x144" type="image/png" href="app/images/icons/novnc-144x144.png">
-    <link rel="icon" sizes="152x152" type="image/png" href="app/images/icons/novnc-152x152.png">
-    <link rel="icon" sizes="192x192" type="image/png" href="app/images/icons/novnc-192x192.png">
+    <link rel="icon" sizes="16x16" type="image/png" href="app/images/icons/novnc-16x16.png" />
```
CendioOssman: We should be consistent the other way IMO. `/>` is a XHTML thing, that HTML happens to tolerate. The normal way is to not have a `/` and adding it might give the impression that this file is valid XHTML.

It's a shame you can't make the validator complain about this...

vnc.html
samhed: I should probably have given a better explanation, but in most of these cases the images are purely decorative icons that doesn't provide any added information. According to the HTML specification, an empty alt is the proper way to go here:

https://www.w3.org/TR/html51/semantics-embedded-content.html#icon-images

vnc.html
samhed: * mozactionhint - firefox only, the default is the enter-symbol anyway
* autocorrect - safari extension, has no effect now-a-days

vnc.html
samhed: I'm fine either way.

CendioOssman: Adjusted in 2aa3b5bc79a6ae33be93985469be27a6b787176f.

PR #1220, opened 2019-04-02
Restore token forwarding in vnc_lite

melwitt:
This is a partial revert of commit
51f9f0098d306bbc67cc8e02ae547921b6f6585c which intended to remove
unnecessary options.

In OpenStack Nova, we use the vnc_lite (formerly vnc_auto)
functionality as a default in our novnc proxy [1] and the proxy relies
on token validation. We need to be able to access the token in the URL
in our websockify proxy server.

Websockify itself also offers a token plugin feature that requires
forwarding of the token [2]. Restoring token forwarding allows users to
continue using token plugins with vnc_lite.

[1]
https://docs.openstack.org/nova/latest/admin/remote-console-access.html#novnc-based-vnc-console
[2]
https://github.com/novnc/websockify/wiki/Token-based-target-selection

### DIFF
--- a/vnc_lite.html
+++ b/vnc_lite.html
@@ -154,6 +154,14 @@
         const password = readQueryVariable('password', '');
         const path = readQueryVariable('path', 'websockify');
 
+        // If a token variable is passed in, forward it in the URL so that it
+        // is accessible by a websockify token plugin
+        var token = WebUtil.getConfigVar('token', null);
+        if (token) {
+            // if token is already present in the path we should use it
+            path = WebUtil.injectParamIfMissing(path, "token", token);
+        }
+
         // | | |         | | |
         // | | | Connect | | |
         // v v v         v v v


CendioOssman:
Sorry, but I'm afraid this is not something we want to merge. The point of `vnc_lite.html` is to be a very simple example, not to have lots of features. We also explicitly wanted to get rid of `token` as it is not a standard feature but rather something specific to websockify.

You can achieve the same thing using the `path` variable though, which is still present.

mnaser:
This breaks the OpenStack nova project which depends on tokens. 

The URL with path is quite odd looking too.

samhed:
OpenStack nova has to adjust their application then.

mnaser:
Does that mean we have to build and use our own `vnc.html` ? I'm just not sure how that would be possible.. or by any other means.

kk7ds:
Passing the token in the path variable as @CendioOssman said seems like it should work okay and with minimal change on our side, unless I'm missing something.

mnaser:
The URLs we'll end up with will look like:

https://foo.com/vnc_lite.html?path=?token=foo

I'm not *super* opposed it but I also think that seems very.. hacky 🤷‍♂️

kk7ds:
Actually it will be:

https://foo.com/vnc_lite.html?path=%3Ftoken%3Dfoo

Which doesn't seem as bad to me.

CendioOssman:
I would recommend you look at moving to the standard `vnc.html` long term though. It is there all the features are being added to make a good user experience. If there are features missing then it is something we can discuss adding. Please let us know on the mailing list or via feature request issues here.

PR #1226, opened 2019-04-14
Add pass through for token

PexMor:
Proposing to add pass through for token. That let us create links embedding the token that gets pushed to the websocket for authentication connection.


PexMor:
Duplicate of other pull request

PR #1231, opened 2019-04-24
Added a Snap package for noVNC

tkedwards:
I've added a Snap package to make noVNC easy to install and run as a service. This involves 3 changes:
- to README.md to add a docs section for the Snap package
- the addition of snap/snapcraft.yaml which contains everything needed to generate the Snap package
- added extra logic for finding the WEBSOCKIFY binary. This is due to the fact that although the Snap package uses the Ubuntu 'websockify' package the binary it installs (inside the Snap confinement directory) is usr/bin/python2-websockify. This is because the update-alternatives system that creates usr/bin/websockify doesn't run in DEB packages used as part of a Snap

If you accept this change there's one more thing - someone needs to submit the Snap package to the store. I can do it, but the store's docs say that the project team should usually be the ones to do it: https://docs.snapcraft.io/releasing-to-the-snap-store/6848

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,9 @@ Create a new service that listens on port 6082 and connects to the VNC server ru
 
 `sudo snap set novnc services.n6082.listen=6082 services.n6082.vnc=localhost:5902`
 
+(Any services you define with 'snap set' will be automatically started)
+Note that the name of the service, 'n6082' in this example, can be anything as long as it doesn't start with a number or contain spaces/special characters.
+
 View the configuration of the service just created:
 
 ```
@@ -154,6 +157,8 @@ services.n6082.vnc     localhost:5902
 Disable a service (note that because of a limitation in  Snap it's currently not possible to unset config variables, setting them to blank values is the way to disable a service):
 
 `sudo snap set novnc services.n6082.listen='' services.n6082.vnc=''`
+
+(Any services you set to blank with 'snap set' like this will be automatically stopped)
 
 Verify that the service is disabled (blank values):
 
--- a/README.md
+++ b/README.md
@@ -115,6 +115,13 @@ proxy.
   script. Hit the Connect button, enter a password if the VNC server has one
   configured, and enjoy!
 
+### Installation from Snap Package
+To install from Snap:
+    `sudo snap install novnc`
+By default the Snap package creates 6 services on different ports which allow access to the local VNC server. So novnc.n5900 listens on port 6080 and connects to the VNC services on localhost:5901, novnc.n5901 listens on port 6081 and connects to localhost:5901 and so on up to port 6085 and localhost:5905. These services can be started/stopped with the usual `snap service` commands, e.g: `sudo snap stop novnc.n5901`
+
+If you stop the services noVNC can then be run manually with:
+    `novnc --vnc localhost:5901 # /snap/bin/novnc if /snap/bin is not in your PATH`
 
 ### Integration and Deployment
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ for a more complete list with additional info and links.
 - [Browser Requirements](#browser-requirements)
 - [Server Requirements](#server-requirements)
 - [Quick Start](#quick-start)
-- [Installation from Snap Package](#snap-package)
+- [Installation from Snap Package](#installation-from-snap-package)
 - [Integration and Deployment](#integration-and-deployment)
 - [Authors/Contributors](#authorscontributors)
 
@@ -119,7 +119,7 @@ proxy.
 ### Installation from Snap Package
 To install from Snap:
 
-    `sudo snap install novnc`
+`sudo snap install novnc`
 
 By default the Snap package creates 6 services on different ports which allow access to the local VNC server. So novnc.n5900 listens on port 6080 and connects to the VNC services on localhost:5901, novnc.n5901 listens on port 6081 and connects to localhost:5901 and so on up to port 6085 and localhost:5905. These services can be started/stopped with the usual `snap service` commands, e.g: 
 
@@ -127,7 +127,7 @@ By default the Snap package creates 6 services on different ports which allow ac
 
 If you stop the services noVNC can then be run manually with:
 
-    `novnc --vnc localhost:5901 # /snap/bin/novnc if /snap/bin is not in your PATH`
+`novnc --vnc localhost:5901 # /snap/bin/novnc if /snap/bin is not in your PATH`
 
 ### Integration and Deployment
 
--- a/README.md
+++ b/README.md
@@ -123,7 +123,12 @@ To install from Snap:
 
 By default the Snap package creates 6 services on different ports which allow access to the local VNC server. So novnc.n5900 listens on port 6080 and connects to the VNC services on localhost:5901, novnc.n5901 listens on port 6081 and connects to localhost:5901 and so on up to port 6085 and localhost:5905. These services can be started/stopped with the usual `snap service` commands, e.g: 
 
-`sudo snap stop novnc.n5901`
+```
+snap services novnc # list services
+sudo snap stop novnc.n5901 # stop the services that listens on 6081 connecting to localhost:5901
+sudo snap start novnc.n5901 # start it
+sudo snap restart novnc.n5901 # restart it
+```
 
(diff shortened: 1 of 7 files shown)


utils/launch.sh
```diff
@@ -128,9 +128,12 @@ if [[ -e ${HERE}/websockify ]]; then
 
     echo "Using local websockify at $WEBSOCKIFY"
 else
-    WEBSOCKIFY=$(which websockify 2>/dev/null)
+    WEBSOCKIFY_FROMSYSTEM=$(which websockify 2>/dev/null)
+    WEBSOCKIFY_FROMSNAP=${HERE}/../usr/bin/python2-websockify
+    [ -f $WEBSOCKIFY_FROMSYSTEM ] &&  WEBSOCKIFY=$WEBSOCKIFY_FROMSYSTEM
```
samhed: Remove the extra space here.

tkedwards:
So the issue with the services is that if they're not defined in the Snap package there's no way to add them (unless the user does it themselves by manually creating a systemd unit file). Having 6 is arbitrary but the idea is that it'd cover almost all use cases, e.g. I run VNC on 5901 on my virtual server, I know many people use 5900, and I'd think there's very few that would run more than 3 or 4 let alone 6. 

I'm not an expert in Snap packaging either, but from what I could see in their docs and forums comments this might be the only way... I'll try a Snapcraft forum post and see if anyone has a better suggestion

tkedwards:
Snap devs suggested I use Snap's built-in config interface so users can define which port(s) they want to run the service on. I'll look into this and update this branch when I get it working...

samhed:
> Snap devs suggested I use Snap's built-in config interface so users can define which port(s) they want to run the service on. I'll look into this and update this branch when I get it working...

That sounds great! Thank you for your effort.

CendioOssman:
I agree with Samuel, having the servers hard coded is very weird. Let's see what is possible with the configuration interface and go from there. :)

tkedwards:
I got a configurable service (including the ability to run multiple novnc instances under the one Snap 'service') working! I've updated the README.md with full instructions on how to use it. Let me know what you guys think

tkedwards:
Just fixed up the source: line to point at the main novnc repo (instead of current directory when I was building the snap locally). Only step to do if you guys merge this in is for someone to create an account on the Snap store and upload the snap there: https://docs.snapcraft.io/releasing-your-app

not sure if you guys want to do it, as it's your project, otherwise I can

CendioOssman:
Looks promising so far. Nice work.

1. I'm not too familiar with snap, so there are some unclear points though. The hooks are using `novncsvc.listen_port`, but the configuration seems to be in the form of `services.<name>.listen_port`. Was this just overlooked?

2. I guess `n6082` is just a name? Perhaps it would be more obvious if that is `ServerA` or something like that?

3. I assume some service needs to be restarted after changing the service configuration? Perhaps that should be mentioned in the documentation?

tkedwards:
Thanks! 
Answers to your questions:



> I'm not too familiar with snap, so there are some unclear points though. The hooks are using novncsvc.listen_port, but the configuration seems to be in the form of services.<name>.listen_port. Was this just overlooked?

Yep, I've deleted the redundant lines and the unused 'install' hook

> I guess n6082 is just a name? Perhaps it would be more obvious if that is ServerA or something like that?

Correct, I've now updated the docs saying you can name it what you like (as long as it's one word and starts with a letter, not a number)

> I assume some service needs to be restarted after changing the service configuration? Perhaps that should be mentioned in the documentation?

No, the service restart is automatic every time the configuration changes. I've updated the docs to mention that

tkedwards:
Hi guys,

Any further questions on this? Or do you think it looks good to merge?

CendioOssman:
Sorry, things are a bit hectic at the moment. Please give me a few more days and I hope to get back to this.

tkedwards:
No worries, I know how it is. Just asking as I'm about to go away for a couple of weeks. Take your time if there's anything I'll look at it later

README.md
```diff
@@ -115,6 +116,58 @@ proxy.
   script. Hit the Connect button, enter a password if the VNC server has one
   configured, and enjoy!
 
+### Installation from Snap Package
+To install from Snap:
+
+`sudo snap install novnc`
```
CendioOssman: I'm a bit cautious about this. The rest of the README is for running your current working copy. Doing an install like this would mean running the latest stable, not what you have right now.

What is the process for building your own snap package and using that? Could we have that, and suggest this line if you just want the latest stable version?

README.md
```diff
@@ -115,6 +116,58 @@ proxy.
...
+
+#### Running noVNC
+
+You can run the Snap-package installed novnc directly with, for example:
+`novnc --listen 6081 --vnc localhost:5901 # /snap/bin/novnc if /snap/bin is not in your PATH`
+
+#### Running as a Service (Daemon)
+The Snap package also has the capability to run a 'novnc' service which can be configured to listen on multiple ports connecting to multiple VNC servers (effectively a service runing multiple instances of novnc). Instructions (with example values):
```
CendioOssman: We've tried to keep the lines over 80 columns to keep it sane in consoles. Could you break your lines to that as well?

Exceptions are allowed when breaking them isn't really feasible, e.g. an example command line.

README.md
```diff
@@ -115,6 +116,58 @@ proxy.
...
 
+### Installation from Snap Package
+To install from Snap:
+
+`sudo snap install novnc`
+
+#### Running noVNC
+
+You can run the Snap-package installed novnc directly with, for example:
+`novnc --listen 6081 --vnc localhost:5901 # /snap/bin/novnc if /snap/bin is not in your PATH`
```
CendioOssman: Should perhaps be its own paragraph?

README.md
tkedwards: Should be fixed in latest commit

README.md
tkedwards: Should be fixed in latest commit

README.md
tkedwards: I'm not really sure what you mean here - the Snap package isn't meant to be for users to build*. The idea is that by adding in the snap/ directory (metadata/scripts for the snap package itself) plus the small changes I had to make to stuff under utils/ you have everything needed to submit the Snap package to Canonical's Snap store. Then anyone can install noVNC with just a 'snap install novnc'. There shouldn't be any major changes needed to the stuff under snap/ as you make further releases of noVNC

* technically you, or anyone else, can build your own copy of the Snap package and use it like I have been for testing. Note that you have to use the --dangerous switch to install it, this is something Canonical deliberately put in Snap to encourage people not to build and host their own .snap packages but to submit them to the store. Build instructions:

- git clone https://github.com/tkedwards/noVNC.git # the correct URL will be https://github.com/novnc/noVNC.git once everything is merged in
- cd noVNC/
- git checkout snap_package
- edit parts-novnc-source in snap/snapcraft.yaml to specify https://github.com/tkedwards/noVNC.git as the source (this step shouldn't be necessary once everything's merged in)
- also in snap/snapcraft.yaml add source-branch: snap_package (this shouldn't be necessary once this is all merged)
- snapcraft snap # add --use-lxd if on hardware which doesn't have the VT-x virtualisation extensions (e.g. a VPS)
- install the snap with sudo snap install --dangerous novnc_1.1.0_amd64.snap

CendioOssman: > I'm not really sure what you mean here - the Snap package isn't meant to be for users to build*.

Most of the time that is indeed true. The problem is that right now `README.md` describes how to run *this specific* copy of noVNC. How to run the latest stable is currently something we only refer to on the home page.

Perhaps we need to have an "Installation" section in the README as well. Or if it is sufficient to document the snap stuff on just the home page.

tkedwards: Yes, it would be good to have those instructions on the home page (after these changes have been merged if you approve and the novnc Snap submitted to Canonical's Snap 'store').

tkedwards: Hi, 

So is there something else I should add to the README in this pull request?

samhed: You should clarify in the README that it's the latest stable version they will install, just something simple like this:

> Running the command below will install the latest release of noVNC from Snap:

tkedwards: All good, I've updated the README, let me know if anything else

samhed: Great! Thank you @tkedwards

PR #1236, opened 2019-05-03
Fullscreen from iframe

jan-jockusch:
The request for this feature comes from making a page consisting of several `<iframe>`s, each with a noVNC viewer inside it.

In this setup, the fullscreen buttons do not work as expected, they just do nothing. 

This is because the `ui.js` implementation of `toggleFullscreen()` has no awareness for the fullscreen window actually being the parent window of the viewer.

Making the fullscreen buttons work is actually quite straightforward, except for the identification of the correct `<iframe>` in the parent window.

To solve the problem, I read a request variable, `iframeid`, and use that to select the correct element to put into fullscreen mode.

Testing the feature:

- build a simple page containing a viewer in an `<iframe>`,
- give the iframe an arbitrary `id` attribute,
- pass that value to the viewer as URL parameter `iframeid`,
- click the fullscreen button in the `<iframe>` — the viewer should go fullscreen,
- click the fullscreen button again — the viewer should jump back into its cage.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1168,10 +1168,15 @@ const UI = {
  * ------v------*/
 
     toggleFullscreen() {
+        // Determine the document using fullscreen. This may either be
+        // just the "document", but if using the viewer in an iframe, you need
+        // to use the parent window's "document" instead.
         let doc = document;
         if (window.self !== window.top) {
             doc = window.parent.document;
         }
+        // Check the fullscreen status using the correct document as
+        // a reference. The same document is then used to exit fullscreen.
         if (doc.fullscreenElement || // alternative standard method
             doc.mozFullScreenElement || // currently working methods
             doc.webkitFullscreenElement ||
@@ -1186,8 +1191,16 @@ const UI = {
                 doc.msExitFullscreen();
             }
         } else {
+            // To activate fullscreen, we need to find the correct document
+            // element, which is usually "document.documentElement". But when
+            // using the viewer in an iframe, this is actually the iframe
+            // element itself in the parent document.
             let doc_el = document.documentElement;
             if (window.self !== window.top) {
+                // Seek out the correct iframe from the parent document.
+                // This will produce errors if the iframe does not come from
+                // the same origin, so it is advisable to put the viewer in
+                // iframes only if you are on the same server.
                 let iframes = window.parent.document
                     .getElementsByTagName('iframe');
                 for (let i in iframes) {
@@ -1197,6 +1210,8 @@ const UI = {
                     }
                 }
             }
+            // Activate fullscreen. All except MS use the document element for
+            // this. The behaviour of body.msRequestFullscreen is untested.
             if (doc_el.requestFullscreen) {
                 doc_el.requestFullscreen();
             } else if (doc_el.mozRequestFullScreen) {
--- a/app/ui.js
+++ b/app/ui.js
@@ -1206,6 +1206,11 @@ const UI = {
                 for (let i in iframes) {
                     if (iframes[i].contentDocument === document) {
                         doc_el = iframes[i];
+                        // To use .body.msRequestFullscreen in IE, we need to
+                        // set the document element accordingly.
+                        // Note that the <iframe> must have the attribute
+                        // "allowfullscreen" set for IE to allow the feature.
+                        doc = iframes[i].contentDocument;
                         break;
                     }
                 }
--- a/app/ui.js
+++ b/app/ui.js
@@ -1168,7 +1168,7 @@ const UI = {
  * ------v------*/
 
     toggleFullscreen() {
-        var doc = document;
+        let doc = document;
         if (window.self !== window.top) {
             doc = window.parent.document;
         }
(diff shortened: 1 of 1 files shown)


jan-jockusch:
The URL parameter is unnecessary now. While correcting the linter errors, I added the according detection. 

Test now runs as follows:

- build a simple page containing a viewer in an `<iframe>,`
- click the fullscreen button in the `<iframe>` — the viewer should go fullscreen,
- click the fullscreen button again — the viewer should jump back into its cage.

If someone can point me at a way to write a travis test for this, I'd be glad to add that, too.

jan-jockusch:
Comments added. About the testing: I had a go in Firefox and Chrome, but not yet on Safari, IE or Edge. I'll need to test the MS products, because they use a different reference object in their API.

jan-jockusch:
OK: Edge has no problems.

IE doesn't seem to know a lot about fullscreen, it just removes the top bar, so this can be considered "not working".

Safari doesn't even offer the fullscreen option in iframe mode. The patch needs to extend to the detection of fullscreen capability for this browser. I'll look into it as soon as I can, but that will have to wait until Monday.

jan-jockusch:
Safari: Fullscreen is already disabled in "initFullscreen", because of trouble with alphanumeric input. So there's no additional issue there.

IE: With the latest patch, IE is confirmed fully working now. `<iframe allowfullscreen>` must be used.

jan-jockusch:
From my point of view the requested changes are there.

As for testing: Safari is out of the run, because the fullscreen feature is disabled for this browser in the master anyway. Edge, IE, Firefox, and Chrome all work.

Do I have to do anything to remove the red flag "Changes requested"?

jan-jockusch:
@samhed: Feel free to re-review anytime. I am done with the patch (and the comments).

app/ui.js
```diff
@@ -1168,28 +1168,63 @@ const UI = {
...
+            // To activate fullscreen, we need to find the correct document
+            // element, which is usually "document.documentElement". But when
+            // using the viewer in an iframe, this is actually the iframe
+            // element itself in the parent document.
+            let doc_el = document.documentElement;
+            if (window.self !== window.top) {
+                // Seek out the correct iframe from the parent document.
+                // This will produce errors if the iframe does not come from
+                // the same origin, so it is advisable to put the viewer in
+                // iframes only if you are on the same server.
```
samhed: What kind of errors? Would it be possible to do a try-catch and write a helpful error message in these cases?

jan-jockusch: Thanks for reminding me of that thing, I'm sorry I didn't fix this earlier. The comment was there as a reminder to myself that inspecting iframes just like that might not sit well with some browsers. I just put it there and forgot about it out of laziness.

The test is simple: I added an `<iframe src="https://www.perfact.de">` (shameless plug for my company) before any viewer iframes, so it would come first in the loop, forcing the browser to evaluate "contentDocument" from a foreign source.

Actually, Firefox and Chrome don't care that much (yet), but Edge and IE both give an error.

The fix for that is to wrap only the incriminating "contentDocument" lookup into a "try" block and ignoring any iframes we don't have permission for.

I corrected the comments accordingly and repeated all tests. Now I'm waiting for the travis run, and then I think we're good.

That is, if you don't find other bits where I've been sloppy…

jan-jockusch:
@samhed, @CendioOssman: Can you please clarify if the "nested frames" idea mentioned above is stopping the merge? 

I looked into it, and it seems the code already supports that, given that you set the `allowfullscreen` attribute in the iframes. I tested three levels of nesting.

If there's anything else I can do to make this branch join the master, please let me know.

samhed:
No, it's fine. Thanks for your contribution, you have been very responsive and understanding!

Cheers

jan-jockusch:
Thanks for merging!

I hope to be back soon with a feature addition to websockify. See you then!

CendioOssman:
> @samhed, @CendioOssman: Can you please clarify if the "nested frames" idea mentioned above is stopping the merge?

Putting an iframe inside an iframe (inside an iframe, etc.).

I just tested that and I'm afraid it does not work. Could you see if that is possible and do a follow-up PR?

jan-jockusch:
@CendioOssman: As I said above, I tested it and it works perfectly for the browsers Firefox, Chrome, IE, and Edge.

My test is as follows:

- Define three web pages, one with an `<iframe allowfullscreen >` containing the viewer.
- The other two contain an `<iframe allowfullscreen >` containing the previous page.

Please note the `allowfullscreen` attribute. If this is not present, all browsers will block the fullscreen request (some of them with a note in the console).

If you need proof on file, please clarify for me if and how I can build this test into Travis-CI.

But, I cannot do a follow-up PR if the nested iframes work perfectly for me. At the very least, I must understand what's the problem in your setup. So, if possible, also clarify what exactly you have tested.

CendioOssman:
Sorry, I missed the part about `allowfullscreen`. If I add that it works.

But! If I add that, I also don't need the changes in this PR. It works just fine with the original code.

So what scenario did we actually fix with this PR?

jan-jockusch:
@CendioOssman Hi and sorry for the late reply.

It is entirely possible that I just wrote some completely unnecessary code.

My only excuse is that the first tests were run without the `allowfullscreen` attribute (which I only found out about later during the tests, when most of the code was already there).

To confirm that, I'll re-run all tests with the original code as soon as I get a chance.

If I don't find any scenarios which need the patch, I would vote for rolling back this pull request, because unnecessary code shouldn't clutter up this very clean project. But before we rush to that, give me some days to confirm that.

CendioOssman:
No rush. Just let us know how your testing goes and we'll see after that.

CendioOssman:
@jan-jockusch, have you been able to test anything?

jan-jockusch:
Okay, I re-tested everything with the master branch before the merge.

And I feel really stupid right now.

Everything works as intended with no change whatsoever.

I didn't know about the "allowfullscreen" attribute until I reached the point where changes were already in place and tested (on Firefox and Chrome), and only ran into trouble when using Edge and IE.

@CendioOssman: Ah, well. You're very welcome to roll back the change. I'm sorry for the trouble I created. People just need to know they should set the attribute on iframes, and that's that.

samhed:
Thanks for verifying @jan-jockusch, and thanks for your work anyways. The changes were reverted in  32e0819

PR #1245, opened 2019-05-22
Add token setting support in latest version.

valia0906:
I`ve made and checked some changes for token support in new version.


CendioOssman:
I'm afraid explicit token support will not be coming back. See previous closed issues for details.

PR #1262, opened 2019-07-17
Add support in websocket sub-protocols

shiramax:


samhed:
Just curious - what are you using this for?

shiramax:
@samhed I need to open a vnc console for kubevirt VM, In order to do that, I have to authenticate with kubevirt-api and therefore I have to send bearer token.
there are two options to send this bearer token,
One way is via a header, However that will not work from browsers since they can't add headers when upgrading. The second option is to use a special subprotocol format to transfer the bearer token.

docs/EMBEDDING.md
```diff
@@ -67,6 +67,8 @@ query string. Currently the following options are available:
 * `logging` - The console log level. Can be one of `error`, `warn`, `info` or
   `debug`.
 
+* `wsProtocols` - websocket sub-protocols
```
samhed: I would like us to use proper capitalization, write 'WebSocket' please.

docs/EMBEDDING.md
```diff
@@ -67,6 +67,8 @@ query string. Currently the following options are available:
 * `logging` - The console log level. Can be one of `error`, `warn`, `info` or
   `debug`.
 
+* `wsProtocols` - WebSocket sub-protocols
```
CendioOssman: `EMBEDDING.md` is for `vnc.html`, which I can't see any changes for? So either omit this part, or make sure `vnc.html` can pass it on to the `RFB` constructor.

core/rfb.js
```diff
@@ -53,6 +53,7 @@ export default class RFB extends EventTargetMixin {
         this._shared = 'shared' in options ? !!options.shared : true;
         this._repeaterID = options.repeaterID || '';
         this._showDotCursor = options.showDotCursor || false;
+        this._wsProtocols = options.wsProtocols || ['binary'];
```
CendioOssman: Argh. Are we still including that `"binary"` default. That is very non-standard.

@samhed, could you remove that and just have nothing as the default once this is merged?

samhed: Sure

shiramax:
@CendioOssman @samhed thanks, could you please review again?

core/rfb.js
samhed: Fixed in e14aa4d0fe43a939378a3dd7666806fa7018dfe7

Issue #1222, opened 2019-04-09
[helpful] add discreet version number in side menu


pm-pm:
**Is your feature request related to a problem?
Not really.

**Describe the solution you'd like**
Please add a version number in side menu. (discreet would be fine) :-)

This is helpful when testing new versions in parallel. In the moment you can differentiate 1.0.0 vs. 1.1.0 easily by the new "window key icon" .

samhed:
It's not a bad idea. What if it showed when you hovered the noVNC logo in the side menu?

pm-pm:
Hello samhed,

that would be a cool Feature, but hovering is "to difficult" if you do (family) telephone support and ask "whats the number on your display." ;-) You have to explain where to put the mouse, etc. ... 

I would prefer something like this:
![nv110](https://user-images.githubusercontent.com/47995430/55838425-bd09d800-5b24-11e9-910f-3ee21ba095a9.png)

Of course you can decide what you like. :-)

Thanks

CendioOssman:
Maybe we should have an About dialog? We could have a button that opens in on the bar in that case.

samhed:
@pm-pm, I see what you mean. @CendioOssman yes, perhaps that would be the best solution.

pm-pm:
Just keep it simple. ;-)

samhed:
Fixed with #1265

samhed:
As a side note - we should probably mark development versions of noVNC as X.Y.Zpost or something similar. Since right now, if you learn to depend on that version number in the settings you can't differentiate the proper 1.1.0 release from the current master for example.

PR #1264, opened 2019-07-19
Add version number to UI

juanjoDiaz:
A very simple solution to #1222 that looks like the suggestion in the issue

### DIFF
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -564,6 +564,14 @@ select:active {
 
 #noVNC_control_bar .noVNC_logo {
   font-size: 13px;
+  margin-bottom: 0;
+}
+
+.noVNC_version {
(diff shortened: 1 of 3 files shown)


samhed:
noVNC_version

It looks decent, but I'd rather want something more hidden. Either if we hide it when connected, or put it in the settings somewhere. What do you think?

juanjoDiaz:
I've submitted an alternative proposal.

Btw, why do your sidebar has so many icons? Mine only has "Fullscreen" and "Settings".

samhed:
> Btw, why do your sidebar has so many icons? Mine only has "Fullscreen" and "Settings".

I'm connected.

samhed:
Closing in favor of #1265

PR #1269, opened 2019-07-29
QEMU Pointer Motion Change Pseudo-encoding

Ekstazi:
From [community docs](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#777qemu-pointer-motion-change-pseudo-encoding):

> ### 7.7.7   QEMU Pointer Motion Change Pseudo-encoding
> A client that supports this encoding declares that is able to send pointer motion events either as absolute coordinates, or relative deltas. The server can switch between different pointer motion modes by sending a rectangle with this encoding. If the x-position in the update is 1, the server is requesting absolute coordinates, which is the RFB protocol default when this encoding is not supported. If the x-position in the update is 0, the server is requesting relative deltas.
> 
> When relative delta mode is active, the semantics of the PointerEvent message are changed. The x-position and y-position fields are to be treated as S16 quantities, denoting the delta from the last position. A client can compute the signed deltas with the logic:
> 
> `uint16 dx = x + 0x7FFF - last_x`
> `uint16 dy = y + 0x7FFF - last_y`
> If the client needs to send an updated button-mask without any associated motion, it should use the value 0x7FFF in the x-position and y-position fields of the PointerEvent
> 
> Servers are advised to implement this pseudo-encoding if the virtual desktop is associated a input device that expects relative coordinates, for example, a virtual machine with a PS/2 mouse. Prior to this extension, a server with such a input device would have to perform the absolute to relative delta conversion itself. This can result in the client pointer hitting an "invisible wall".
> 
> Clients are advised that when generating events in relative pointer mode, they should grab and hide the local pointer. When the local pointer hits any edge of the client window, it should be warped back by 100 pixels. This ensures that continued movement of the user's input device will continue to generate relative deltas and thus avoid the "invisible wall" problem.

I implemented it.  But i doesn't have much knowledge to make autotests for it (I'm backender). But i tested it manually. Please verify and merge to upstream.

### DIFF
--- a/core/encodings.js
+++ b/core/encodings.js
@@ -19,6 +19,7 @@ export const encodings = {
     pseudoEncodingDesktopSize: -223,
     pseudoEncodingLastRect: -224,
     pseudoEncodingCursor: -239,
+    pseudoEncodingQEMUPointerMotionChange: -257,
     pseudoEncodingQEMUExtendedKeyEvent: -258,
     pseudoEncodingExtendedDesktopSize: -308,
     pseudoEncodingXvp: -309,
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -84,6 +84,9 @@ export default class RFB extends EventTargetMixin {
         this._screen_flags = 0;
 
         this._qemuExtKeyEventSupported = false;
+        this._qemuEnabledRelativeCursor = false;
+        this._lastMouseX = 0;
+        this._lastMouseY = 0;
 
         // Internal objects
         this._sock = null;              // Websock object
@@ -727,6 +730,14 @@ export default class RFB extends EventTargetMixin {
             this._mouse_buttonMask &= ~bmask;
         }
 
+        let mouseX = this._display.absX(x);
+        let mouseY = this._display.absY(y);
+
+        if (this._qemuEnabledRelativeCursor) {
+            mouseX += 0x7FFF - this._lastMouseX;
+            mouseY += 0x7FFF - this._lastMouseY;
+        }
+
         if (this.dragViewport) {
             if (down && !this._viewportDragging) {
                 this._viewportDragging = true;
@@ -748,8 +759,8 @@ export default class RFB extends EventTargetMixin {
                 // Send the button down event here, as the button up
                 // event is sent at the end of this function.
                 RFB.messages.pointerEvent(this._sock,
-                                          this._display.absX(x),
-                                          this._display.absY(y),
+                                          mouseX,
+                                          mouseY,
                                           bmask);
             }
         }
@@ -757,7 +768,7 @@ export default class RFB extends EventTargetMixin {
         if (this._viewOnly) { return; } // View only, skip mouse events
 
         if (this._rfb_connection_state !== 'connected') { return; }
-        RFB.messages.pointerEvent(this._sock, this._display.absX(x), this._display.absY(y), this._mouse_buttonMask);
+        RFB.messages.pointerEvent(this._sock, mouseX, mouseY, this._mouse_buttonMask);
     }
 
     _handleMouseMove(x, y) {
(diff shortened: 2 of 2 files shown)


samhed:
Thanks for the contribution. We'll try to find some time to look at this soon, but I'm quite swamped in the upcoming days.

Ekstazi:
Ok, thx!

CendioOssman:
Hmm... How can we test this? This extension doesn't seem to be possible to implement correctly in noVNC as the browser will not let us grab and manipulate the pointer.

Ekstazi:
@CendioOssman 
> Hmm... How can we test this? 

You can connect to windows installed in qemu with spice enabled and tablet mode disabled. Then try to use mouse without my patch and with. Without patch pointer is crazy because user can't position it(motion speed is inadequate). With this patch mouse pointer moved smoothly so you can position it. 

> as the browser will not let us grab and manipulate the pointer.

I also created some kind of synchronization areas near the edges of Display. This allow to synchronize mouse position with remote position. And solve this problem partially.

You can look deeper into my commit and saw my comments about this.

CendioOssman:
> You can connect to windows installed in qemu with spice tablet mode disabled and without spice/virtio drivers. Then try to use mouse without my patch and with. Without patch pointer is crazy because user can't position it(motion speed is inadequate). With this patch mouse pointer moved smoothly so you can position it.

I'm afraid I don't have easy access to a QEMU machine with Windows on it. Can this be tested with a free OS? I have a CentOS 7 machine here, but I'm not getting any sane behaviour.

> I also created some kind of synchronization areas near the edges of Display. This allow to synchronize mouse position with remote position. And solve this problem partially.
> 
> You can look deeper into my commit and saw my comments about it.

I did, and it looked very odd to me. That's why I wanted to test things. I fail to see how exaggerating the mouse movement can reliably work.

CendioOssman:
Hmm... It looks like there is actually some support for this scenario in the browsers:

https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API

Have you looked at that and see how it works here?

Ekstazi:
> I'm afraid I don't have easy access to a QEMU machine with Windows on it. Can this be tested with a free OS? I have a CentOS 7 machine here, but I'm not getting any sane behaviour.

As i know linux kernel include vmmouse driver which enable absolute pointing method. Your can test this patch only with windows or not *nix based systems. 

> Hmm... It looks like there is actually some support for this scenario in the browsers:
> https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API
> Have you looked at that and see how it works here?

Hm, i'll take look at it. Thx for the link.

Ekstazi:
> Hmm... It looks like there is actually some support for this scenario in the browsers:
> https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API
> Have you looked at that and see how it works here?

As i understand this is exactly that needed. I'll make some experiments at weekend. After that i'll share results.

CendioOssman:
> As i know linux kernel include vmmouse driver which enable absolute pointing method. Your can test this patch only with windows or not *nix based systems.

vmmouse is for VMware. If you remove the tablet device from QEMU, then you get an emulated PS/2 device handled by evdev on Linux. I.e. exactly the same as real hardware.

So there is something else missing in getting this working reliable. Or I simply haven't found a client that properly supports relative mouse movement. I thought that vinagre did...

Ekstazi:
> vmmouse is for VMware. If you remove the tablet device from QEMU, then you get an emulated PS/2 device handled by evdev on Linux. I.e. exactly the same as real hardware.

This driver also works for qemu. I tested it on windows. And it enables absolute positioning mode without tablet device.
https://github.com/qemu/qemu/blob/f2cfa1229e539ee1bb1822912075cf25538ad6b9/hw/i386/pc.c#L2036

As you can see qemu emulates vmware mouse. Probably you can blacklist vmmouse driver in linux guest and then relative pointer will be enabled.

Why you can't test it on windows vm ?

Ekstazi:
Also i think you can modify my code to see if absolute or relative pointer mode used.

Ekstazi:
If you can't find a guest with relative pointing mode I can temporary share windows guest with you. If you contact me by email.

samhed:
@Ekstazi I sent you an email

samhed:
We couldn't find a way to test this, and I wrote an email regarding sharing that windows guest of yours. If we are going to get this merged we need to be able to do some testing ourselves.

Ekstazi:
Hi, @samhed  I received your email. I'll provide test sandbox for you on this week.

CendioOssman:
And to reiterate, I suspect this won't be complete without the pointer lock stuff is in place.

Did you have a chance to test using that?

samhed:
No response.

Issue #358, opened 2014-04-16
unix auth support missing


jpecar:
In our environment we authenticate vnc users against ldap via pam. We use turbovnc set to pam-userpwd auth method. On protocol level this means tightvnc auth with capability set to 129 "TGHTULGNAUTH". I see it should be implemented around line 830 in rfb.js, together with "username" input field in UI that sends this value down to rfb.js.

I'll look into implementing it, but since I'm not familiar with this code, it will take me some time. Can anyone familiar with the code do it faster?

PR #1272, opened 2019-08-03
ADD: support Unix tight auth

koying:
Ref: #358

This is a somewhat crude (HTML/CSS/JS is not exactly my domain 😉) implementation of unix identification (TightVNC "ULGNAUTH")
Tested with TurboVNC

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -940,9 +940,9 @@ export default class RFB extends EventTargetMixin {
 
     // authentication
     _negotiate_xvp_auth() {
-        if (!this._rfb_credentials.username ||
-            !this._rfb_credentials.password ||
-            !this._rfb_credentials.target) {
+        if (this._rfb_credentials.username === undefined ||
+            this._rfb_credentials.password === undefined ||
+            this._rfb_credentials.target === undefined) {
             this.dispatchEvent(new CustomEvent(
                 "credentialsrequired",
                 { detail: { types: ["username", "password", "target"] } }));
@@ -961,7 +961,7 @@ export default class RFB extends EventTargetMixin {
     _negotiate_std_vnc_auth() {
         if (this._sock.rQwait("auth challenge", 16)) { return false; }
 
-        if (!this._rfb_credentials.password) {
+        if (this._rfb_credentials.password === undefined) {
             this.dispatchEvent(new CustomEvent(
                 "credentialsrequired",
                 { detail: { types: ["password"] } }));


app/ui.js
```diff
@@ -1148,25 +1148,45 @@ const UI = {
 
     credentials(e) {
         // FIXME: handle more types
+
+        document.getElementById("noVNC_username_block").classList.remove("noVNC_hidden");
+        document.getElementById("noVNC_password_block").classList.remove("noVNC_hidden");
+
+        let inpFocus = "none";
```
samhed: why is it called inpFocus?

app/ui.js
```diff
@@ -1148,25 +1148,45 @@ const UI = {
...
-    setPassword(e) {
+    setCredentials(e) {
         // Prevent actually submitting the form
         e.preventDefault();
 
-        const inputElem = document.getElementById('noVNC_password_input');
+        let inputElem = document.getElementById('noVNC_username_input');
+        const username = inputElem.value;
+
+        inputElem = document.getElementById('noVNC_password_input');
```
samhed: just use a separate variable for the password input element

app/ui.js
```diff
@@ -1148,25 +1148,45 @@ const UI = {
...
         const password = inputElem.value;
         // Clear the input after reading the password
         inputElem.value = "";
-        UI.rfb.sendCredentials({ password: password });
+
```
samhed: Remove one blank line pls

vnc.html
```diff
@@ -297,12 +297,16 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br>VNC</h1>
     <div class="noVNC_center noVNC_connect_layer">
     <div id="noVNC_password_dlg" class="noVNC_panel"><form>
```
samhed: Maybe the name of this container div should be changed to noVNC_credentials_dlg to be in line with your other name changes

app/ui.js
koying: "Input Focus" ;)

samhed:
Thanks for quickly fixing the things I pointed out. When I have the time I'll setup a server that supports this and do some testing.

koying:
BTW, I should have pointed out that my interest in the PR is to be able to do TOTP login (through PAM), not really to do plain Unix login. But yeah, it's transparent to both novnc and the vnc server...

app/styles/base.css
```diff
@@ -790,19 +790,23 @@ select:active {
...
+#noVNC_credentials_dlg.noVNC_open {
   transform: translateY(0);
 }
-#noVNC_password_dlg ul {
+#noVNC_credentials_dlg ul {
   list-style: none;
   margin: 0px;
   padding: 0px;
 }
+.noVNC_hidden {
```
CendioOssman: We already have `.noVNC_hidden` in some more specific places. Having it unqualified like this can get a bit confusing. Perhaps we should drop the others and have a generic one like this?

@samhed?

koying:
> I also noticed that this isn't documented over in rfbproto. @koying, feel like trying to provide a patch?

Honestly, I just "reverse-engineered" the TurboVNC code. Someone should test with TightVNC to confirm this is any way standard...

app/styles/base.css
koying: Yep, I didn't quite got the point of having specific ones all over the place, all doing exactly the same thing, tbh

samhed: > Perhaps we should drop the others and have a generic one like this?

sounds good to me

samhed:
> I'm wondering if we want to split it in two commits for clarity though. One adding the protocol support, and one improving the UI. @samhed, thoughts?

I'm fine either way. Having everything in one commit like this works too

koying:
BTW, I noticed https://github.com/novnc/noVNC/blob/master/core/rfb.js#L937 while coding.
Not sure if it's dead code or anything, but looks to me like xvp (whatever that is) won't work.

If it's meant to work, a third "target" field should be added to the credential dialog.

CendioOssman:
Right, you cannot currently specify target interactively with `vnc.html`. It can be given as an argument though, and other users of `RFB` might be implementing better support. So the code isn't entirely pointless. :)

Feel free to fix that up as well if you want.

koying:
I have no clue, really, just wanted to point it out in case it was not known.
Adding an additional field is trivial, but I wouldn't know how to test it even if I wanted to ;)

koying:
As I said, I have no clue re the "official" protocol, if any.

Older TightVNC code says "not used in the mainstream version" (https://github.com/TurboVNC/tightvnc/blob/master/vnc_unixsrc/include/rfbproto.h#L278) so my guess is that it was implemented for specific customer needs.

The way TurboVNC implemented it is exactly https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#plain-subtype

Maybe @dcommander, who seems the TurboVNC maintainer, might comment...

dcommander:
As far as I'm aware, the TurboVNC Server is the only VNC server that supports the Unix Login authentication sub-type.  That feature was added to the TurboVNC Server nearly 10 years ago by developers at UCAR.  Unix Login authentication was apparently implemented in the TightVNC Server at one time, but it was later removed (https://github.com/quentinsf/vnc2dl/commit/bfab378ffd202fcda764b992ad60045e6233e275) for unknown reasons.  I'm not sure which version of the TightVNC Server had that feature, but 1.3.10-- the version from which TurboVNC forked many years ago-- definitely doesn't.  

I see no issues with this implementation, except for the fact that noVNC seems to be ignoring the vendor and signature information sent by the server as part of the Tight security selector negotiation.  Refer to the Java TurboVNC Viewer code (https://github.com/TurboVNC/turbovnc/blob/master/java/com/turbovnc/rfb/CSecurityTight.java) for an example of how to correctly check that information.

Other notes:
* The Unix Login authentication sub-type has exactly the same capabilities as the VeNCrypt Plain security type.  The TurboVNC Server supports both.  The TigerVNC Server supports only VeNCrypt, so implementing VeNCrypt (even without TLS encryption-- just implementing the security selector) would be a more compatible solution.
* If VeNCrypt support is eventually added to noVNC, then developers should be mindful of the fact that the VeNCrypt security selector has to be implemented both as a top-level authentication type and as an authentication sub-type of the Tight security selector.  Refer again to the Java TurboVNC Viewer code for an example.  However, it would also be completely reasonable to drop support for the Tight security selector once VeNCrypt is implemented, since the only known server that supports the Tight security selector is TurboVNC, and it also supports VeNCrypt.

samhed:
Thanks for the details @dcommander 

So the best way forward would be to implement VeNCrypt handling:
 * Better server support
 * Already documented in rfbproto

VeNCrypt includes some handshaking and stuff before we can re-use this code for the VeNCrypt Plain subtype, but it doesn't look too complicated. Would you be up for adding this @koying ?

Another short-term option as I can see it is to simply rename "_negotiate_tight_unix_auth" to "_negotiate_VeNCrypt_plain_auth" and add a comment in "_negotiate_authentication" explaining that the two types are identical. Thoughts @CendioOssman ?

CendioOssman:
You misunderstood, the two types are not the same extension. They just happen to implement the same feature. So I don't think we should mix them up in the code.

So I think this PR is good to go. Even though I would have liked a patch for @rfbproto as well. :)

samhed:
Alright, yeah it's probably a bad idea to mix names. I'll make a PR with Tight Unix Auth for the rfbproto project then.

The specification for VeNCrypt plain doesn't have any information regarding the encoding of the username and password. @dcommander what string-encoding does Turbo use for VeNCrypt plain and for Tight Unix Auth?

dcommander:
Both types of authentication are back-ended by PAM, and the PAM functions only accept C `char` strings.

samhed:
A specification for Tight Unix Auth has now been added to the RFB protocol.

samhed:
While this works great with normal usernames, I tested this with a unicode username "userä" and it didn't work. TurboVNC server says:

28/11/2019 16:39:47 Adding user 'userä' to ACL with full control privileges
28/11/2019 16:43:38 User 'user�' is not in the ACL and has been denied access

koying:
@samhed Isn't the reason exactly what is described 2 comments above yours?
Anyway, that would a TurboVNC issue, not a noVNC one, as it seems the UTF8 username properly passed over.

samhed:
Connecting with the user "userä" works using TurboVNC's own viewer, so it is a noVNC issue.

samhed:
Whops, I accidentally pushed noVNC:master to your master, and now I can't undo it since the PR got closed :man_facepalming: sorry

Anyway, I committed your changes in these two commits: 5b453ed4a8593d68cc4e535ef6b5614c59ff4a4d and 94c89284fc1f4e4c41c0cf949fb6ef4c6e8eb332

Issue #1276, opened 2019-08-08
400 Client must support 'binary' or 'base64' protocol


BruceGujinhe:
**Describe the bug**
1. It works ok in local environment, such like: 
```
http://localhost:6080/vnc.html?host=cb0pek3c.xxx.com&port=10147&password=ODKY35Xuaf3IwpNNaSDZW&autoconnect=true
```
We got the web desk normally. and the websockify response is ok:
```
Request URL: wss://cb0pek3c.xxx.com:10147/websockify
Request Method: GET
Status Code: 101 Switching Protocols
```
2. But we got something unexpected in our server.
```
https://vnc.xxxx.vip/vnc.html?host=cb0pek3c.xxx.com&port=10147&password=WODKY35Xuaf3IwpNNaSDZW&autoconnect=true
```
We got the `websockify ` response like that:
```
Request URL: wss://cb0pek3c.xxx.com:10147/websockify
Request Method: GET
Status Code: 400 Client must support 'binary' or 'base64' protocol
```

**Client (please complete the following information):**
 - OS: centos 6.8
 - Browser: chrome
 - Browser version: the latest

**Server (please complete the following information):**
 - noVNC version:  latest
 - VNC server: none
 - WebSocket proxy: websockify

BruceGujinhe:
try lots of different methods, I found that the latest master code can't work correctly. 
the code I clone two weeks ago works fine at present.

tkedwards:
Same problem here, except I found that this issue doesn't happen when using the Opera browser, but does when using Chrome or Firefox (latest versions). I've tested on different computers - a Mac and a Chromebook and the issue happens in Chrome or Firefox. Changing the Websockify settings (e.g. replacing 'localhost' with 127.0.0.1) in the Novnc config on the browser does not help.

therevoman:
Ran into this error while trying to use it in a docker container today.

CendioOssman:
I'm afraid your websockify is too old and you need to upgrade it. It requires a non-standard behaviour that we recently removed from noVNC.

tkedwards:
I saw your latest (0.90) release of Websockify just before. I'm working on updating the Snap and should be done soon

tkedwards:
This is now 'fixed' in the Snap package version latest-20190813, in that the Snap package now uses the latest Websockify version

working-name:
sorry, it's not, getting the same error on 19.04

tkedwards:
Can you provide output of:

[...]

Issue #1271, opened 2019-08-01
Snap installation doesn't work


larryathey:
Tried "sudo snap install novnc" on both an Ubuntu 16.04 and 18.04 machines, package not found. I need to daemonize this and need your snap implementation to do so.

samhed:
We haven't gotten around to submitting the snap package to the store yet. We added the snap package to the master branch just a week ago.

tkedwards:
Sorry about the dealy with this. I created the Snap package (see snap/ directory in the source) and I'm trying to get it into Ubuntu's Snap store. For some reason the name 'novnc' is reserved, so I've submitted a request and am waiting to hear back from Canonical.

In the meantime you can build the Snap package locally with:
```
git clone https://github.com/novnc/noVNC.git
cd noVNC
snapcraft snap # Add --use-lxd if using a virtual machine which itself can't run other VMs (no VT-X extensions)
sudo snap install --dangerous novnc_*.snap
```

tkedwards:
So the Snap package is now in the store and released. Unfortunately it's affected by https://github.com/novnc/noVNC/issues/1276 so it will only currently work in Opera, but not in Firefox or Chrome. When that bug is fixed I'll probably need to update the Snap again

tkedwards:
Snap package (latest-20190813) is released on the store, it uses the latest Websockify so should no longer be affected by #1276. I think this issue can be closed

franciscoprin:
I am getting this error, any clue?

$ snapcraft snap
Issues while validating snapcraft.yaml: The 'version' property does not match the required schema: '@VERSION@' is not a valid snap version. Snap versions consist of upper- and lower-case alphanumeric characters, as well as periods, colons, plus signs, tildes, and hyphens. They cannot begin with a period, colon, plus sign, tilde, or hyphen. They cannot end with a period, colon, or hyphen

tkedwards:
Hi, what do the logs show:
`sudo snap logs novnc -n 200`

PR #1277, opened 2019-08-15
Latest updates for Snap package

tkedwards:
Changed Snap package to pull in Websockify directly from Github (https://github.com/novnc/websockify.git) instead of relying on the Ubuntu package which is now outdated. Also changed the version to make it clearer this is pulling directly from git

In my testing this fixes https://github.com/novnc/noVNC/issues/1276 and https://github.com/novnc/noVNC/issues/1271

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,6 @@
 name: novnc
 base: core18 # the base snap is the execution environment for this snap
-version: '1.1.0'
+version: 'latest-20190813'
 summary: Open Source VNC client using HTML5 (WebSockets, Canvas)
 description: |
   Open Source VNC client using HTML5 (WebSockets, Canvas).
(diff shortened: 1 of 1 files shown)


CendioOssman:
I think we should avoid bundling if we can. And I don't think we should keep making snapshot snap packages. Isn't it supposed to just be stable releases?

Whoever maintains the websockify package should make sure it is updated.

tkedwards:
Hi,

The whole point of snap packages is that they bundle what they need, this is by design. The original Snap package bundled websockify as installed from the websockify DEB package from Ubuntu 16.04. This no longer works because of the recent changes that caused https://github.com/novnc/noVNC/issues/1276
So instead I changed it to grab the latest websockify from git.

As for stable versions, once there is a next release of novnc (e.g. 1.2.0) I'll switch the Snap to package that. Currently 1.1.0 doesn't have the changes I made for the Snap packaging, so I had to take the latest from git on the date I built the package. Not ideal but it's only a temporary solution.

If you won't let me maintain the Snap package files in the repo I'll have to create a separate repo for it.

samhed:
> Currently 1.1.0 doesn't have the changes I made for the Snap packaging, so I had to take the latest from git on the date I built the package. Not ideal but it's only a temporary solution.

I know I'm late to the party here, but the latest release of websockify works well with the latest release of noVNC. I know that 1.1.0 of noVNC doesn't have snap. But why can't we keep 1.1.0 as noVNC version in our snap package and 0.9 of websockify?

What changes are you referring to?

tkedwards:
I was referring to some of the changes made originally when I first submitted the Snap package - I changed a bit of the logic for finding websockify in utils/launch.sh as well as adding  utils/svc_wrapper.sh. As these commits were done post-1.1.0, so the Snap package wouldn't work if it packaged just plain 1.1.0.

The Snap is currently pegged to those particular commits, so it's as close to 1.1.0 as I can get it while still including those commits. So the Snap package already sticks as close to a released version (1.1.0) as it can be.

Because both this and my last pull request have been closed I've created a separate git repo just for the Snap packaging, which I hope will just be a temporary solution. Maybe this will be simpler once 1.2.0 is out.

samhed:
Ok, I understand.

> Because both this and my last pull request have been closed I've created a separate git repo just for the Snap packaging, which I hope will just be a temporary solution. Maybe this will be simpler once 1.2.0 is out.

Yes, that sounds like a good plan.

PR #1278, opened 2019-08-15
Upgrade development tools

CendioOssman:
We have a lot of stuff version locked right now in `package.json`. That means we do not get new fixes or features in these tools, and we might have a lot of work piling up for the day we need to upgrade.

This PR removes the version lock and lets us use the latest and greatest. Some compatibility fixes were needed for this.

Note that we do have some minimum requirements that we might want to code in here, but I think we can handle them as they arise.

There is also the risk that upstream changes something that breaks our usage. But I think it is better if we notice that and fix our stuff in those cases.

Users should be largely unaffected as these dependencies are for development and not normal use.


samhed:
This commit message "Make sure translation tools are translated" should probably be "Make sure translation tools are scanned by lint", right?

CendioOssman:
Oops. Indeed. :)

Okay otherwise?

PR #1280, opened 2019-08-16
Avoid recursion in Alt check on Firefox

jalfd:
The Firefox workaround which checks for missing Alt key events may
synthesise new KeyboardEvents. On these events, checkAlt should not be
recursively triggered. Otherwise, we get "too much recursion" errors
whenever the Alt key is pressed.

### DIFF
--- a/core/input/keyboard.js
+++ b/core/input/keyboard.js
@@ -301,6 +301,9 @@ export default class Keyboard {
 
     // Firefox Alt workaround, see below
     _checkAlt(e) {
+        if (e.skipCheckAlt) {
+            return;
+        }
         if (e.altKey) {
             return;
         }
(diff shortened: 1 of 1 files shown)


CendioOssman:
Hmm... I wonder how this got overlooked. Thanks for having a look at it.

samhed:
The firefox workaround doesn't seem to work in the first place for me.. Monitoring keyboard events using 'xev' in the remote session, I'm only getting half of the events. The first time I press Alt I only get the KeyPress event, the second time I press Alt I only get the KeyRelease event.

What am I missing? And how do I test this fix?

jalfd:
I just pressed the left Alt key a couple of times, and noticed the recursion errors in Firefox' dev console. That was enough to reproduce the problem for me. Aside from that though, I didn't notice any missing key events, but I also didn't particularly look out for any. Other than this issue, keyboard input *seemed*  to generally work.

PR #1281, opened 2019-08-16
Optimize receive buffer

jalfd:
Fix two inefficiencies in how and when we compact the websocket receive buffer.
In most browsers, the impact of this is negligible, but on a high-res monitor in IE11, these issues could freeze the buffer for over half a second when a large FBU is received.

With these changes, compactions effectively disappear from IE11's profiler graph.

### DIFF
--- a/core/websock.js
+++ b/core/websock.js
@@ -247,12 +247,12 @@ export default class Websock {
         if (resizeNeeded) {
             const old_rQbuffer = this._rQ.buffer;
             this._rQ = new Uint8Array(this._rQbufferSize);
-            this._rQ.set(new Uint8Array(old_rQbuffer, this._rQi));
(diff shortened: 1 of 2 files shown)


jalfd:
Thanks for the kind words. :)

> And I can see a slight improvement, very nice.

FYI, I should mention that because  noVNC currently downscales the canvas to compensate for your DPI scaling, you won't see the full benefit unless you set the scaling factor in Windows to 100%. If you do that, you get a canvas matching the size of your actual desktop resolution, and then the difference in performance becomes *very* noticeable. That's how we became aware of the problem.

samhed:
Yeah, by default the scaling factor in Windows was 200%, before testing this I changed this to 100%. I guess the client machine I tested on was a bit too powerful to notice a big difference. I never did any profiling etc.

PR #1282, opened 2019-08-16
Don't process images until their 'load' event has been emitted

jalfd:
In IE11, the 'complete' property on an image is set before the image is
fully initialized (so it still has 0x0 dimensions). So rather than
relying on this property, always just wait for the 'load' event.

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -375,13 +375,16 @@ export default class Display {
 
     imageRect(x, y, mime, arr) {
         const img = new Image();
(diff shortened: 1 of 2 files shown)


CendioOssman:
I could reproduce it here. Thanks for finding this.

I'd like to reuse the existing infrastructure where possible though, so I've done a slightly different fix in c4eb4ddcfe284a84fe147544cddfce2962fd4a87. It should resolve the issue in the same way though.

PR #1285, opened 2019-08-25
Snap release latest 20190813

tkedwards:
I've modified the Snap to now explicitly pull in the 0.9.0 version of Websockify, instead of the latest from git. Note that this is how Snap packages work, they bundle their dependencies.

I've also fixed the noVNC 'part' to pull in the specific git commit that added Snap support. This is a temporary solution until the next release of noVNC (e.g. 1.2.0) is out

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,6 @@
 name: novnc
 base: core18 # the base snap is the execution environment for this snap
-version: '1.1.0'
+version: 'latest-20190813'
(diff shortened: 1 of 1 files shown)


tkedwards:
This is a follow on from https://github.com/novnc/noVNC/pull/1277, to try and fix the concerns raised there

samhed:
As discussed in #1277 we will not be committing changes to the snap package until noVNC 1.2.0 is out.

PR #1297, opened 2019-09-19
Encode & decode message in Base64 according to the websocket protocol

cbourgois:
Add support of `base64` encode/decode message when using websocket `base64` protocol.

### DIFF
--- a/core/websock.js
+++ b/core/websock.js
@@ -13,6 +13,7 @@
  */
 
 import * as Log from './util/logging.js';
+import Base64 from './base64.js';
 
 // this has performance issues in some versions Chromium, and
 // doesn't gain a tremendous amount of performance increase in Firefox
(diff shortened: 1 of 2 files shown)


core/websock.js
```diff
@@ -264,7 +268,13 @@ export default class Websock {
 
     _decode_message(data) {
         // push arraybuffer values onto the end
-        const u8 = new Uint8Array(data);
+        let u8;
+        if (this._websocket.protocol === 'base64') {
+            u8 = Base64.decode(data);
+        } else {
+            u8 = new Uint8Array(data);
+        }
```
marie-j: Could have been a ternary :)

CendioOssman:
The base64 stuff is a non-standard variant that was used early in the WebSocket prototypes. Since it isn't standardised it isn't really something we want in noVNC.

What environment do you have that needs this?

CendioOssman:
No response. Closing.

PR #1299, opened 2019-09-23
Add support for VMware cursor encoding

CendioNiko:
Supports both classic cursor type and alpha cursor type. In classic mode the server can send 'inverted' pixels for the cursor, our code does not support this but handles these pixels as opaque black.

Both classic and alpha cursors can be tested by using a Windows XP virtual machine in VMware. If cursor shadows are enabled in XP you will get VMware alpha encoded cursors and if shadows are disabled you will get classic ones.
You will probably also have to enable VNC connections for your VMware guest in the settings.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1239,6 +1239,7 @@ export default class RFB extends EventTargetMixin {
         encs.push(encodings.pseudoEncodingContinuousUpdates);
 
         if (this._fb_depth == 24) {
+            encs.push(encodings.pseudoEncodingVMwareCursor);
             encs.push(encodings.pseudoEncodingCursor);
         }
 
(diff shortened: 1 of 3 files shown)


core/rfb.js
```diff
@@ -1237,6 +1237,7 @@ export default class RFB extends EventTargetMixin {
         encs.push(encodings.pseudoEncodingXvp);
         encs.push(encodings.pseudoEncodingFence);
         encs.push(encodings.pseudoEncodingContinuousUpdates);
+        encs.push(encodings.pseudoEncodingVMwareCursor);
```
CendioOssman: I think we need to only advertise this encoding when `this._fb_depth` is `24`, just like for `pseudoEncodingCursor` as we assume a certain pixel format in the incoming data (for classic cursors at least).

samhed: Good point. Fixed now

PR #1301, opened 2019-09-25
Auto copy-paste with navigator.clipboard API

akamos:
Hi!
I added the automatical copy-paste function that is available through the `navigator.clipboard` API in some browsers (e.g. Google Chrome) as mentioned earlier in PR #1174.

Co-authored-by: bulldozier <kentdozier@gmail.com>

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -916,7 +916,7 @@ const UI = {
  * ------v------*/
 
     writeLocalClipboard(text) {
-        if (typeof navigator.clipboard !== "undefined" && typeof navigator.clipboard.readText !== "undefined") {
+        if (typeof navigator.clipboard !== "undefined" && typeof navigator.clipboard.writeText !== "undefined") {
             navigator.clipboard.writeText(text).then(() => {
                 let debugMessage = text.substr(0, 40) + "...";
                 Log.Debug('>> UI.setClipboardText: navigator.clipboard.writeText with ' + debugMessage);
--- a/app/ui.js
+++ b/app/ui.js
@@ -915,6 +915,32 @@ const UI = {
  *   CLIPBOARD
  * ------v------*/
 
+    writeLocalClipboard(text) {
+        if (typeof navigator.clipboard !== "undefined" && typeof navigator.clipboard.readText !== "undefined") {
+            navigator.clipboard.writeText(text).then(() => {
+                let debugMessage = text.substr(0, 40) + "...";
+                Log.Debug('>> UI.setClipboardText: navigator.clipboard.writeText with ' + debugMessage);
+            }).catch(() => {
+                Log.Error(">> UI.setClipboardText: Failed to write system clipboard (trying to copy from NoVNC clipboard)");
+            });
+        }
+    },
+
+    readLocalClipboard() {
+        // navigator.clipboard and navigator.clipbaord.readText is not available in all browsers
+        if (typeof navigator.clipboard !== "undefined" && typeof navigator.clipboard.readText !== "undefined") {
+            navigator.clipboard.readText()
+                .then((clipboardText) => {
+                    const text = document.getElementById('noVNC_clipboard_text').value;
+                    if (clipboardText !== text) {
+                        document.getElementById('noVNC_clipboard_text').value = clipboardText;
+                        UI.clipboardSend();
+                    }
+                })
+                .catch(err => Log.Warn("<< UI.readLocalClipboard: Failed to read system clipboard-: " + err));
+        }
+    },
+
     openClipboardPanel() {
         UI.closeAllPanels();
         UI.openControlbar();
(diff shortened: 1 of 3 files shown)


app/ui.js
```diff
@@ -915,6 +915,32 @@ const UI = {
  *   CLIPBOARD
  * ------v------*/
 
+    writeLocalClipboard(text) {
+        if (typeof navigator.clipboard !== "undefined" && typeof navigator.clipboard.readText !== "undefined") {
```
samhed: Shouldn't that check for navigator.clipboard.writeText?

app/ui.js
```diff
@@ -944,17 +970,20 @@ const UI = {
...
     clipboardClear() {
         document.getElementById('noVNC_clipboard_text').value = "";
+        UI.writeLocalClipboard("");
         UI.rfb.clipboardPasteFrom("");
     },
 
     clipboardSend() {
         const text = document.getElementById('noVNC_clipboard_text').value;
         Log.Debug(">> UI.clipboardSend: " + text.substr(0, 40) + "...");
+        UI.writeLocalClipboard(text);
```
samhed: Why writeLocalClipboard here?

app/ui.js
akamos: That's correct! It was a mistake of mine. I'll fix this.

app/ui.js
akamos: Thanks for pointing this out. I've recognized after that, that my code is still buggy in the case, if you change the clipboard in the "clipboard overlay textfield". I'll fix this in the next days.

app/ui.js
```diff
@@ -321,6 +321,8 @@ const UI = {
     addClipboardHandlers() {
         document.getElementById("noVNC_clipboard_button")
             .addEventListener('click', UI.toggleClipboardPanel);
+        document.getElementById("noVNC_clipboard_text")
+            .addEventListener('input', UI.syncClipboardPanelToLocalClipboard);
```
samhed: Does the 'input' event trigger when text is removed?

Shouldn't we use the same event as below, 'change'?

app/ui.js
samhed: What was the bug you saw?

Seeing the new comment you added to the handler for 'input', my question regarding why you placed writeLocalClipboard was answered.

What's the reason for separating the sync to the 'input' event instead of the 'change' one?

app/ui.js
akamos: The 'input' event even works when text is removed, e.g it fires on every text change immediately, instead of the 'change' event, which reacts only when you leave the text field. 

For seameless UX, my approach send the content from localClipboard via RFB when the vnc canvas event 'focus' is triggered. (e.g. it reads the content from localClipboard when clicking the browser canvas for example after copying new content). If somebody changes the content of clipboardPanel, the content additionally has to be written in localClipboard that on 'focus' the correct content is sent. When using only the 'change' event it would copy the content from the clipboardPanel to the localClipboard only after the 'focus' event was triggered (results to the wrong content is copied). That's why I use the 'input' event to synchronize the content from the clipboardPanel to the localClipboard first, before 'focus' is triggered.

samhed: I see, thank you for a good explanation.

samhed:
I took some time today for testing this. I have tested both Chrome 66 and Chrome 78 on Linux, and Chrome 77 on Windows 10. Things didn't work in any of the browsers, navigator.clipboard was undefined in all cases. Am I missing something?

akamos:
I don't know how you run the tests, but clipboard is for me defined in normal desktop environment of Chrome 78 on Windows 10 and Chrome 75 on Windows 7.

samhed:
> EDIT: It works only with HTTPS.

Ah, that's probably it. I'll retest with HTTPS when I have the time

ericfranz:
Both `navigator.clipboard` and `navigator.clipboard.writeText` are defined in Firefox. But Firefox does not support this, so every interaction in NoVNC that invokes `writeLocalClipboard` will result in the `Log.Error` being sent to the console. If you do something like below, you can avoid polluting the console when querying the `clipboard-write` permission would result in a type error:

```diff
     writeLocalClipboard(text) {
-        if (typeof navigator.clipboard !== "undefined" && typeof navigator.clipboard.writeText !== "undefined") {
-            navigator.clipboard.writeText(text).then(() => {
+        if (typeof navigator.clipboard !== "undefined" && typeof navigator.clipboard.writeText !== "undefined"
+            && typeof navigator.permissions !== "undefined" && typeof navigator.permissions.query !== "undefined"
+           ) {
+            navigator.permissions.query({name: 'clipboard-write'})
+            .then(() => { return navigator.clipboard.writeText(text) })
+            .then(() => {
                 let debugMessage = text.substr(0, 40) + "...";
                 Log.Debug('>> UI.setClipboardText: navigator.clipboard.writeText with ' + debugMessage);
-            }).catch(() => {
-                Log.Error(">> UI.setClipboardText: Failed to write system clipboard (trying to copy from NoVNC clipboard)");
+            }).catch((err) => {
+                if(err.name !== 'TypeError')
+                    Log.Error(">> UI.setClipboardText: Failed to write system clipboard (trying to copy from NoVNC clipboard)");
             });
         }
     },
```

ericfranz:
Another problem with this approach that needs to be addressed is that it makes a preexisting issue worse. Copy to the system clipboard >10232 bytes and paste into the clipboard input field and you will get a RangeError: Invalid typed array length, because that would exceed the queue buffer size for a message (the text + 8 chars for buffer): https://github.com/novnc/noVNC/blob/84a8c1b0cc18f5cea5e0cda45cf7795a48247394/core/websock.js#L35

In Chrome with this PR the problem is made worse due to the automation because if you just copy to the system clipboard >10232 bytes the NoVNC browser tab will immediately display the RangeError. Reload the page and you will get the RangeError again.

samhed:
Let's continue this on #1347

Neustradamus:
This PR has been replaced by @tobfah PR:
- https://github.com/novnc/noVNC/pull/1993

Linked to:
- https://github.com/juanjoDiaz/noVNC/pull/1
- https://github.com/novnc/noVNC/issues/1961
- https://github.com/novnc/noVNC/issues/1902
- https://github.com/novnc/noVNC/pull/1900
- https://github.com/novnc/noVNC/pull/1899
- https://github.com/novnc/noVNC/pull/1831
- https://github.com/novnc/noVNC/issues/1822
- https://github.com/novnc/noVNC/pull/1817
- https://github.com/novnc/noVNC/pull/1562
- https://github.com/novnc/noVNC/issues/1511
- https://github.com/novnc/noVNC/pull/1301
- https://github.com/novnc/noVNC/pull/1174

Issue #1302, opened 2019-09-30
Chrome 77 memory leak


gicuuu3:
**Describe the bug**

In Chrome 77(and 78), when running noVnc on a canvas object, if we switch to a different browser tab memory stops being garbage collected and fills up to the brim(graphics memory and then system ram).

**To Reproduce**

1. Access www.demodesk.com/manage
2. Create a free account
3. Start a demo(Via Meet Now button)
4. Share more => local desktop => share local desktop
5. Inspect memory consumption(disabling hardware acceleration helps visualize the problem since the graphics card memory will not be touched)

**Expected behavior**
The memory used should drop every couple of seconds(garbage collected)

**Client (please complete the following information):**
 - OS: Mint 18.2(Linux 4.8.0-53-generic x86_64)
 - Browser: Chrome
 - Browser version: 77 & 78

**Server (please complete the following information):**
 - noVNC version: commit id: 8acadd9e972b72eb5667e397c9dac07feca78949
 - VNC server: TigerVNC
 - WebSocket proxy: websockify

**Additional context**

I'm aware that this is exclusively a Chrome issues(since Chrome 76 worked perfectly fine), but I'm filing in case it helps narrow down the problem.

This seems to happen if the canvas is not being directly visualized(either running in a different tab or being set do `display: none`)

I've also tested on the latest release of noVNC, **1.1.0**, and the issue persisted.

Firefox does not seem to have this problem.

I've been able to also reproduce it on a Mac system.

-------------------
P.S. Thanks a lot for the noVNC project. It's awesome what you guys managed to do.

I'll appreciate any help or input into this issue. Thank you for your time!

samhed:
I've followed your instructions (not sure why they included "share local desktop" however since that has nothing to do with noVNC) but I can't reproduce the issue. 

I used chrome's built-in memory consumption monitor and ran with disabled hardware acceleration.

What kind of numbers are you seeing?

gicuuu3:
I mentioned _Local desktop sharing_ because it creates the perfect scenario to see the effects. But, I realized later that it's a privacy concern for people to share their screen. My bad.

There's another way to prove it:

- Open a youtube video in the internal browser
- Switch to a different tab in the external one

I've added a youtube video that shows the behavior recorded on my machine:
https://www.youtube.com/watch?v=93d7xJn5TD0

coinfastman:
Good afternoon. We also noticed problems with a memory leak after switching to version 77 of chrome.
[...]

PR #1303, opened 2019-10-02
flip display only if tab is visible (fixes issue #1302 memory leak on inactive tabs)

phcapde:


CendioOssman:
Nice find. :)

So it would seem there is some bug with Chrome's `CanvasRenderingContext2D.drawImage()`. Do you think you could report it to them? Perhaps even put together a small test case? Even if we make any changes they'll have to wait for a release until they are widely deployed. So Chrome is in a much better position to fix this.

As for this PR, I think it might be possible to merge something like this. I'm not a huge fan of adding workarounds for clear (and serious) bugs in other software, but in this case it would also be a slight efficiency improvement as there is no need to update a screen that isn't visible. Two notes though:

 * `document.hidden` is deprecated. `document.visibiltyState` is the current property.

 * We need to handle the `visibilitychange` event or we might fail to show the updates once the tab becomes visible again.

gicuuu3:
Hello! I've been looking into this issue for a while and I think that the chrome team is aware and that it should be fixed by this commit:

https://chromium-review.googlesource.com/c/chromium/src/+/1836016

Do you think that matches the current situation?

CendioOssman:
I don't think so, unless there is something subtle going on internally in Chrome. We are copying from one canvas to another. So there is no array object directly involved.

My guess would be that Chrome adds a queue of things to draw. But they've stopped processing that queue for hidden tabs, so it keeps growing.

gicuuu3:
I see. That commit was prompted as a bug fix to another memory leak in their `FileReader` API. I thought it might be connected somehow.

I wonder if they're slowly giving more attention to their `OffscreenCanvas` API and did not notice potential issues with the classical background canvas approach.

Regardless, thanks a lot for your input!

Philippe-Capdepuy:
Yes I guess it is an optimization anyway. For the visibilityState, I'm no expert but as long as the property is correctly updated by the browser I guess we don't really need to catch the event as the next drawing call (which seem to happen more or less constantly) should perform the flip.

I'm afraid I won't have time to report back in chrome, in just really needed it to work properly for my use case ;)

CendioOssman:
The redraws only happen if we're getting updates from the server, which we cannot rely on for all scenarios. So I'm afraid we'll need an event handler or we'll start getting bug reports about the screen not updating properly. :)

juanjoDiaz:
Instead of stop doing the flip conditionally, wouldn't it be possible to stop requesting updates from the server?

At glance, something like modifying the `RfB._onFlush` like:
```js
    _onFlush() {
        this._flushing = false;

         if (this.displayIsVisible) {
             // Resume processing
             if (this._sock.rQlen > 0) {
                 this._handle_message();
             }
          }
    }
```
and have an event listener like:
```js
document.addEventListener("visibilitychange", function() {
  if (document.visibilityState === 'visible') {
    this.displayIsVisible = true;
    this.onFlush(); // To resume processing
  } else {
    this.displayIsVisible = false;
  }
});
```

I'm not an expert in the RFB protocol so I'm not sure if what I'm saying is total nonsense.
What do you guys think?


I've been wondering if it would be possible to do something like this using the IntersectionObserver API so I can have a page with many canvases but only the visible ones are actually consuming resources.

CendioOssman:
For a modern server we don't actively request things, so we have to tell the server to pause. It's definitely possible, but it will require a bit of work.

But that's probably step two. We can start with finishing up this part.

PR #1311, opened 2019-10-25
Dynamically detect scrollbar size

samhed:
Instead of always disabling scrollbars on iOS and Android, this PR adds dynamic detection for this.

The reason scrollbars were disabled on iOS and Android was that on these platforms the scrollbars were very thin and gets hidden when you don't use them. We can detect this by looking at the size of the scrollbars.

Verified on:

**iOS 13**
- [x] Safari - hidden
- [x] Firefox - hidden
- [x] Chrome - hidden

**Android 9**
- [x] Firefox 69 - hidden
- [x] Chrome 77 - hidden

**Chromebook**
- [x] Firefox - hidden
- [x] Firefox - hidden

**Fedora 31**
- [x] Firefox 69 - shows
- [x] Chrome 77 - shows

**macOS 10.15**
- [x] Safari - hidden
- [x] Firefox - hidden
- [x] Chrome - hidden

**Windows 10**
- [x] Firefox - shows
- [x] Chrome - shows
- [x] IE - shows
- [x] Edge - shows

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -8,7 +8,7 @@
 
 import * as Log from '../core/util/logging.js';
 import _, { l10n } from './localization.js';
-import { isTouchDevice, isSafari, isIOS, isAndroid, dragThreshold }
+import { isTouchDevice, isSafari, hasScrollbarGutter, dragThreshold }
     from '../core/util/browser.js';
 import { setCapture, getPointerEvent } from '../core/util/events.js';
 import KeyTable from "../core/input/keysym.js";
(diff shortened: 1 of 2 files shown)


samhed:
This has been fully tested now - fixes issue #1341

samhed:
Might be worth mentioning here about potential future changes in this area:

https://css-tricks.com/almanac/properties/s/scrollbar-gutter/

samhed:
Updated according to your comments :+1:

PR #1312, opened 2019-10-25
Fix so that you can use the keyboard after using the extra keys

samhed:
We got a regression from 2afda54 and other commits that were made at the same time. The regression was that we could no longer use the keyboard directly after using the extra keys, you would have to click the canvas first. This PR fixes that and also adds a visual queue to the fact that the canvas got the focus by fading the controlbar after a click on any of the extra keys.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1503,20 +1503,20 @@ const UI = {
     },
 
     sendEsc() {
-        UI.rfb.sendKey(KeyTable.XK_Escape, "Escape");
+        UI.sendKey(KeyTable.XK_Escape, "Escape");
     },
 
     sendTab() {
-        UI.rfb.sendKey(KeyTable.XK_Tab, "Tab");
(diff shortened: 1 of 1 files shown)


samhed:
One potentially controversial change that's part of this is that moving the mouse within the control bar will not activate it any longer. With this change you have to either click in it or move out and in again.

This change was added so that the visual queue for the focus change would work well.

An option would be to keep the mousemove listener but temporarily disable it when any of the extra keys are used. This would add complexity however.

CendioOssman:
> One potentially controversial change that's part of this is that moving the mouse within the control bar will not activate it any longer. With this change you have to either click in it or move out and in again.

:-1: We are trading one issue for another, and I'm not convinced that it's the best trade-off. I think we should keep the existing behaviour for now.

samhed:
Fair enough, I removed that commit. I'll keep the fadeout though as it still has a bit value on touch devices.

Issue #1317, opened 2019-11-07
Empty Password not allowed to connect


chrisjdev:
**Describe the bug**
I have a VNC where you can configure an empty password. I am then able to connect to it with RealVNC, but unable to connect to it with noVNC. noVNC says credentialsrequired.

**To Reproduce**
1. Configure a VNC server with no username and an empty password
2. Attempt to connect with noVNC
3. See error

**Expected behavior**
Connects to the VNC server with an empty-string password without issue.

**Client (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Browser version: 78

**Server (please complete the following information):**
 - noVNC version: 1.1
 - VNC server: x11vncserver

**Additional context**
I think it comes down to this check in rfb.js, "if (!this._rfb_credentials.password)", since an empty password is falsy, it doesn't accept it as potentially valid. I found by changing it to "if (this._rfb_credentials.password === undefined)", it then worked as I expected.

samhed:
Could you share the command used to start x11vncserver?

chrisjdev:
Looks like it's `/usr/bin/x11vnc -quiet -passwdfile /tmp/.vncpass -forever -shared -rotate +90 ~` and the contents of /tmp/.vncpass is `__EMPTY__`

samhed:
Thanks, I can reproduce the issue.

samhed:
Fixed in #1318

PR #1318, opened 2019-11-07
Fix for issue #1317

chrisjdev:
Checking password for undefined instead of falsy. That way an empty password string is allowed.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -940,9 +940,9 @@ export default class RFB extends EventTargetMixin {
 
     // authentication
     _negotiate_xvp_auth() {
-        if (!this._rfb_credentials.username ||
-            !this._rfb_credentials.password ||
-            !this._rfb_credentials.target) {
+        if (this._rfb_credentials.username === undefined ||
+            this._rfb_credentials.password === undefined ||
+            this._rfb_credentials.target === undefined) {
             this.dispatchEvent(new CustomEvent(
                 "credentialsrequired",
                 { detail: { types: ["username", "password", "target"] } }));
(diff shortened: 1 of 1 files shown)


CendioOssman:
Nice find. Thanks.

The other methods also seem to have the same issue. Could you fix those as well?

CendioOssman:
And please change the commit message heading to something that is independent of GitHub. We want to make sure `git shortlog` works well.

samhed:
I can verify that your fix works well.

> The other methods also seem to have the same issue. 

@CendioOssman did you see any functions other than _negotiate_xvp_auth() that need to be fixed?

A good commit message heading would be like "Allow connecting with empty passwords"

CendioOssman:
Didn't check more than that one. :)

core/rfb.js
```diff
@@ -961,7 +961,7 @@ export default class RFB extends EventTargetMixin {
     _negotiate_std_vnc_auth() {
         if (this._sock.rQwait("auth challenge", 16)) { return false; }
 
-        if (!this._rfb_credentials.password) {
+        if (this._rfb_credentials.password === undefined) {
```
samhed: Could you add the same check for `undefined` for username, password, and target in the function `_negotiate_xvp_auth()` above?

chrisjdev: Sure

chrisjdev:
Perhaps if this gets merged, the commits can be squashed with the message, "Allow connecting with empty string credentials"?

CendioOssman:
Indeed. Should be fixed in 06a8f7d91a1555f9309791b6cda662cbb4539aa6. Thanks for reporting.

Issue #1326, opened 2019-11-16
decodeUTF8 throws an exception when window name contains non utf8 characters


vanym:
**Describe the bug**
If window name contains non utf8 characters decodeURIComponent inside decodeUTF8 throws
```
URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at decodeUTF8 (http://localhost:6080/core/util/strings.js:13:12)
    at RFB._negotiate_server_init (http://localhost:6080/core/rfb.js:1158:20)
    at RFB._init_msg (http://localhost:6080/core/rfb.js:1285:29)
    at RFB._handle_message (http://localhost:6080/core/rfb.js:725:22)
    at Object.message (http://localhost:6080/core/rfb.js:185:18)
    at Websock._recv_message (http://localhost:6080/core/websock.js:283:33)
```
**To Reproduce**
1. Start `xdemineur` (it contains non utf8 char in window name)
2. Get it window id using `xwininfo`
3. Start `x11vnc -id` with xdemineur window id
4. Connect with noVNC
5. See `noVNC encountered an error` dialog

**Expected behavior**
Work around strings with non utf8 chars and do not crash

**Client:**
 - OS: GNU/Linux
 - Browser: chromium
 - Browser version: 77

**Server:**
 - noVNC version: commit id: 686c8d2
 - VNC server: x11vnc (0.9.16)
 - WebSocket proxy: websockify (4ecff11)

**Additional context**
Window name: `xdémineur`
Window name in hex: `78 64 e9 6d 69 6e 65 75 72`
Window name after js escape: `xd%E9mineur`
Looks like issue #688 describes similar problem

samhed:
Fixed in #1327

PR #1327, opened 2019-11-16
Solution option to solve #1326 problem

vanym:
Add try catch to `decodeUTF8` function to solve #1326


CendioOssman:
Hmm... This change will assume it is ISO 8859-1 if the decoding fails. This is correct in some cases, but not all. So I'm not sure this is the correct place to put that code.

samhed:
Thanks!

Issue #1267, opened 2019-07-24
Touch device - can't send mouse scrolls to remote


juanjoDiaz:
**Describe the bug**
Mobile scrolling doesn't seem to work (it might just not be supported).
Normal mobile scrolling (pressing and dragging) just selects text.
Choosing the wheel button from the sidebar doesn't make any difference. It stops selecting text but still it doesn't scroll.

**To Reproduce**
1. Open a noVNC connection to any VNC-enabled server
2. Open a browser and navigate to your favorite news site
3. Try to scroll down

**Expected behavior**
Ideally, I'd expect scroll to work as it would in a locally opened application.
As a minimum, choosing the scroll button of the mouse from the side menu should allow scrolling.

**Client (please complete the following information):**
 - OS: iOS
 - Browser: safari
 - Browser version: latest

**Server (please complete the following information):**
 - noVNC version: 1.1.0
 - VNC server: TigerVNC
 - WebSocket proxy: websockify

samhed:
Yeah. noVNC doesn't have any touch gesture support at all yet.

So you're expecting that pressing and dragging one finger across the screen would send a scroll to the application? In that case, we would need that to be behind a toggle to allow for example moving a window on the remote desktop when you're expected to click and drag.

The mouse button in the sidebar toggles which mouse button is sent. When the 'wheel' button is toggled you will send a middle-mouse click - i.e. the same as when pressing (not scrolling) the scroll wheel on a physical mouse. So we would need a different toggle for these things.

Taking a step back to reflect on gestures as a whole, I think the most important actions that would feel natural to accomplish with gestures would be:

* _Send mouse scrolls_
* _Send right click_
* _Send Ctrl + Scroll_ (many applications handle this as zoom)
* _Zoom the viewport_
* _Pan the viewport_

Then comes the question what type of gesture should trigger these actions? I can see two paths here, either you try to find gestures that doesn't conflict with most common mouse actions, for example two-finger gestures:

* **Drag two fingers** to _send mouse scrolls_
* **Tap with two fingers** to _send right clicks_
* **Pinch with two fingers** to _send Ctrl + scroll_

This way you could still work as normal with one-finger press and drag allowing you to drag and drop in the remote session as normal for example.

The second path would be more in line to what you were expecting, to have gestures mimicking local applications. But in that case I see more a focus on controlling the noVNC viewport rather than sending mouse events to the remote:

* **One-finger press and drag** to _pan the viewport_
[...]

PR #1346, opened 2019-12-17
Separate mouse and touch events

juanjoDiaz:
This PR is an initial step towards #1267 
It separates the mouse event handling from the touch event handling.

A separate step could be to implement multiple Touch modules for different interaction styles a.k.a convert touch events to mouse events in different ways.
For example, one that would work for me would be something like:

| Touch action | Click event |
|------|------|
| Single Tap| Click|
| Long press | Right-click (contextmenu or click event)|
| Press and then drag | Click and drag (mousemove event) |
| Two-finger press and drag | Pan (Wheel event)|
| Two-finger pinch | Zoom (Not sure of the events here) |

A library like hammer.js would make such implementations very easy.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -10,13 +10,14 @@
 import { toUnsigned32bit, toSigned32bit } from './util/int.js';
 import * as Log from './util/logging.js';
 import { encodeUTF8, decodeUTF8 } from './util/strings.js';
-import { dragThreshold } from './util/browser.js';
+import { dragThreshold, isTouchDevice } from './util/browser.js';
 import EventTargetMixin from './util/eventtarget.js';
 import Display from "./display.js";
 import Inflator from "./inflator.js";
 import Deflator from "./deflator.js";
 import Keyboard from "./input/keyboard.js";
 import Mouse from "./input/mouse.js";
+import Touch from "./input/touch.js";
 import Cursor from "./util/cursor.js";
 import Websock from "./websock.js";
 import DES from "./des.js";
@@ -115,6 +116,7 @@ export default class RFB extends EventTargetMixin {
         this._flushing = false;         // Display flushing state
         this._keyboard = null;          // Keyboard input handler object
         this._mouse = null;             // Mouse input handler object
+        this._touch = null;             // Touch input handler object
 
         // Timers
         this._disconnTimer = null;      // disconnection timer
@@ -202,8 +204,14 @@ export default class RFB extends EventTargetMixin {
         this._keyboard.onkeyevent = this._handleKeyEvent.bind(this);
 
         this._mouse = new Mouse(this._canvas);
-        this._mouse.onmousebutton = this._handleMouseButton.bind(this);
-        this._mouse.onmousemove = this._handleMouseMove.bind(this);
+        this._mouse.onmousebutton = this._handlePointerPress.bind(this);
+        this._mouse.onmousemove = this._handlePointerMove.bind(this);
+
+        if (isTouchDevice) {
+            this._touch = new Touch(this._canvas);
+            this._touch.ontouch = this._handlePointerPress.bind(this);
+            this._touch.ontouchmove = this._handlePointerMove.bind(this);
+        }
 
         this._sock = new Websock();
         this._sock.on('message', () => {
@@ -292,17 +300,19 @@ export default class RFB extends EventTargetMixin {
             if (viewOnly) {
                 this._keyboard.ungrab();
                 this._mouse.ungrab();
+                if (this._touch) { this._touch.ungrab(); }
             } else {
                 this._keyboard.grab();
                 this._mouse.grab();
+                if (this._touch) { this._touch.grab(); }
             }
         }
     }
 
     get capabilities() { return this._capabilities; }
 
-    get touchButton() { return this._mouse.touchButton; }
-    set touchButton(button) { this._mouse.touchButton = button; }
+    get touchButton() { return this._touch.touchButton; }
+    set touchButton(button) { this._touch.touchButton = button; }
 
     get clipViewport() { return this._clipViewport; }
     set clipViewport(viewport) {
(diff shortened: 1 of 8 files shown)


core/rfb.js
```diff
@@ -177,8 +179,14 @@ export default class RFB extends EventTargetMixin {
         this._keyboard.onkeyevent = this._handleKeyEvent.bind(this);
 
         this._mouse = new Mouse(this._canvas);
-        this._mouse.onmousebutton = this._handleMouseButton.bind(this);
-        this._mouse.onmousemove = this._handleMouseMove.bind(this);
+        this._mouse.onmousebutton = this._handlePointerPress.bind(this);
+        this._mouse.onmousemove = this._handlePointerMove.bind(this);
+
+        if (isTouchDevice) {
+            this._touch = new Touch(this._canvas, this._mouse);
```
samhed: The second argument to the constructor looks like it should be removed.

core/rfb.js
```diff
@@ -177,8 +179,14 @@ export default class RFB extends EventTargetMixin {
...
+            this._mouse.ontouch = this._handlePointerPress.bind(this);
+            this._mouse.ontouchmove = this._handlePointerMove.bind(this);
```
samhed: I believe these two should be `this._touch.ontouch` and `this._touch.ontouchmove`

docs/API-internal.md
```diff
@@ -58,29 +58,52 @@ callback event name, and the callback function.
...
+| name   | parameters | description
+| ------ | ---------- | ------------
+| grab   | ()         | Begin capturing touch events
+| ungrab | ()         | Stop capturing touch events
+
+### 2.2.2 Callbacks
+
+| name          | parameters          | description
+| ------------- | ------------------- | ------------
+| ontouch       | (x, y, down, bmask) | Handler for touch even (as button click/release)
```
samhed: "event"

core/rfb.js
```diff
@@ -264,17 +272,20 @@ export default class RFB extends EventTargetMixin {
             if (viewOnly) {
                 this._keyboard.ungrab();
                 this._mouse.ungrab();
+                if (this._touch) { this._touch.ungrab(); }
             } else {
                 this._keyboard.grab();
                 this._mouse.grab();
+                this._mouse.grab();
```
samhed: Looks like one `_mouse.grab()` too many

samhed:
And btw regarding hammerjs it would have been nice, but that project seems dead seeing as the latest release was more than 3 years ago and the maintainers don't seem to be handling issues or pull requests anymore.

If anyone knows of a good alternative library please tell us.

juanjoDiaz:
Fixed your comments.
I wanted to check if this is wanted before going further since I've had plenty of PRs rejected in this project 🙂
Some testing is still needed for sure. I'd be testing more on Monday.

Regarding Hammer.js, I know that it's not the best-maintained project but it is still the de-facto library for touch support on websites and it has over 1 million weekly downloads in npm. So not sure if there is a better alternative but I couldn't find any and it seems unlikely that there is one.

CendioOssman:
:+1: to the principle of splitting things. Haven't checked the code yet.

FYI, we (Cendio) are planning on getting some gestures up and running in the coming months. This would probably be a good base for that.

juanjoDiaz:
Is there anything left for me to do in this PR?

samhed:
Looks like all my code-comments are addressed. How well tested is this?

juanjoDiaz:
I've been using noVNC.html on a tablet with no problem.

Can you think of any particular thing that I should test?

CendioOssman:
I did a quick read of the code and it looks good AFAICT.

But I realized a big issue; the browser are moving away from the mouse and touch events in favour of common pointer events. Is this split here then perhaps counter productive as we have to have common code once we handle pointer events?

juanjoDiaz:
I don't think that it's counter-productive.
We have a handler for a mouse event, a handler for touch events, and we could add a handler for Pointer events.

Then the logic would be:
```js
if (arePointerEventsSupported) {
  // use pointer events handler
} else {
  // use mouse events handler

  if (isTouchDevice) {
    // use touch events handler
  }
}
```

CendioOssman:
I guess. No risk of a need for sharing code in the pointer handling and the other two though?

We will probably be looking at the pointer events soon-ish. Is it okay if we hold off on this PR a few weeks until we can get a better idea of how those work? I'd just like to confirm that we're not shuffling code back and forth needlessly and making git blame a pain.

juanjoDiaz:
I don't think that there is any shared code between those 3 handlers since the shared parts are the actual event handlers which are set in the `rfb.js` file.
Also, if anything else should be shared it can be placed in a helper class that's composed into the handlers. That would provide a nice and clean separation of concerns imho.

I'm OK waiting a few weeks for this. Just let me know if there is anything that I can do to help.

When looking at pointer-events are you just looking at using them instead of mouse or touch events? Or are you planning to add better gesture support as you also commented above?

CendioOssman:
> When looking at pointer-events are you just looking at using them instead of mouse or touch events? Or are you planning to add better gesture support as you also commented above?

The primary motivator is gesture support, which was originally intended to be built upon touch events. But as we're seeing issues with them, we will probably look at pointer events primarily now and keep mouse/touch support around only as much as is needed for legacy. I don't know what this means specifically until we've looked at pointer events more closely and evaluated which browsers support them.

CendioNiko:
We have been in contact with upstream about pointer events, and we reached a decision not to migrate fully to pointer events as of now.  This is because they are ending support for 'ontouch' and createEvent('TouchEvent') etc. but they are seemingly keeping addEventListener(). So this PR is of great interest!

However when testing this code I encountered the following problem:
 * 'Touch and drag' events does not seem to get captured by our session anymore. I encountered this error on most platforms with touch. Tested on Chrome/Firefox on android, Chrome in Linux and Chrome/Edge on Windows 10.

Could you have a look at this?
If you don't have any touch device to try this out you can use touch emulation on Chrome.

juanjoDiaz:
I'll try to take a look at this during this week.

What's not working is, for example, dragging icons on a desktop, right?

CendioNiko:
Thank you!
Yeah, you should be able to see the problem by trying to drag an icon.

samhed:
Replaced by changes made in #1414

PR #1347, opened 2019-12-18
Add automatic clipboard support

juanjoDiaz:
This PR adds support for automatic copy-pasting.
So when you are focused on the canvas and paste text it's pasted in the server and when you copy something in the server it's automatically copied to your local keyboard.

I've seen #1301 and #1174. But this takes a very different approach since it adds support for copy-pasting to noVNC's core. And I think that it's quite cleaner.

I've tested it in Safari, Firefox, and Chrome.
Unfortunately, the Clipboard API is only supported by Chrome and the `paste` event is broken (see https://bugs.chromium.org/p/chromium/issues/detail?id=634426).
So it just does nothing in Safari and Firefox, while in Chrome the copying of data works but not the pasting.

So, for now, this is not super useful but it should become better as browser support improves.

WDYT?

### DIFF
--- a/core/clipboard.js
+++ b/core/clipboard.js
@@ -1,0 +1,46 @@
+export default class Clipboard {
+    constructor(target) {
+        this._target = target;
+
+        this._eventHandlers = {
+            'copy': this._handleCopy.bind(this),
+            'paste': this._handlePaste.bind(this)
+        };
+
+        // ===== EVENT HANDLERS =====
+
+        this.onpaste = () => {};
+    }
+
+    // ===== PRIVATE METHODS =====
+
+    _handleCopy(e) {
+        if (navigator.clipboard.writeText) {
+            navigator.clipboard.writeText(e.clipboardData.getData('text/plain')).catch(() => {/* Do nothing */});
+        }
+    }
+
+    _handlePaste(e) {
+        if (navigator.clipboard.readText) {
+            navigator.clipboard.readText().then(this.onpaste).catch(() => {/* Do nothing */});
+        } else if (e.clipboardData) {
+            this.onpaste(e.clipboardData.getData('text/plain'));
+        }
+    }
+
+    // ===== PUBLIC METHODS =====
+
+    grab() {
+        if (!Clipboard.isSupported) return;
+        this._target.addEventListener('copy', this._eventHandlers.copy);
+        this._target.addEventListener('paste', this._eventHandlers.paste);
+    }
+
+    ungrab() {
+        if (!Clipboard.isSupported) return;
+        this._target.removeEventListener('copy', this._eventHandlers.copy);
+        this._target.removeEventListener('paste', this._eventHandlers.paste);
+    }
+}
+
+Clipboard.isSupported = (navigator && navigator.clipboard) ? true : false;
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -13,6 +13,7 @@ import { encodeUTF8, decodeUTF8 } from './util/strings.js';
 import { dragThreshold } from './util/browser.js';
 import EventTargetMixin from './util/eventtarget.js';
 import Display from "./display.js";
+import Clipboard from "./clipboard.js";
 import Inflator from "./inflator.js";
 import Deflator from "./deflator.js";
 import Keyboard from "./input/keyboard.js";
@@ -113,6 +114,7 @@ export default class RFB extends EventTargetMixin {
         this._sock = null;              // Websock object
         this._display = null;           // Display object
         this._flushing = false;         // Display flushing state
+        this._clipboard = null;         // Clipboard object
         this._keyboard = null;          // Keyboard input handler object
         this._mouse = null;             // Mouse input handler object
 
@@ -197,6 +199,9 @@ export default class RFB extends EventTargetMixin {
             throw exc;
         }
         this._display.onflush = this._onFlush.bind(this);
+
+        this._clipboard = new Clipboard(this._canvas);
+        this._clipboard.onpaste = this.clipboardPasteFrom.bind(this);
 
         this._keyboard = new Keyboard(this._canvas);
         this._keyboard.onkeyevent = this._handleKeyEvent.bind(this);
@@ -292,9 +297,11 @@ export default class RFB extends EventTargetMixin {
             if (viewOnly) {
                 this._keyboard.ungrab();
                 this._mouse.ungrab();
+                this._clipboard.ungrab();
             } else {
                 this._keyboard.grab();
                 this._mouse.grab();
+                this._clipboard.grab();
             }
         }
     }
@@ -1398,8 +1405,11 @@ export default class RFB extends EventTargetMixin {
         this._setDesktopName(name);
         this._resize(width, height);
 
-        if (!this._viewOnly) { this._keyboard.grab(); }
-        if (!this._viewOnly) { this._mouse.grab(); }
+        if (!this._viewOnly) {
+            this._keyboard.grab();
+            this._mouse.grab();
+            this._clipboard.grab();
+        }
 
         this._fb_depth = 24;
 
@@ -1516,9 +1526,13 @@ export default class RFB extends EventTargetMixin {
                 return true;
             }
 
-            this.dispatchEvent(new CustomEvent(
-                "clipboard",
-                { detail: { text: text } }));
+            this.dispatchEvent(new CustomEvent("clipboard", { detail: { text: text } }));
+
+            if (Clipboard.isSupported) {
+                const clipboardData = new DataTransfer();
+                clipboardData.setData("text/plain", text);
+                this._canvas.dispatchEvent(new ClipboardEvent('copy', { clipboardData }));
+            }
 
         } else {
             //Extended msg.
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1531,7 +1531,12 @@ export default class RFB extends EventTargetMixin {
             if (Clipboard.isSupported) {
                 const clipboardData = new DataTransfer();
                 clipboardData.setData("text/plain", text);
-                this._canvas.dispatchEvent(new ClipboardEvent('copy', { clipboardData }));
+                const clipboardEvent = new ClipboardEvent('paste', { clipboardData });
+                // Force initialization since the constructor is broken in Firefox
+                if (!clipboardEvent.clipboardData.items.length) {
+                    clipboardEvent.clipboardData.items.add(text, "text/plain");
+                }
+                this._canvas.dispatchEvent(clipboardEvent);
             }
 
         } else {
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1531,7 +1531,7 @@ export default class RFB extends EventTargetMixin {
             if (Clipboard.isSupported) {
                 const clipboardData = new DataTransfer();
                 clipboardData.setData("text/plain", text);
-                const clipboardEvent = new ClipboardEvent('paste', { clipboardData });
+                const clipboardEvent = new ClipboardEvent('copy', { clipboardData });
                 // Force initialization since the constructor is broken in Firefox
                 if (!clipboardEvent.clipboardData.items.length) {
                     clipboardEvent.clipboardData.items.add(text, "text/plain");
--- a/docs/API-internal.md
+++ b/docs/API-internal.md
@@ -21,6 +21,8 @@ keysym values.
 * __Display__ (core/display.js): Efficient 2D rendering abstraction
 layered on the HTML5 canvas element.
 
+* __Clipboard__ (core/clipboard.js): Clipboard event handler.
+
 * __Websock__ (core/websock.js): Websock client from websockify
 with transparent binary data support.
 [Websock API](https://github.com/novnc/websockify-js/wiki/websock.js) wiki page.
@@ -28,7 +30,7 @@ with transparent binary data support.
 
 ## 1.2 Callbacks
 
-For the Mouse, Keyboard and Display objects the callback functions are
+For the Mouse, Keyboard, Display and Clipboard objects the callback functions are
 assigned to configuration attributes, just as for the RFB object. The
 WebSock module has a method named 'on' that takes two parameters: the
 callback event name, and the callback function.
@@ -118,3 +120,23 @@ None
 | name    | parameters | description
 | ------- | ---------- | ------------
 | onflush | ()         | A display flush has been requested and we are now ready to resume FBU processing
+
+
+## 2.4 Clipboard Module
+
+### 2.4.1 Configuration Attributes
+
+None
+
+### 2.4.2 Methods
+
+| name               | parameters        | description
+| ------------------ | ----------------- | ------------
+| grab               | ()                | Begin capturing clipboard events
+| ungrab             | ()                | Stop capturing clipboard events
+
+### 2.3.3 Callbacks
+
+| name    | parameters | description
+| ------- | ---------- | ------------
+| onpaste | (text)     | Called with the text content of the clipboard when the user paste something
--- a/tests/test.clipboard.js
+++ b/tests/test.clipboard.js
@@ -1,0 +1,54 @@
+const expect = chai.expect;
+
+import Clipboard from '../core/clipboard.js';
+
+describe('Automatic Clipboard Sync', function () {
+    "use strict";
+
+    if (Clipboard.isSupported) {
+        beforeEach(function () {
+            if (navigator.clipboard.writeText) {
+                sinon.spy(navigator.clipboard, 'writeText');
+            }
+            if (navigator.clipboard.readText) {
+                sinon.spy(navigator.clipboard, 'readText');
+            }
+        });
+
+        afterEach(function () {
+            if (navigator.clipboard.writeText) {
+                navigator.clipboard.writeText.restore();
+            }
+            if (navigator.clipboard.readText) {
+                navigator.clipboard.readText.restore();
+            }
+        });
+    }
+
+    it('incoming clipboard data from the server is copied to the local clipboard', function () {
+        const text = 'Random string for testing';
+        const clipboard = new Clipboard();
+        if (Clipboard.isSupported) {
+            const clipboardData = new DataTransfer();
+            clipboardData.setData("text/plain", text);
+            clipboard._handleCopy(new ClipboardEvent('paste', { clipboardData }));
+            if (navigator.clipboard.writeText) {
+                expect(navigator.clipboard.writeText).to.have.been.calledWith(text);
+            }
+        }
+    });
+
+    it('should copy local pasted data to the server clipboard', function () {
+        const text = 'Another random string for testing';
+        const clipboard = new Clipboard();
+        clipboard.onpaste = pasterText => expect(pasterText).to.equal(text);
+        if (Clipboard.isSupported) {
+            const clipboardData = new DataTransfer();
+            clipboardData.setData("text/plain", text);
+            clipboard._handlePaste(new ClipboardEvent('paste', { clipboardData }));
+            if (navigator.clipboard.readText) {
+                expect(navigator.clipboard.readText).to.have.been.called;
+            }
+        }
+    });
+});
--- a/tests/test.clipboard.js
+++ b/tests/test.clipboard.js
@@ -31,7 +31,12 @@ describe('Automatic Clipboard Sync', function () {
         if (Clipboard.isSupported) {
             const clipboardData = new DataTransfer();
             clipboardData.setData("text/plain", text);
-            clipboard._handleCopy(new ClipboardEvent('paste', { clipboardData }));
+            const clipboardEvent = new ClipboardEvent('paste', { clipboardData });
+            // Force initialization since the constructor is broken in Firefox
+            if (!clipboardEvent.clipboardData.items.length) {
+                clipboardEvent.clipboardData.items.add(text, "text/plain");
+            }
+            clipboard._handleCopy(clipboardEvent);
             if (navigator.clipboard.writeText) {
                 expect(navigator.clipboard.writeText).to.have.been.calledWith(text);
             }
@@ -45,7 +50,12 @@ describe('Automatic Clipboard Sync', function () {
         if (Clipboard.isSupported) {
             const clipboardData = new DataTransfer();
             clipboardData.setData("text/plain", text);
-            clipboard._handlePaste(new ClipboardEvent('paste', { clipboardData }));
+            const clipboardEvent = new ClipboardEvent('paste', { clipboardData });
+            // Force initialization since the constructor is broken in Firefox
+            if (!clipboardEvent.clipboardData.items.length) {
+                clipboardEvent.clipboardData.items.add(text, "text/plain");
+            }
+            clipboard._handlePaste(clipboardEvent);
             if (navigator.clipboard.readText) {
                 expect(navigator.clipboard.readText).to.have.been.called;
             }


CendioOssman:
Looks nice and clean, so a good start. But if browser support is too poor then I don't think we want to merge it yet.

I'm not sure how paste is supposed to work even with that chrome bug fixed. We intercept Ctrl+V, so what will be firing the paste event?

juanjoDiaz:
This works in 2 ways:

1. The `Clipboard` class captures the `ClipboardEvent` on the canvas of type `paste` and paste the text content to the server using RFB's `clipboardPasteFrom` method.
2. The cut/copy events from the server that are handled by the RFB's `_handle_server_cut_text` method now also trigger a `ClipboardEvent` on the canvas of type `copy`. The `Clipboard` class captures the event and updates the local clipboard.

It's true that support varies. However, all browsers seem to be working on it and it's a change that doesn't break anything so I don't really see a reason to hold back.

samhed:
According to MDN this is only supported on Chrome and Opera. Not Firefox, IE, Edge or Safari. I think that this can result in a lot of confused users.

Where can I find information about all browsers working on this?

samhed:
ping @juanjoDiaz

juanjoDiaz:
Sorry @samhed, totally missed this.
At the moment I've moved to work on very different stuff so I'm not actively looking into this.

You are right. It seems that:
* Firefox currently only considers the cliboard support for extensions: https://bugzilla.mozilla.org/show_bug.cgi?id=1344410
* Edge seems to support it according to https://github.com/ckeditor/ckeditor4/issues/468 but I haven't tested it.
* There are references in different blogs and unofficial channels of Safari working on this but I couldn't find anything official.


One thing to keep in mind is that this PR was mean to be just the foundation of a possible clipboard functionality and it just uses the Clipboard Async API if available.
It could be extended to fallback to the Sync API if available or to use browser-specific workarounds (like https://stackoverflow.com/questions/40147676/javascript-copy-to-clipboard-on-safari) if possible.
We could look into a library like https://github.com/zenorocha/clipboard.js (just to see the approaches that they use, since you guys don't like external dependencies) to maximize browser compatibility.
This page also seems to be useful for browser testing https://whatwebcando.today/clipboard.html

samhed:
> https://github.com/zenorocha/clipboard.js

The list of supported browsers there looks great. If they are able to achieve that so should we be! What's the state of this code at the moment?

juanjoDiaz:
The async clipboard support (which is what's used in this PR) is now also supported in Safari: https://webkit.org/blog/10855/async-clipboard-api/

Firefox keeps having it as an open issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1619251

Just FYI.

TeffenEllis:
@juanjoDiaz @samhed,

This PR looks like it would get noVNC much closer to an expected experience -- especially now that Chrome, Edge, and Safari support the async Clipboard API. If browser support is a blocker, I could follow up with a new PR that:

- Add an browser API check when defining the clipboard handler 
- Log a console warning if either copy/paste event is unsuccessful
- Update the docs and example to include info about noVNC's clipboard API and its requirements

Thoughts?

samhed:
I'm opposed to merging something that isn't supported by all major browsers, Firefox is important.

I don't have the time to investigate this personally right now, but https://github.com/zenorocha/clipboard.js claims support for Firefox as well, couldn't we do what they do?

TeffenEllis:
@samhed from what I understand, Firefox's limitations are mostly around [async clipboard events](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) e.g. when noVNC wants to copy data to the user's clipboard. 

Clipboard.js depends on a "user originating" event to perform a clipboard copy. I suppose a stop-gap would be to offer a button in the UI to let the user retrieve the server's clipboard data.

AdrianSanchezLopez:
Hi guys, this feature looks really neat, I was wondering if you have any ETA on this

snickell:
@CendioOssman this is something we'd love to have for our novnc install, and I'm happy to do the dev work necessary if there's any requests you have.... any sense what you'd want to see here (besides no active merge conflicts and it actually tested and working) to consider a merge?

CendioOssman:
What I'd like to see broad support in the browsers. I.e. this works on at least Chrome, Firefox and Safari (Edge would be nice too though).

Failing that, a clean fallback handling until missing pieces are in place in all browsers.

I haven't kept track of this so I'm afraid I don't know what the current state of the browsers is, or how well this code matches that. Some testing would be helpful to get a better picture.

TeffenEllis:
@CendioOssman, @samhed,

I’ve since taken a sabbatical from tech, but occasionally get pinged on this PR.

Respectfully, I think this isn’t about browser support, but a lack of buy-in from the noVNC team.

The async clipboard API has similar [browser support](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#browser_compatibility) as the day this PR was opened nearly two years ago, and yet Mozilla drags their feet on any real progress. Worse, their marketshare continues to shrink into the single digits thanks to misstep after misstep. Meanwhile, every noVNC user is stuck in limbo with a subpar clipboard experience and told to wait for a perfect solution while this PR has been left to wither on the vine. 

Ultimately this is a volunteer project and nobody can force the team to do anything they don’t want to do. In my opinion, an outside contributor would be wary of putting time into this when the team has shown no interest in championing a feature so core to the experience, that it would move noVNC from a novelty web app to a productive tool used by professionals. 

Anyway, I’m on break from tech to write erotica. With peace and love, I hope this feature finds its home in the project.

samhed:
I'd love to see this merged. Assuming there has to be some sort of limitation for Firefox users, could we boil down some ideas for how those might look like or work in detail? @TeffenEllis you mentioned a button to retrieve the server clipboard?

Looking at Mozilla's browser support list it looks like other browsers shouldn't pose any problems. I'd feel at ease however if someone could assist with some testing anyway.

snickell:
@samhed @CendioOssman thanks for the encouragement, its good to know this PR is of interest, but is mainly blocking (at the moment) on ensuring we understand+test the cross-browser impact and, in particular, make sure that whatever we come up with has a reasonable and understood effect on Firefox user's too.

snickell:
If I can find time in the next 3 days, I plan to test and document current UX behavior for this PR against `noVNC:master` in:
1) Firefox
2) Chrome
3) Safari
4) iOS
5) Android

Since I will need to update the PR to merge with current `master` (at a minimuum), with @juanjoDiaz 's permission (or no response), I'm going to fork the PR to my GH and work there.

I'd love a volunteer to test on Edge once I get my branch merging? Or if there's a trivial no-cost cloud way I can test on Edge I'm happy to do that too.

diemol:
I'd be happy to help by setting an open source account in Sauce Labs for this project. Maintainers, please let me know if there is interest in this.

snickell:
@diemol not a maintainer, but from my perspective, that would be pretty amazing to avoid PRs blocking on time consuming to setup cross-browser checking

snickell:
Just to establish a baseline, on my system @juanjoDiaz 's unchanged branch https://github.com/juanjoDiaz/noVNC/tree/add_clipboard_support has one test failure from `npm test`:
```
SUMMARY:
✔ 431 tests completed
ℹ 4 tests skipped
✖ 1 test failed

FAILED TESTS:
  Mouse Event Handling
    Double-click for Touch
      ✖ should not modify 2nd tap pos if far apart
        Chrome 91.0.4472.114 (Mac OS 10.15.7)
      AssertionError: expected 68 to not equal 68
          at Mouse.mouse.onmousebutton (tests/test.mouse.js:139:41)
          at Mouse._handleMouseButton (core/input/mouse.js:104:14)
          at Mouse._handleMouseDown (core/input/mouse.js:115:14)
          at Context.<anonymous> (tests/test.mouse.js:150:19)

12 07 2021 14:44:11.249:WARN [Chrome 91.0.4472.114 (Mac OS 10.15.7)]: Disconnected (0 times) Client disconnected from CONNECTED state (server shutting down)
Chrome 91.0.4472.114 (Mac OS 10.15.7) ERROR
```

snickell:
I've resolved the merge conflict and merged with current master here: https://github.com/ceresimaging/novnc/tree/add_clipboard_support

After the merge all the `npm test` pass on my local system:
```
SUMMARY:
✔ 504 tests completed
ℹ 6 tests skipped
```

snickell:
I've tentatively submitted a new PR with the `noVNC:master` merge conflict resolved:
https://github.com/novnc/noVNC/pull/1562

this is 100% entirely @juanjoDiaz 's hard work, I just want to clear the way for a mergeable PR that resolves issue #1511 , I have time and need for this now, and this seemed like the easiest way forward. I'm happy to do something else if @samhed wants. 

Unfortunately, I don't think the `from branch` on a GH PR is editable, so it seems a new PR may be required to merge anyway unless @juanjoDiaz wants to push this branch forward? Last time in comments it seemed like no from him.

samhed:
Thanks for moving this forward @snickell, I'm closing this PR in favor of #1562

ChristianSB24:
I have tested this on Chrome, Edge, Android, and Firefox. Don't have a way to test Safari or iOS at the moment. This works great right away for me. I can copy to and from the vnc server. Seems like whatever browser issues there were are now possibly gone.

samhed:
Interesting @ChristianSB24, thanks for testing! I can see if I can find some time to test Safari on iOS.

juanjoDiaz:
Happy to see that this is relevant almost 3 years later 😅

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API),  support should be now there for most mayor browsers.

And, as I said back then, adding functionality to copy-paste that works in most browsers, and doesn't break anything in browsers without support should be a positive addition. Plus, having a clipboard module is a good foundation for future improvements and to add possible workarounds and support for other browsers.

I'd be happy to rebase this change and take a look at anything else needed if there is an intention to get this in 😊

samhed:
Yes, it's been too long, we'd love to get this feature merged!

I can see that the MDN page has marked Firefox with “No support” regarding `readText()` which we use here. But it seems to work for @ChristianSB24 in Firefox despite this?

I have tested here, but I assume I'm doing something wrong, since **it's not working at all for me**.

I used the following combinations:
* Safari on iPadOS 15
* Google Chrome on Android 12
* Google Chrome on Fedora 36
* Firefox on Fedora 36
* Google Chrome on Windows 10
* Firefox on Windows 10

I used the TigerVNC server, and tested with its own vncviewer, which worked fine. Furthermore, I also verified that noVNC's old method of using the clipboard panel worked fine.

Some light debugging shows that only the “Standard msg” clipboard format is supported by this PR, not the “Extended msg” format. Breakpoints in rfb.js shows that, when I select text and copy it inside the remote session, the code ends up using the “Extended msg” part of `_handle_server_cut_text()`.

I'm not sure what goes wrong when I try to copy locally and instead paste into the remote, but that doesn't work either.

ChristianSB24:
@samhed Interesting, it worked right away for me. If we could merge this in that would be amazing. novnc is an awesome project, but adding these kinds of improvements will really help advance it. 

Not sure if I am missing something, but this page says there is support for every major browser: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText. It doesn't support .read(), but for just pure text this is fine. 

For my setup I used tightvnc and I tested for all the previously mentioned browsers on a Windows 10 PC.

samhed:
@ChristianSB24 thanks for providing more details, I'll test with a tightvnc server tomorrow.

Yeah you are right, it does list it as supported in all browsers. My bad.

However, there is an asterix for Firefox on the page you linked:

> Firefox only supports reading the clipboard in browser extensions, using the "clipboardRead" extension permission

That is concerning since noVNC isn't an extension. But you're positive that clipboard.readText (paste to remote) worked when you tested Firefox?

A different subject is the Permissions API. MDN seems to indicate we need to prompt the user for permission in order to read the clipboard. Are you guys getting such prompts automatically? Or is this something we need to implement?

ChristianSB24:
@samhed I could have sworn it worked for all of the browsers I mentioned. I have tested some more tonight and now it is being inconsistent. I can still copy from the vnc server, but now I can't always copy from local to the vnc server. If I refresh enough and clear my cache it seems to sometimes work. 

This is for both firefox and chrome. When I first tested things it worked without an issue. Unfortunately I don't have the time to investigate further, but I assume the issue is related to what you mentioned.

juanjoDiaz:
Hi @samhed ,

How can I run te tests locally?

If I run `TEST_BROWSER_NAME=ChromeHeadless npm test`, I get
```sh
HeadlessChrome 106.0.5249 (Mac OS X 10.15.7) ERROR
  Uncaught SyntaxError: Cannot use import statement outside a module
  at tests/test.base64.js:3
```
this used to work before

samhed:
I usually do `npm test` and then open the URL it outputs in my browser.

gwhizoftv:
Hi Juan, et al,

It would be important if copy/paste from within a noVNC connection to the
local desktop and vie versa would work. It does not work now.

Please let me know what people think.

CendioOssman:
Odd. Headless works for me. This is Linux though...

```console
$ TEST_BROWSER_NAME=ChromeHeadless npm test

> @novnc/novnc@1.3.0 test
> karma start karma.conf.js


START:
18 10 2022 10:54:00.428:INFO [karma-server]: Karma v6.3.17 server started at http://localhost:9876/
18 10 2022 10:54:00.430:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
18 10 2022 10:54:00.434:INFO [launcher]: Starting browser ChromeHeadless
18 10 2022 10:54:01.149:INFO [Chrome Headless 106.0.5249.119 (Linux x86_64)]: Connected on socket L1W4IgzTAnR6kqTUAAAB with id 78656409
  Base64 Tools
    encode
...
```

juanjoDiaz:
When running `npm install` I get the following error on the prepublish script:
image

And when running `TEST_BROWSER_NAME=ChromeHeadless npm test` I get:
image

And when running and opening the browser as @samhed  suggested I get the same:
image


I'm on a MacBook and use Node v18.0.0.

Maybe I'm getting different dependencies due to the lack of versioning on the dependencies and the lack of package-lock.json in the project?
Looking at what actually gets installed, it seems that I'm getting Karma 1.7.1 whereas the latest version is 6.4.
Could you share your package-lock.json?

juanjoDiaz:
Indeed. Re-installing Karma manually re-installed the latest version and fixed the issue.
I'll open a separate issue about that.

juanjoDiaz:
Can you guys test after the rebase and after adding the permissions API?
Everything seems to work on my end.

samhed:
I have finally found some time to do some more testing here. On the server side, I have tested using TurboVNC together with websockify running on Fedora 36. I set up a self-signed certificate and made sure the browsers trusted it.

On the client side, I ran noVNC on several platforms:

| OS | Browser | Version | serverCut | clientCut |
|---|---|---|---|---|
| Fedora 36 | Chrome | 106 | ✔️ | ❌ |
| Fedora 36 | Firefox | 105 | ✔️ | ❌ |
| Windows 10 | Chrome | 106 | ✔️ | ❌ |
| Windows 10 | Edge | 106 | ✔️ | ❌ |
| Windows 10 | Firefox | 105 | ✔️ | ❌ |
| macOS 13 | Safari | 16.1 | ❌ | ❌ |
| macOS 13 | Chrome | 106 | ✔️ | ❌ |
| macOS 13 | Firefox | 105 | ✔️ | ❌ |
| Android 12 | Chrome | 106 | ✔️ | ❌ |
| Android 12 | Firefox | 105 | ✔️ | ❌ |
| iPadOS 15.7 | Safari | 106 | ❌ | ❌ |

As illustrated by the above table, "clientCut" (copying from the client machine and pasting into noVNC) doesn't work at all for me. I'm not getting any permission prompts, and breakpoints show that `clientCutText()` in rfb.js is never called. Do you have any ideas on what I need to do differently here, to get "clientCut" to work @juanjoDiaz and @ChristianSB24?

Furthermore, "serverCut" (copying from the remote and pasting on the client machine) also doesn't work in Safari on either macOS 13 or iPadOS 15.7.

Notes:
* I had to connect to noVNC using `https://` and with a full FQDN, matching what was specified in my certificate, for serverCut to start working.
* Using TigerVNC server the new clipboard features doesn't work at all, but this is because it sends extended clipboard messages.

samhed:
@juanjoDiaz or @ChristianSB24, any ideas on what you're doing differently? I can't get `clientCut` to work at all.

I'd love to get this moving forward.

juanjoDiaz:
Hi @samhed ,

I've added support for the extended messages since that seemed a low hanging fruit.

I'm quite busy this week with some other stuff. So I don't have time to look into the serverCut issue at the moment.
I'll try to take a look next week.

To help you troubleshooting since you already have the setup in place.
The order to set breakpoints should be:
* the `_handlePaste` method from Clipboard.js detect paste action and calls
* the `clipboardPasteFrom` method from Rfb.js process the text and calls
* the `RFB.messages.clientCutText` method that sends the data to

Is the event detected  by `_handlePaste`?
Is the `_handlePaste` method able to extract the text from the clipboard?
Is the `clipboardPasteFrom` able to created the Uint8Array?


Obviously, for this to work, the canvas must be the focused element in the page.

samhed:
Thanks for adding extended message support, I can confirm that it works great with TigerVNC server!

Still can't get `clientCut` to work at all, the `_handlePaste` method in clipboard.js never gets called :disappointed:

juanjoDiaz:
Thanks for the support @samhed!

`serverCut` works great because any time that something is written to the server's clipboard a message is sent to the client and I can sync it to the client´s clipboard:
* Something is written into the clipboard (copy/cut)
* An event is sent to the client
* The client trigger a "copy" event
* The clipboard module captures the event and put the content in the client's clipboard

`clientCut`, the reverse, is not so trivial because there is no way to detect from the browser when something is copied/cut on the client.
We could listen to when something is written to the clipboard from the current browser tab. But if the text is copied from other window, the browser nav bar, or anywhere else, there is no event that we could listen to.

My first solution to solve this limitation was to listen to the "paste" event on the canvas element so whenever the user paste something, I read the clipboard and pass that to the server as a `clientCutText` event.
However, it seems that the canvas object is not able to capture the paste event, as you noticed.

My second solution was to capture the paste event in the window and just check if the canvas is the focused element.
This seems to work sometimes but not reliably. I guess that whether it works or not depends on whether the `clientCutText` reaches the server before or after the key presses.
Also, this means that pasting by right-clicking and selecting paste, and other options like that won't work.

A third solution that I'm considering now is to run a function on an interval (e.g. every 200 ms) to sync the client's clipboard with the server's. This would work as long as the server interact first with the page (which shouldn't be a problem) but it might be a bit intensive to have such code running constantly on the page.

I'm open to other ideas.

juanjoDiaz:
@samhed , you seem to have a more complete setup for testing than I do.
Can you test the code in this PR (the second approach).


Also, what do you guys think about the `setInterval` to sync the client clipboard to the server continuously similar to how `serverCut` works?

alexandru-bagu:
> @samhed , you seem to have a more complete setup for testing than I do. Can you test the code in this PR (the second approach).
> 
> Also, what do you guys think about the `setInterval` to sync the client clipboard to the server continuously similar to how `serverCut` works?

Whenever we have to deal with the clipboard we probably have to copy stuff outside of the noVNC window. We can probably use window blur and focus (or maybe some other) events no? On focus we should save the current clipboard text and on focus check whether that changed. If it did then we should probably push the new clipboard text to vnc. That should be more efficient than an interval.

samhed:
Thanks for being patient, and for working on this. I have had time to do some more testing now!

| OS | Browser | Version | serverCut | clientCut | permission popup |
|---|---|---|---|---|---|
| Fedora 36 | Chrome | 107 | ✔️ | ✔️ | ✔️ |
| Fedora 36 | Firefox | 107 | ✔️ | ❌ | - |
| Fedora 36 | Edge | 107 | ✔️ | ✔️ | ✔️ |
| Windows 10 | Chrome | 107 | ✔️ | ✔️ | ✔️ |
| Windows 10 | Firefox | 107 | ✔️ | ❌ | - |
| Windows 10 | Edge | 107 | ✔️ | ✔️ | ✔️ |
| macOS 13 | Safari | 16.1 | ❌ | ❌ | - |
| macOS 13 | Chrome | 107 | ✔️ | ✔️ | ✔️ |
| macOS 13 | Firefox | 107 | ✔️ | ❌ | - |
| macOS 13 | Edge | 107 | ✔️ | ✔️ | ✔️ |
| Android 13 | Chrome | 107 | ✔️ | ✔️ | ✔️ |
| Android 13 | Firefox | 107 | ✔️ | ❌ | - |
| Android 13 | Edge | 107 | ✔️ | ✔️ | ✔️ |
| iPadOS 16.1.1 | Safari | 16.1.1 | ❌ | ❌ | - |

Things work great now in Chrome and Edge on all platforms. The focus-method is really nice, it feels natural!

In Firefox, only `serverCut` works. Debugging shows that `navigator.clipboard.readText` is undefined in Firefox. However, it seems they are working on this:
https://bugzilla.mozilla.org/show_bug.cgi?id=1770358

If I enable `dom.events.asyncClipboard.readText` in Firefox's about:config I can see a “Paste” button appearing (only in windowed-mode) in the bottom-right corner. Hopefully, things start working in Firefox land soon.

In Safari, nothing works, neither on macOS nor on iPadOS. However, I get the impression that it **should** work. I don't fully understand what the problem is, but they are talking about something like this here:
https://developer.apple.com/forums/thread/691873

samhed:
More info about Firefox from https://bugzilla.mozilla.org/show_bug.cgi?id=1776829:

> > I hope this will be fixed in 106 cycle. But dom.events.asyncClipboard.readText is still turned off even if desktop.
>
> Yup, we're still addressing some issues blocking turning dom.events.asyncClipboard.readText on on desktop. We plan to turn this feature on on both desktop and android the same time

It seems Firefox's approach will mean showing a floating "Paste" button at the cursor position. That button must then be pressed before the page gets access to the device's clipboard.

samhed:
More info about Safari, copy/paste seems to work well if done correctly. I don't know what's wrong with the code in this PR, however.

When testing this fiddle in Safari on an iPad things work well, both Copy and Paste:
https://jsfiddle.net/developit/c0aw4sua/

Note that a “Paste” button, like the one discussed for Firefox, appears in Safari.

juanjoDiaz:
After some investigation I found out that the problem with safari is that interacting with the clipboard only works from user actions like click, pointerdown, etc. However, serverCut events can arrive at any times and the clientCut pasting is don using the focus event.

A possible workaround is to keep a fake clipboard in memory and use user actions to actually do the clipboard interactions.

juanjoDiaz:
I've added some extra logic to handle the sync in Safari bit I'm having some issues testing related to permissions.
Maybe it's because of you mentioned regarding a FQDN.

In any case, the logic works when running on a CodePen.
Can you give it a go?

samhed:
Thanks for looking at this. However, now in Safari on macOS, whenever I click anywhere – the click gets “stuck” and I get a floating [PASTE] button appearing next to the mouse pointer. This makes things unusable, see this screen recording:

https://www.youtube.com/watch?v=NdNKz-d0u0g&ab_channel=SamuelM

cebas:
Is there a chance you disable the support for Safari and then do the merge? I'm using chrome and firefox and would love to have this working!

samhed:
No, we want this working across the board. If you want to get things moving more quickly, you can help out. We need to troubleshoot the Safari issue and find a fix.

cebas:
> No, we want this working across the board. If you want to get things moving more quickly, you can help out. We need to troubleshoot the Safari issue and find a fix.

I understand. I would like to help, but I'm a Linux user, so I can't help with Safari testing.

LiuTangLei:
Very much looking forward to this feature can be launched as soon as possible

cebas:
All checks have passed.  We're closer to have clipboard support! Thanks @juanjoDiaz

juanjoDiaz:
Hi all,

I've rebased and reverted the hacks that I tried for Safari.

This PR has been open now for 3.5 years. Browsers have evolved a bit, but the the problems not so much 😅

Firefox continues to not support reading from clipboard. They seem to be working on it but it will be based on the annoying "Paste" button as Safari is. See https://bugzilla.mozilla.org/show_bug.cgi?id=1770358

Safari continues to work based on the "Paste" button. So, with the current implementation based on the `focus` event. Whenever you focus on the canvas you get a "Paste" button. If you click on that button, then your data gets copied over. See https://webkit.org/blog/10855/async-clipboard-api/.

So, not much else that I can do here.
I could add a condition so the focus event is not used in Safari and Firefox to avoid the annoying "Paste" button. But that's about it.

juanjoDiaz:
> Clicked outside the noVNC remote screen, but still in Safari, then clicked back in noVNC remote
I got a paste button that wouldn't go away, when I clicked it, it went away for half a second then reappeared
After spamming different buttons on the mouse for a while, it finally gave up showing the paste button (after ~15 seconds perhaps?)

This is the "onFocus" workaround.
We sync the clipboard during the focus event.
On Safari, that triggers the "paste" button.

This is why  suggested disabling the pasting completely in Safari.

> I also tested Safari on iPadOS, it still doesn't seem to work at all sadly. As I wrote in an earlier post, it seems it should be possible:
>
> When testing this fiddle in Safari on an iPad things work well, both Copy and Paste:
https://jsfiddle.net/developit/c0aw4sua/

No, it doesn't.
It works if you copy using the "copy" button and paste using the "paste" button.
If you copy text for somewhere else and press the "paste" button, you get the annoying "paste" floating menu.
image


So, all in all, Chrome works so nicely because of the "focus" workaround.
Safari (and in the future Firefox seems like will follow the same approach) prevent this type of workarounds by showing the "paste" floating menu.

So the real question is: what do we do with pasting on Safari and Firefox?
They are trying to actively limit it so it only works if the user press a "paste" button and then the "paste" floating menu; i.e. we can only achieve this by adding a "paste" button on the noVNC UI. But it won't work as "natively" as Chrome does.

Wdyt?

juanjoDiaz:
> Sorry, I wasn't clear. I have no problem at all with the concept of a floating “paste” button. I think it is an acceptable inconvenience when weighed against the safety issues Chrome users have to face.

hehe we are not understanding each other here 😅

> The focus handler seems to trigger too much. My guess is that clicking the floating focus button briefly removes focus from the page (I have not verified this).

That's it!
The problem is on how to sync the local and remote clipboard.
Due to security restrictions, all browsers require that you do it as part of a user action.
**Which action should we use?** Based on the discussion above, I opted for sync the clipboard during the focus event.
That works great in Chrome but means that every `focus` event on the canvas will show the "Paste" button.
And for some reason, the canvas seems to focus/blur quite often.

Should we use the `click` event?
That doesn't sound much better. Every user click will show the paste button.

Keypresses don't sound great either...

Which other event do we have that we can use?


So, it seems that adding a button to noVNC UI is the only working approach here... 😕

juanjoDiaz:
Hi!

Sorry, I didn't see your answer.

AFAIK, we can't detect if the browser is using the "paste" button other than detecting if the browser is chrome, firefox or safari.

samhed:
In case we expect the "paste" button to appear on focus, perhaps we can approximately detect it by seeing if we immediately lose focus again?

The next steps should be to solve these two issues:

 * limit the "paste" button spam
 * ensure the "paste" button is visible even if remote screen covers the entire browser window

spags-lacework:
Hi, is there any recent update on the status of this PR?

I0x0I:
It seems this issue is resolved in a forked version [kasmtech/noVNC](https://github.com/kasmtech/noVNC/)

MaximMonin:
Firefox 127 added more clipboard api support

pleandre:
@samhed can you merge this PR. it works in Chrome.
I don't see why you are being a pain to merge it, it seems you just don't want it from the start. Even if it doesn't work in all the browsers it's still an improvement. So many issues are requesting this.

pleandre:
In Apache Guacamole, they seems to use a mix of the Asynchronous Clipboard API and some hacky way using a textarea set focus inside select all and call document.execCommand('copy')

https://github.com/apache/guacamole-client/blob/main/guacamole/src/main/frontend/src/app/clipboard/services/clipboardService.js#L192

pleandre:
In Apache Guacamole, they seems to use a mix of the Asynchronous Clipboard API like you do and some hacky way with a textarea as a fallback
https://github.com/apache/guacamole-client/blob/main/guacamole/src/main/frontend/src/app/clipboard/services/clipboardService.js#L192

And they sync clipboard on these events
https://github.com/apache/guacamole-client/blob/main/guacamole/src/main/frontend/src/app/index/controllers/indexController.js#L276

image

Their explanation: https://guacamole.apache.org/faq/#clipboard
image

This is what Apache team ended up to after 6 or 7 PRs and 8 years of evolution.
https://issues.apache.org/jira/browse/GUACAMOLE-55
https://issues.apache.org/jira/browse/GUACAMOLE-310
https://issues.apache.org/jira/browse/GUACAMOLE-128
https://issues.apache.org/jira/browse/GUACAMOLE-559
etc.
Their implementation also supports copy pasting images.

samhed:
Sounds promising that Firefox added more support in 127! I would love to see this get merged.

I'd be happy to help out with testing if the PR is updated to address the issues mentioned above:

> In case we expect the "paste" button to appear on focus, perhaps we can approximately detect it by seeing if we immediately lose focus again?
> 
> The next steps should be to solve these two issues:
> 
> * limit the "paste" button spam
> * ensure the "paste" button is visible even if remote screen covers the entire browser window

samhed:
@pleandre thank you for looking into what Apache achieved! If you are up for it, feel free to open a new PR. We'll help out as best we can.

williamsjoblom:
Below I have tried to summarize the discussion, as well as doing some testing on
my own to hopefully gain some traction in the issue.

1.  [safari] Paste prompt is intrusive
    
    As per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API), Safari will show a paste prompt before the browser gets to access
    the current content of the clipboard.
    
    In Firefox, this prompt is rather non-intrusive as it pops up on the bottom
    of the screen. In Safari on the other hand, it pops up next to the mouse,
    covering the VNC viewport. This in itself is a bit annoying.
    
    To make matters worse, the current implementation in this PR tries to read
    the clipboard memory every time the viewport canvas gains focus. This, in
    combination with the fact that Safari lets go of the focus when displaying
    the paste prompt, make the paste prompt show up all the time.
    
    In contrast, when Firefox shows its paste prompt, the canvas does not lose
    focus.
    
    This problem is very much a blocker for merging this PR as highlighted by
    previous commenters.
    
    One possible way around this would be to check if the browser window has lost
    and gained focus since the last time the canvas gained focus. I'm not 100%
    certain that this will work, but it may be worth a shot.

2.  [firefox] Paste prompt is easy to miss
    
    Similar to Safari, Firefox will also show a paste prompt. This one, however,
    is easy to miss as it is displayed at the bottom of the window.

3.  [firefox, safari] "Paste" prompt is confusing in this context
    
    In both Safari and Firefox (that prompt the user before giving up the current
    state of the clipboard memory), this prompt says "Paste". This is not intuitive
    at the slightest as clicking does not paste anything. All it does is sending
    the current client clipboard to the server.
    
    This has a high potential to confuse new users.

4.  [general cleanup] Copying from server to client is triggered by the canvas
    `copy` event. These events can both be fired by the browser itself when it
    detects a client-native copy actions and programmatically when the receiving a
    `ServerCutText` message. The `copy` event handler then calls
    `clipboard.writeText`.
    
    Based on some testing, I have not been able to get the browser (Firefox 127
    and Chrome 126) to fire these events. All events I have observed has been
    fired from the `ServerCutText` handler in `rfb.js`. If the browser never
    fires these events when copying text inside the session, does it really make
    sense to indirectly call `Clipboard._handleCopy` from
    `RFB._handleServerCutText` by firing an artificial event?
    
    This is further strengthened by Apaches reasoning to not rely on the browser
    fired `copy` events: <https://guacamole.apache.org/faq/#clipboard>.
    
    In my opinion, directly calling `Clipboard._handleCopy` from
    `RFB._handleServerCutText` would remove this complexity.

To conclude, Chrome currently works without a hiccup and gives a great user
experience. Safari and Firefox (the browser with the largest market share after
Chrome) is more problematic.

My suggestion would be to consider Chrome as a separate entity than Safari and
Firefox. That is, use the work in this pull request exclusively in Chrome, and
treat Safari and Firefox separately.

williamsjoblom:
One way we could differentiate between the Chrome way of doing things and the Firefox/Safari way would be to look for the presence of the `clipboard-read` and `clipboard-write` permissions. If they are present, the browser in question is very likely following Chromes example.

That is, of course, if we are dead set on avoiding looking at the user agent or `window.isChrome`.

williamsjoblom:
Worth noting is that Firefox has this issue still open, indicating that the UX of the paste button may change in the future:

  https://bugzilla.mozilla.org/show_bug.cgi?id=1777197

0xAungkon:
Man I Fixed This Automatic Clipboard Snyc Issue XD . Check the latest pull .


@juanjoDiaz  @CendioOssman  @samhed  @GirlBossRush  @AdrianSanchezLopez

liyimeng:
> Man I Fixed This Automatic Clipboard Snyc Issue XD . Check the latest pull .
> 
> @juanjoDiaz @CendioOssman @samhed @GirlBossRush @AdrianSanchezLopez
@0xAungkon what is your PR really do.  I deploy it. Expecting I can do copy on my local machine and paste it into the remote machine I connect via vnc, but nothing happen.

GirlBossRush:
I’m honestly quite stunned how such a slam dunk of a feature has become a Herculean effort of epic proportions. It’s an API call, not skunkworks moonshot. We’ve built a 10 billion dollar space telescope in less time. How long has the conversation been going on? Let’s take a look together…

[Six years ago](https://bugzilla.mozilla.org/show_bug.cgi?id=1312260#c54), the brain trust at Mozilla decided that their users are too foolish to be trusted with the decision of complete asynchronous clipboard access. The guards hold firm for another two years, [renewing their vow](https://bugzilla.mozilla.org/show_bug.cgi?id=1560373) to never let their flock be duped into self destruction. Wow!

It’s sort of incredible if you stop to appreciate the impact of inaction. A global pandemic has circumnavigated Earth thrice in as many years. Empires have [risen and returned to dust](https://gs.statcounter.com/browser-market-share). And now in 2024, the software industry forages for answers in the ashes of a zero interest rate economy.

Is it really fair to blame all of this on a single brood of stubborn and unflappable butterflies? I’ve got a better one for you: Why would you presume fairness from a “non-profit” that uses your donation money to do everything but build the gosh damn browser?

There’s nothing wrong with pinning all of the bad things in the world on them because of a simple fact: Mozilla’s leadership holds your heart in contempt. The C-Suite and development teams do not prioritize what’s best for Firefox, your privacy, or anything else other than you sending them more money. 

Despite everything and anything, it’s important to put this half decade into perspective. Over a billion human lives have begun life on this planet, their formative years shaped by rapid technological advancements once exclusive to the future fantasies of only the most implausible science fiction novels. 

Indoor plumbing, Greek fire, MRNA vaccines, global satellite internet, LLM artificial intelligence. These are just a select few of the mundane miracles that stand in comparison to our greatest achievement: [Asynchronous clipboard APIs](https://caniuse.com/async-clipboard). Sometimes all you need is hope and a little bit of faith! 

But sadly there are at least odd several hundred people trapped in an inescapable realm beyond hope’s grasp. It’s called Mozilla’s pocket and it’s a dank, damp pit. Words fail to describe just how a  dark and sweaty are the conditions of this sour swamp for those unfortunate, wretched, coffin warmers. They enmeshed their hopes and dreams into Firefox like mollusks on the bottom of a beached whale on a hot summer day. They can’t even comprehend just how long their host has been a corpse. “How dare you say that!” they cry. “Sure, I might live in a slowly decaying husk, but if *I* won’t then *who* will?” Have some respect for yourself, please.

Really though, what’s even more gross is how open source users are thought of as pawns to defend Mozilla’s corporation as when it’s subsidized by millions of dollars from the most credible threat and direct competitor. We’re supposed to give them more money, and only target the features that they decide for us? This isn’t it freedom. It’s subjugation with extra steps!

Now you listen to me, and you listen good. You’re all circus clowns, and since I had the genius to leave thread  notifications on for the past four years ago, that makes me the ringleader. 

If you’re a maintainer, it’s takes about two hours to add the feature. Less with caffeine and prescription stimulants. It’s most definitely easier than it was to read this thread.

Everyone else in the thread: fork the repo and do not send a pull request. Leave this place and never return. Install a different browser. Take a vitamin. Try a different salad dressing while you still have time. 

All that matters is that person who’s reading this finds the courage to stand up for themselves and stop letting other people decide what’s best for them. Why? Obviously because I said so!

Now get to work! 👏 👏

grisha765:
Hi, is there any progress on this pull request?

sillynocodecat:
It's so needed and we still want to get this

anjunatl:
> efox continues to not support reading from clipboard. They seem to be working on it but it will be based on the annoying "Paste" button as Safari is. See https://bugzilla.mozilla.org/show_bug.cgi?id=1770358

I could be mistaken I'm pretty sure Firefox has everything fixed and ready

- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText - widely available
- https://bugzilla.mozilla.org/show_bug.cgi?id=1770358 - Closed metabug
- https://bugzilla.mozilla.org/show_bug.cgi?id=1619251#c8 - metabug still open for wrapup, but async api is merged: https://bugzilla.mozilla.org/show_bug.cgi?id=1887845

DaWe35:
It's been 6 years, can I help something to make this happen?

webysther:
Any fork that add this feature available? 5 years of blocking from opinions than the get things done.
To list of some projects with this in production for years now:
- Selkies: https://github.com/selkies-project/selkies/commit/0d61c6499fc87951f8933d898523c77a2832303d
- KasmVNC

Anyone knows any fork of noVNC with support to this feature to be used as drop-in replacement?

Neustradamus:
Dear all,

Have you seen the @tobfah PR?
- https://github.com/novnc/noVNC/pull/1993

What do you think?

samhed:
Closing this, fixed by #1993.

Neustradamus:
Linked to:
- https://github.com/juanjoDiaz/noVNC/pull/1
- https://github.com/novnc/noVNC/pull/1993
- https://github.com/novnc/noVNC/issues/1961
- https://github.com/novnc/noVNC/issues/1902
- https://github.com/novnc/noVNC/pull/1900
- https://github.com/novnc/noVNC/pull/1899
- https://github.com/novnc/noVNC/pull/1831
- https://github.com/novnc/noVNC/issues/1822
- https://github.com/novnc/noVNC/pull/1817
- https://github.com/novnc/noVNC/pull/1562
- https://github.com/novnc/noVNC/issues/1511
- https://github.com/novnc/noVNC/pull/1301
- https://github.com/novnc/noVNC/pull/1174

PR #1348, opened 2019-12-19
Limit MouseMove Messages to once per 100mS

uklatt:
Perhaps this should be adjustable from the settings page...

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -776,6 +776,10 @@ export default class RFB extends EventTargetMixin {
             const deltaX = this._viewportDragPos.x - x;
             const deltaY = this._viewportDragPos.y - y;
 
+            if( typeof _handleMouseMove.timestamp == 'undefined' ) {
+              _handleMouseMove.oldtimestamp = Date.now();
+            }
+            
             if (this._viewportHasMoved || (Math.abs(deltaX) > dragThreshold ||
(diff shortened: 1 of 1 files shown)


samhed:
Why do you want this? What problem are you seeing?

CendioOssman:
We saw issues with this on TigerVNC as well:

https://github.com/TigerVNC/tigervnc/commit/157f54202ee570b1f00e8f0e5db660750b548c3c

The timeout chosen there was 60 Hz/17 ms as that seemed like a safe bet. 100 ms might be noticeable.

You also have to have a timeout for a final update, or the cursor position might not be properly updated.

uklatt:
My "problem" was a nearly unusable mouse because many move events add a great lag.
UltraVNC client can limit this, novnc cant.
My patch solved this for me.
This should be adjustable in thesettings...

Uwe

CendioOssman:
I don't think this needs to be configurable as long as we put a decent value on it.

So please change to 17 ms, and add handling for a final update once the cursor stops moving. After that we can merge this.

The variables could also use some more specific naming so it's obvious what they are used for.

uklatt:
Closed because new version created

samhed:
For future reference, you can force-push to an existing branch in order to update an existing PR.

PR #1351, opened 2019-12-31
Added RealVNC's RA2ne security type

pdlan:
Supports RA2ne security type for compatibility with RealVNC servers


CendioOssman:
Many thanks for having a look at this. :)

Two initial things before we start a more detailed review:

 1. Have you tested this with IE 11? I think I saw some things in there that might not work on IE.
 1. Do you think you could write up how this is supposed to work for @rfbproto? We really want to have some form of specification to refer to when fixing things in the future.

pdlan:
1. Sorry that I haven't tested it with IE. I have modified the code to make it work though.
2. The protocol and encryption algorithm come from my reverse-engineering RealVNC's Viewer. So I can provide the specification (according to my analysis) later but maybe the specification will not be so precise (it works though).

pdlan:
[https://gist.github.com/pdlan/81245e3d83d7fa586b8dcbb4e2d507cc](https://gist.github.com/pdlan/81245e3d83d7fa586b8dcbb4e2d507cc)
A simple specification

CendioOssman:
Great, thanks. Do you think that code can be explained by referencing existing standards? Or have Real come up with their own custom crypto handling?

pdlan:
It seems to be Real's own algorithm.

samhed:
>https://gist.github.com/pdlan/81245e3d83d7fa586b8dcbb4e2d507cc
A simple specification

Nice! Could you make a PR for @rfbproto with this?

samhed:
One more comment, the code that you haven't written yourself like for example core/sha1.js should be moved to the vendor directory instead of core. We like to have this separation to make it clearer what's what.

samhed:
What's the status on this @pdlan?

pdlan:
Sorry for the delay. I refactored my code and moved the third-party code to the vendor directory. Sorry that I don't have time to write some documentation for the security type.

samhed:
@CendioOssman what's your take on this? Is it feasible to maintain this without matching documentation in rfbproto?

CendioOssman:
Maybe, although I'd like to avoid it given the complexity.

I also haven't tested it or done a final code review.

app/ui.js
```diff
@@ -1043,6 +1043,8 @@ const UI = {
         UI.rfb.addEventListener("clipboard", UI.clipboardReceive);
         UI.rfb.addEventListener("bell", UI.bell);
         UI.rfb.addEventListener("desktopname", UI.updateDesktopName);
+        UI.rfb.addEventListener("rsakeyget", UI.getRSAKey);
+        UI.rfb.addEventListener("rsakeyset", UI.setRSAKey);
```
CendioOssman, line 1047: These are used for authentication, so I think these could be handled by the credentials mechanism we already have.

app/ui.js
```diff
@@ -1718,6 +1720,15 @@ const UI = {
...
     },
 
+    getRSAKey(e) {
+        const key = WebUtil.readSetting("rsa_key");
+        UI.rfb.setRSAKey(key);
+    },
+
+    setRSAKey(e) {
+        WebUtil.writeSetting("rsa_key", e.detail.key);
+    }
```
CendioOssman, line 1730: I guess we need some UI for users to be able to provide the key as well.

core/ra2.js
```diff
@@ -0,0 +1,259 @@
+import "../vendor/aes.js";
+import "../vendor/sha1.js";
+import "../vendor/jsencrypt.js";
```
CendioOssman, line 3: We no longer support Internet Explorer, so please have a look at using SubtleCrypto instead so we can avoid these extra scripts.

core/ra2.js
```diff
@@ -0,0 +1,259 @@
...
+import Base64 from './base64.js';
+
+export class RA2Cipher {
```
CendioOssman, line 6: Given that you've figured out the algorithm used, can this be reorganised so it follows the same naming and structure as the algorithm's specifications?

core/rfb.js
```diff
@@ -967,6 +977,8 @@ export default class RFB extends EventTargetMixin {
                 this._rfb_auth_scheme = 2; // VNC Auth
             } else if (includes(19, types)) {
                 this._rfb_auth_scheme = 19; // VeNCrypt Auth
+            } else if (includes(6, types)) {
+                this._rfb_auth_scheme = 6; // RA2ne
```
CendioOssman, line 981: I guess RA2ne_256 could also easily be supported?

core/rfb.js
```diff
@@ -1271,6 +1283,127 @@ export default class RFB extends EventTargetMixin {
...
+            return false;
+        } else if (this._rfb_ra2_state.state === 1) {
+            // Receive server RSA public key
+            // Send RSA encrypted client random
+            if (this._sock.rQwait("ra2 public key", 516, 0)) {
+                return false;
+            }
+            if (!this._rfb_ra2_state.parse_server_rsa_key(this._sock.rQshiftBytes(516))) {
+                return this._fail("Invalid RSA key");
+            }
```
CendioOssman, line 1306: We need to let the user approve this key as well, so we can be sure we are talking to the right server.

I guess we need a new event for this. Or if we add a field to the credentials where approved server keys can be added.

core/rfb.js
```diff
@@ -1271,6 +1283,127 @@ export default class RFB extends EventTargetMixin {
         return this._fail("No supported sub-auth types!");
     }
 
+    _negotiate_ra2ne_auth() {
```
CendioOssman, line 1286: This handling is very complex, so we need some unit tests for this to make sure it won't break in the future.

CendioOssman (changes requested):
I see you've revisited this, so let's get a more detailed review in place. There are some things I'd like you to have a look at.

app/ui.js
pdlan: How about generating a new private key for every connection? So we don't need to bother saving the private key securely.

CendioOssman: That usually takes a very long time. Have you tested it and see how long it takes?

pdlan: I tried to generate a 2048bit rsa key using SubtleCrypto for 100 times and it only took about 6 seconds. So I think on modern browsers it's quite fast.

CendioOssman: Fair enough. Let's try that then.

pdlan:
Hi @CendioOssman. It seems that GitHub didn't send notification also for this PR. Could you have a look at this?

CendioOssman:
Sure. I'll try to have a look tomorrow.

core/rfb.js
```diff
@@ -137,6 +138,7 @@ export default class RFB extends EventTargetMixin {
         this._keyboard = null;          // Keyboard input handler object
         this._gestures = null;          // Gesture input handler object
         this._resizeObserver = null;    // Resize observer object
+        this._rfbRA2State = null;
```
CendioOssman, line 141: Authentication related attributes are further up. Better to keep it there to clarify what it is used for.

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+    get e() {
+        return this._e;
+    }
+
+    get d() {
+        return this._d;
+    }
+}
+
+export default class RA2State {
```
CendioOssman, line 308: I think we should use the same name as what we got in rfbproto as that is more descriptive.

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+    constructor(sock, onFinish, getCredentials, requireCredentials) {
```
CendioOssman, line 309: Have you looked at our `EventTargetMixin` class? It gives a bit more standard event based API for things like this.

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+            if (!this._sock.rQwait('RA2', len)) {
+                resolve();
+            } else {
+                this.addEvent(resolve, () => !this._sock.rQwait('RA2', len));
+            }
+        });
+    }
+
+    _waitCredentialsAsync(subtype) {
+        const hasCredentials = () => {
```
CendioOssman, line 329: Why not make this a normal method for clarity?

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+            if (subtype === 1 && this._getCredentials().username !== undefined &&
+                this._getCredentials().password != undefined) {
```
CendioOssman, line 331: `!==`

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+        return new Promise((resolve) => {
+            if (hasCredentials()) {
+                resolve();
+            } else {
+                this.addEvent(resolve, hasCredentials);
+            }
+        });
+    }
+
+    async negotiateRA2neAuthAsync() {
```
CendioOssman, line 347: This function needs a bit of spacing. It's very hard to follow right now. :)

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+        this._hasStarted = true;
+        // 1: Receive server public key
+        await this._waitSockAsync(4);
+        const serverKeyLengthBuffer = this._sock.rQslice(0, 4);
+        const serverKeyLength = this._sock.rQshift32();
+        if (serverKeyLength < 1024) {
+            throw new Error("RA2: server public key is too short: " + serverKeyLength);
+        } else if (serverKeyLength > 8192) {
+            throw new Error("RA2: server public key is too long: " + serverKeyLength);
```
CendioOssman, line 356: I don't see any handling to catch these and abort the connection?

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+        fingerprint = Array.from(new Uint8Array(fingerprint).slice(0, 8)).map(
+            x => x.toString(16).padStart(2, '0')).join('-');
+        if (subtype === 1) {
+            if (this._getCredentials().username === undefined ||
+                this._getCredentials().password === undefined) {
+                this._requireCredentials(true, fingerprint);
+            }
+        } else if (subtype === 2) {
+            if (this._getCredentials().password === undefined) {
+                this._requireCredentials(false, fingerprint);
```
CendioOssman, line 465: Even if the username and password are provided the UI still needs to approve the server key.

core/ra2.js
```diff
@@ -0,0 +1,514 @@
...
+        let subtype = (await serverCipher.receiveMessage(
+            1, this._sock.rQshiftBytes(1), this._sock.rQshiftBytes(16)));
+        if (subtype === null) {
+            throw new Error("RA2: failed to authenticate the message");
+        }
+        subtype = subtype[0];
+        let fingerprint = await window.crypto.subtle.digest("SHA-1", serverPublickey);
+        // The same fingerprint format as RealVNC
+        fingerprint = Array.from(new Uint8Array(fingerprint).slice(0, 8)).map(
+            x => x.toString(16).padStart(2, '0')).join('-');
```
CendioOssman, line 457: Doing a hash is a user interface decision. You should provide the entire key and then the UI can decide how this should be presented.

core/rfb.js
```diff
@@ -1623,6 +1627,31 @@ export default class RFB extends EventTargetMixin {
...
+        if (this._rfbRA2State === null) {
+            this._rfbRA2State = new RA2State(this._sock, () => {
+                this._rfbInitState = "SecurityResult";
+                this._rfbRA2State = null;
+                this._initMsg();
+            }, () => this._rfbCredentials, (hasUsername, fingerprint) => {
+                if (hasUsername) {
+                    this.dispatchEvent(new CustomEvent(
+                        "credentialsrequired",
+                        { detail: { types: ["username", "password"], fingerprint: fingerprint } }));
```
CendioOssman, line 1640: This API change needs to be documented as well.

And after thinking a bit more about things, I think it is better if we create a new event type for this. It avoids confusing things.

CendioOssman, line 1640: In order for the API to remain stable in the future you should also include a type with the server key. So that new types can be added for other things that need to be verified.

tests/test.ra2.js
```diff
@@ -0,0 +1,348 @@
...
+
+let realGetRandomValues = window.crypto.getRandomValues;
+let realGeneratekey = window.crypto.subtle.generateKey;
+
+describe('RA2 handshake', function () {
+
+    before(() => {
+        FakeWebSocket.replace();
+        window.crypto.getRandomValues = fakeGetRandomValues;
+        window.crypto.subtle.generateKey = fakeGeneratekey;
```
CendioOssman, line 89: You can probably use `sinon.stub()` for these, which is generally safer.

tests/test.ra2.js
```diff
@@ -0,0 +1,348 @@
...
+            expect(sentData).to.eql(sendData);
+            done();
+        }, () => credentials, (hasUsername, fingerprint) => {});
+        sock.on('message', () => {
+            state.poll();
+        });
+        state.negotiateRA2neAuthAsync();
+        sock._websocket._receiveData(receiveData);
+    });
+
```
CendioOssman, line 347: It is more interesting to test these things via the `RFB` object. There we can make sure events are fired properly and such things.

vnc.html
```diff
@@ -286,6 +286,10 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br>VNC</h1>
                 <label>Password:</label>
                 <input id="noVNC_password_input" type="password">
             </li>
+            <li id="noVNC_fingerprint_block">
+                <label>Fingerprint:</label>
+                <div id="noVNC_fingerprint"></div>
+            </li>
```
CendioOssman, line 292: There needs to be an option to approve or reject this key. And there should probably be a more descriptive text what this question means.

CendioOssman (changes requested):
Nice. This is a lot neater. :)

There are some API and interface things that needs a look at though.

samhed:
Hi @pdlan do you think you will have the time to revisit the comments by @CendioOssman ?

pdlan:
Sorry I’m busy with my own things recently. But I will revisit it when I have more spare time.

core/ra2.js
pdlan: hasCredentials uses subtype, so I put it inside _waitCredentialsAsync

pdlan:
H @CendioOssman, I made some modification, could you have a look?

core/rfb.js
```diff
@@ -1628,26 +1632,30 @@ export default class RFB extends EventTargetMixin {
...
-                        { detail: { types: ["password"], fingerprint: fingerprint } }));
-                }
+        if (this._rfbRSAAESAuthenticationState === null) {
+            this._rfbRSAAESAuthenticationState = new RSAAESAuthenticationState(this._sock, () => this._rfbCredentials);
+            this._rfbRSAAESAuthenticationState.addEventListener("serverkeyverification", (event) => {
+                this.dispatchEvent(event);
+            });
+            this._rfbRSAAESAuthenticationState.addEventListener("credentialsrequired", (event) => {
+                this.dispatchEvent(event);
+            });
```
CendioOssman, line 1642: Ideally we also have some cleanup of these listeners.

vnc.html
```diff
@@ -274,6 +274,25 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br>VNC</h1>
         </div>
     </div>
 
+    <!-- Server Key Verification Dialog -->
+    <div class="noVNC_center noVNC_key_verification_layer">
+    <div id="noVNC_verify_key_dlg" class="noVNC_panel"><form>
+        <ul>
```
CendioOssman, line 280: Using a list here is a bit odd. But I can see that this is already done for the credentials dialog, so I can do that cleanup later.

CendioOssman (changes requested):
This is starting to look really good! I can also verify that it works when testing here. Tried it against RealVNC 6.7.2 without issues.

There are a few more details and then we can merge this.

app/styles/base.css
```diff
@@ -260,6 +260,10 @@ select:active {
  * ----------------------------------------
  */
 
+.noVNC_key_verification_layer {
+  z-index: 55;
+}
```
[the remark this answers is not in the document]
pdlan: Sorry that I'm not an expert of frontend developing. When I tried to put them in the same layer, neither of them appeas in the center of the screen. I guess it's because the elements are hidden by `visibility: hidden` not `display: none` so I just put them in the different layers and it seems to work. Is there a better solution for that issue?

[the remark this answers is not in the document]
CendioOssman: Let me have a look and see what needs to be done.

core/ra2.js
```diff
@@ -365,6 +388,13 @@ export default class RA2State {
         serverPublickey.set(serverKeyLengthBuffer);
         serverPublickey.set(serverN, 4);
         serverPublickey.set(serverE, 4 + serverKeyBytes);
+
+        // verify server public key
+        this.dispatchEvent(new CustomEvent("serverkeyverification", {
+            detail: { serverPublickey: serverPublickey }
```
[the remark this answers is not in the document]
CendioOssman: Sound reasonable. But then we need to include the type somewhere so that the UI knows how to handle them when other types will be added. E.g. elliptic curves.

app/styles/base.css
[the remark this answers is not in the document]
CendioOssman: Ah, I see the problem. If you put them in the same `<div>` then the centering is broken, yes. But keep it in a separate `<div>`, just like you have now. Then just change the class from `noVNC_key_verification_layer` to `noVNC_connect_layer`.

app/ui.js
```diff
@@ -1152,6 +1157,35 @@ const UI = {
...
  * ==============
+ *   SERVER KEY
+ * ------v------*/
+
+    async serverKey(e) {
+        const serverPublickey = e.detail.serverPublickey;
+        let fingerprint = await window.crypto.subtle.digest("SHA-1", serverPublickey);
+        // The same fingerprint format as RealVNC
+        fingerprint = Array.from(new Uint8Array(fingerprint).slice(0, 8)).map(
+            x => x.toString(16).padStart(2, '0')).join('-');
```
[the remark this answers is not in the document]
pdlan: Actually I've found the (binary) code for generating it but if we need to support generating it we have to add some 10KB+ word lists, so I think it's not so necessary as we already have the fingerprint.

pdlan:
Hi @CendioOssman, I wonder if you're planning to add this feature to "notable contribution"? If so could you use the name "USTC Vlab Team" instead of my name?

pdlan:
Hi @CendioOssman, any update on this PR?

CendioOssman:
Very swamped with other things I'm afraid. This is on my todo though so it won't be forgotten.

app/ui.js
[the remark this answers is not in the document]
CendioOssman: Okay. Great.

CendioOssman (changes requested):
Looks great!

Please squash all the commits and I can go ahead and merge this.

CendioOssman:
> Hi @CendioOssman, I wonder if you're planning to add this feature to "notable contribution"? If so could you use the name "USTC Vlab Team" instead of my name?

Sure, we can do that.

CendioOssman (approved):

PR #1352, opened 2020-01-03
Limit mouse move events to one every 17 mS.

uklatt:
Limit mouse move events to one every 17 mS.

### DIFF
--- a/core/input/mouse.js
+++ b/core/input/mouse.js
@@ -11,6 +11,7 @@ import { setCapture, stopEvent, getPointerEvent } from '../util/events.js';
 const WHEEL_STEP = 10; // Delta threshold for a mouse wheel step
 const WHEEL_STEP_TIMEOUT = 50; // ms
 const WHEEL_LINE_HEIGHT = 19;
+const MOUSE_MOVE_DELAY = 17; // Minimum wait (ms) between two mouse moves
 
 export default class Mouse {
     constructor(target) {
@@ -22,6 +23,7 @@ export default class Mouse {
         this._pos = null;
         this._wheelStepXTimer = null;
         this._wheelStepYTimer = null;
+        this._oldMouseMoveTime = 0;
         this._accumulatedWheelDeltaX = 0;
         this._accumulatedWheelDeltaY = 0;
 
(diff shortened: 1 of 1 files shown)


core/rfb.js
```diff
@@ -790,8 +797,15 @@ export default class RFB extends EventTargetMixin {
 
         if (this._viewOnly) { return; } // View only, skip mouse events
 
-        if (this._rfb_connection_state !== 'connected') { return; }
-        RFB.messages.pointerEvent(this._sock, this._display.absX(x), this._display.absY(y), this._mouse_buttonMask);
+        // Limit mouse move events to one every MOUSE_MOVE_DELAY mS
+        clearTimeout(this.timerID);     // Remove old one
```
CendioOssman: `timerID` is a bit generic. Please include something that identifies what timer it is.

core/rfb.js
```diff
@@ -790,8 +797,15 @@ export default class RFB extends EventTargetMixin {
...
+        this.newMouseMoveTime = Date.now();
```
CendioOssman: This can be local as far as I can tell.

core/rfb.js
```diff
@@ -790,8 +797,15 @@ export default class RFB extends EventTargetMixin {
...
+        if ( this.newMouseMoveTime < this.oldMouseMoveTime + MOUSE_MOVE_DELAY ) {
+            this.timerID = setTimeout(this._setMousePosition, MOUSE_MOVE_DELAY, x, y);  // Set last mouse position after timeout
```
CendioOssman: I don't think this works properly. `this` would be incorrectly set when this timeout fires. Have you tested it?

On that note, some unit tests would be nice for this.

samhed:
Ping!

This would be nice to get merged, is there anything we can do to help get this moving forward?

core/rfb.js
samhed: fixed

samhed:
I moved the code to `core/input/mouse.js`, combined the commits, and fixed the comments by @CendioOssman.

core/rfb.js
samhed: fixed bind()

and added unit tests in a672168d4dac5735b40cfe618a47f76b69c85882

PR #1359, opened 2020-01-23
Remove generated HTML by Cursor when it detaches

### LINKED ISSUE #1358: Multiple pointers on the screen on iPad
**Describe the bug**
Running noVNC on an iPad for a while I end up with a bunch of mouse pointers all over the screen:

```html
<canvas style="position: fixed; z-index: 65535; pointer-events: none; left: 740px; top: 280px;" width="16" height="16"></canvas>
<canvas style="position: fixed; z-index: 65535; pointer-events: none; left: 446px; top: 521px;" width="16" height="16"></canvas><canvas style="position: fixed; z-index: 65535; pointer-events: none; visibility: hidden;" width="0" height="0"></canvas>
<canvas style="position: fixed; z-index: 65535; pointer-events: none; visibility: hidden;" width="0" height="0"></canvas>
<canvas style="position: fixed; z-index: 65535; pointer-events: none; left: 765px; top: 461px;" width="10" height="16"></canvas>
<canvas style="position: fixed; z-index: 65535; pointer-events: none; visibility: hidden;" width="0" height="0"></canvas><canvas style="position: fixed; z-index: 65535; pointer-events: none; visibility: hidden;" width="0" height="0"></canvas>
```

It seems to have to do with fast touches or multitouches.

Should the mouse pointer show at all when using touch devices?
Can it be hidden?

**To Reproduce**
I haven't tested in the noVNC UI since I'm embedding the library in my own UI.

**Expected behavior**
Only one pointer is visible at a time.

**Screenshots**

 - OS: iOS
 - Browser: safari
 - Browser version: latest

 - noVNC version: 1.1.0
 - VNC server: TigerVNC
 - WebSocket proxy: websockify

juanjoDiaz:
Fixes #1358 

I still think that a separate `connect` method to make `Rfb` instances reusable is the way to go.
But this is an improvement nevertheless.

### DIFF
--- a/core/util/cursor.js
+++ b/core/util/cursor.js
@@ -20,7 +20,6 @@ export default class Cursor {
             this._canvas.style.pointerEvents = 'none';
             // Can't use "display" because of Firefox bug #1445997
             this._canvas.style.visibility = 'hidden';
-            document.body.appendChild(this._canvas);
         }
 
         this._position = { x: 0, y: 0 };
@@ -45,6 +44,8 @@ export default class Cursor {
         this._target = target;
 
         if (useFallback) {
+            document.body.appendChild(this._canvas);
+
             // FIXME: These don't fire properly except for mouse
             ///       movement in IE. We want to also capture element
             //        movement, size changes, visibility, etc.
@@ -78,6 +79,8 @@ export default class Cursor {
             window.removeEventListener('touchstart', this._eventHandlers.touchstart, options);
             this._target.removeEventListener('touchmove', this._eventHandlers.touchmove, options);
             this._target.removeEventListener('touchend', this._eventHandlers.touchend, options);
+
+            document.body.removeChild(this._canvas);
         }
 
         this._target = null;


juanjoDiaz:
The tests seem to be failing on IE with error like `Invalid descriptor for property 'navigator'` and `Invalid descriptor for property 'localStorage'` which doesn't seem to be related to my commit. Any idea?

CendioOssman:
This fix looks correct, thanks for finding the bug. Not sure why IE is upset. I agree it looks unrelated. I'll see if I can find what has gone wrong.

CendioOssman:
Ah, there is a new version of mocha out. I can reproduce it locally here as well after an update. I'll see if I can figure out what changed and fix it.

juanjoDiaz:
Adding a `package-lock.json` to the repo would solve the issue and prevent things like this in the future. 🙂

CendioOssman:
That merely delays problems and doesn't really solve them. So we purposefully avoid such things.

Issue #1360, opened 2020-01-30
Supporting RTCDataChannel / Existing WebSocket


audiolion:
**Is your feature request related to a problem? Please describe.**
Would noVNC be willing to accept a PR  that adds support for using existing RTCDataChannels / WebSockets when setting up the vnc client?

**Describe the solution you'd like**
The RFB constructor takes a url, but with a few modifications it could instead take an existing RTCDataChannel / WebSocket connection instead as the basis for its WebSock

I forked the code base and was able to modify the source to accept the RTCDataChannel / WebSocket and work with that. If you don't want a feature that is ok, I can use my own fork, but wanted to offer to merge it back upstream 👍

CendioOssman:
That would be very welcome. I'm fairly sure it was already discussed on an existing issue.

kanaka:
@audiolion This is a great idea (especially if the change is simple) but I'm curious about your use case. Are you connecting to another browser that is acting as a server (i.e. screen sharing) or is this to connect to a VNC server that uses RTCDataChannel? If it's the latter, what is the benefit over a plain WebSocket. Perhaps it's easier to setup short running sessions because you can use STUN/TURN instead of requiring manual inbound firewall tweaks?

audiolion:
@kanaka 

> to a VNC server that uses RTCDataChannel? If it's the latter, what is the benefit over a plain WebSocket. Perhaps it's easier to setup short running sessions because you can use STUN/TURN instead of requiring manual inbound firewall tweaks?

Yes exactly, the VNC Server uses RTCDataChannel so that it can use STUN/TURN instead of manually modifying the firewall.

CendioOssman:
Fixed via #1523.

PR #1362, opened 2020-01-31
feature: Support existing RTCDataChannel or WebSocket

audiolion:
Refs #1360 

Added detail to my commits about the various technical choices I made.

If I could get some help, I have one failing test I am not really understanding it. I am having trouble following where `open` is actually called after creating a `new RFB`

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -501,7 +501,7 @@ export default class RFB extends EventTargetMixin {
         } else {
             try {
                 Log.Info(`attaching ${this._rawChannel} to WebChannel`);
-                this._webchannel.attach(this._rawChannel);
+                this._webchannel.attach(this._rawChannel, true);
             } catch (e) {
                 this._fail("Error attaching channel (" + e + ")");
             }
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -219,7 +219,7 @@ export default class RFB extends EventTargetMixin {
             }
         });
         this._sock.on('close', (e) => {
-            Log.Debug("WebChannel on-close event");
+            Log.Debug("Connection on-close event");
             let msg = "";
             if (e.code) {
                 msg = "(code: " + e.code;
@@ -252,7 +252,7 @@ export default class RFB extends EventTargetMixin {
             }
             this._sock.off('close');
         });
-        this._sock.on('error', e => Log.Warn("WebChannel on-error event"));
+        this._sock.on('error', e => Log.Warn("Connection on-error event"));
 
         // Slight delay of the actual connection so that the caller has
         // time to set up callbacks
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -15,7 +15,7 @@ import Display from "./display.js";
 import Keyboard from "./input/keyboard.js";
 import Mouse from "./input/mouse.js";
 import Cursor from "./util/cursor.js";
-import Websock from "./websock.js";
+import WebChannel from "./websock.js";
 import DES from "./des.js";
 import KeyTable from "./input/keysym.js";
 import XtScancode from "./input/xtscancodes.js";
@@ -34,18 +34,18 @@ const DISCONNECT_TIMEOUT = 3;
 const DEFAULT_BACKGROUND = 'rgb(40, 40, 40)';
 
 export default class RFB extends EventTargetMixin {
-    constructor(target, url, options) {
+    constructor(target, urlOrChannel, options) {
         if (!target) {
             throw new Error("Must specify target");
         }
-        if (!url) {
-            throw new Error("Must specify URL");
+        if (!urlOrChannel) {
+            throw new Error("Must specify URL or WebSocket / RTCDataChannel");
         }
 
         super();
 
         this._target = target;
-        this._url = url;
+        this._urlOrChannel = urlOrChannel;
 
         // Connection details
         options = options || {};
@@ -85,7 +85,7 @@ export default class RFB extends EventTargetMixin {
         this._qemuExtKeyEventSupported = false;
 
         // Internal objects
-        this._sock = null;              // Websock object
+        this._sock = null;              // WebSocket or RTCDataChannel object
         this._display = null;           // Display object
         this._flushing = false;         // Display flushing state
         this._keyboard = null;          // Keyboard input handler object
@@ -180,7 +180,7 @@ export default class RFB extends EventTargetMixin {
         this._mouse.onmousebutton = this._handleMouseButton.bind(this);
         this._mouse.onmousemove = this._handleMouseMove.bind(this);
 
-        this._sock = new Websock();
+        this._sock = new WebChannel();
         this._sock.on('message', () => {
             this._handle_message();
         });
@@ -195,7 +195,7 @@ export default class RFB extends EventTargetMixin {
             }
         });
         this._sock.on('close', (e) => {
-            Log.Debug("WebSocket on-close event");
+            Log.Debug("WebChannel on-close event");
             let msg = "";
             if (e.code) {
                 msg = "(code: " + e.code;
@@ -228,7 +228,7 @@ export default class RFB extends EventTargetMixin {
             }
             this._sock.off('close');
         });
-        this._sock.on('error', e => Log.Warn("WebSocket on-error event"));
+        this._sock.on('error', e => Log.Warn("WebChannel on-error event"));
 
         // Slight delay of the actual connection so that the caller has
         // time to set up callbacks
@@ -398,11 +398,22 @@ export default class RFB extends EventTargetMixin {
     _connect() {
         Log.Debug(">> RFB.connect");
 
-        Log.Info("connecting to " + this._url);
-
         try {
-            // WebSocket.onopen transitions to the RFB init states
-            this._sock.open(this._url, this._wsProtocols);
+            // WebChannel.onopen transitions to the RFB init states
+            if (typeof this._urlOrChannel === "string") {
+                Log.Info(`connecting to ${this._urlOrChannel}`);
+                this._sock.open({ uri: this._urlOrChannel, protocols: this._wsProtocols });
+            } else {
+                let proto = Object.getPrototypeOf(this._urlOrChannel);
+                let protoName = proto && proto.constructor && proto.constructor.name || "";
+
+                Log.Info(`connecting with established ${protoName}`);
+                if (protoName === "WebSocket" || protoName === "RTCDataChannel") {
+                    this._sock.open({ webChannel: this._urlOrChannel, channelType: protoName });
+                } else {
+                    throw new Error("Expected url and protocols or established WebSocket / RTCDataChannel object");
+                }
+            }
         } catch (e) {
             if (e.name === 'SyntaxError') {
                 this._fail("Invalid host or port (" + e + ")");
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -52,6 +52,18 @@ const extendedClipboardActionPeek    = 1 << 26;
 const extendedClipboardActionNotify  = 1 << 27;
 const extendedClipboardActionProvide = 1 << 28;
 
+// Properties a raw channel must have, WebSocket and RTCDataChannel are two examples
+const rawChannelProps = [
+    "send",
+    "close",
+    "binaryType",
+    "onclose",
+    "onerror",
+    "onmessage",
+    "onopen",
+    "protocol",
+    "readyState",
+];
 
 export default class RFB extends EventTargetMixin {
     constructor(target, urlOrChannel, options) {
@@ -59,13 +71,30 @@ export default class RFB extends EventTargetMixin {
             throw new Error("Must specify target");
         }
         if (!urlOrChannel) {
-            throw new Error("Must specify URL or WebSocket / RTCDataChannel");
+            throw new Error("Must specify URL or raw channel");
         }
 
         super();
 
         this._target = target;
-        this._urlOrChannel = urlOrChannel;
+
+        if (typeof urlOrChannel === "string") {
+            this._url = urlOrChannel;
+        } else {
+            this._url = null;
+        }
+
+        this._rawChannel = urlOrChannel;
+        for (let prop of rawChannelProps) {
+            if (!urlOrChannel[prop]) {
+                this._rawChannel = null;
+                break;
+            }
+        }
+
+        if (!this._url && !this._rawChannel) {
+            throw new Error("Must specify URL or raw channel");
+        }
 
         // Connection details
         options = options || {};
@@ -109,7 +138,7 @@ export default class RFB extends EventTargetMixin {
         this._clipboardServerCapabilitiesFormats = {};
 
         // Internal objects
-        this._sock = null;              // WebSocket or RTCDataChannel object
+        this._webchannel = null;        // WebSocket or RTCDataChannel object
         this._display = null;           // Display object
         this._flushing = false;         // Display flushing state
         this._keyboard = null;          // Keyboard input handler object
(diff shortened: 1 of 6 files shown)


CendioOssman:
Nice. I will try to find some time to do a review of this.

Feel free to have a look at the linter errors until then though. :)

audiolion:
> Feel free to have a look at the linter errors until then though. :)

Woops, my bad. Keep forgetting to run the lint fix before pushing. You should be able to see that pesky failing test case now. I am fairly certain it is a false positive because I am using the lib based off this PR and it is working.

core/rfb.js
```diff
@@ -34,18 +34,18 @@ const DISCONNECT_TIMEOUT = 3;
...
-            throw new Error("Must specify URL");
+        if (!urlOrChannel) {
+            throw new Error("Must specify URL or WebSocket / RTCDataChannel");
         }
 
         super();
 
         this._target = target;
-        this._url = url;
+        this._urlOrChannel = urlOrChannel;
```
CendioOssman: Can we avoid spreading the ambiguity beyond the constructor? I would prefer keeping `url` and setting it to  `null` when we don't know it.

core/rfb.js
```diff
@@ -85,7 +85,7 @@ export default class RFB extends EventTargetMixin {
         this._qemuExtKeyEventSupported = false;
 
         // Internal objects
-        this._sock = null;              // Websock object
+        this._sock = null;              // WebSocket or RTCDataChannel object
```
CendioOssman: Comment confusion. :)

core/rfb.js
```diff
@@ -195,7 +195,7 @@ export default class RFB extends EventTargetMixin {
             }
         });
         this._sock.on('close', (e) => {
-            Log.Debug("WebSocket on-close event");
+            Log.Debug("WebChannel on-close event");
```
CendioOssman: These are user visible and I'd like to avoid exporting our new internal "WebChannel" term. Perhaps these should have something more generic, like "Connection close event"?

core/rfb.js
```diff
@@ -398,11 +398,22 @@ export default class RFB extends EventTargetMixin {
     _connect() {
         Log.Debug(">> RFB.connect");
 
-        Log.Info("connecting to " + this._url);
-
         try {
-            // WebSocket.onopen transitions to the RFB init states
-            this._sock.open(this._url, this._wsProtocols);
+            // WebChannel.onopen transitions to the RFB init states
+            if (typeof this._urlOrChannel === "string") {
```
CendioOssman: You might want to split `RFB._connect()` so that the "url or object" check can be done already in the constructor.

core/rfb.js
```diff
@@ -398,11 +398,22 @@ export default class RFB extends EventTargetMixin {
...
+                Log.Info(`connecting to ${this._urlOrChannel}`);
+                this._sock.open({ uri: this._urlOrChannel, protocols: this._wsProtocols });
+            } else {
+                let proto = Object.getPrototypeOf(this._urlOrChannel);
+                let protoName = proto && proto.constructor && proto.constructor.name || "";
```
CendioOssman: I would prefer if we can duck type this rather than only supporting those low level objects.

core/websock.js
```diff
@@ -1,12 +1,12 @@
 /*
- * Websock: high-performance binary WebSockets
+ * WebChannel: high-performance binary WebSocket / RTCDataChannel
```
CendioOssman: Probably time to fix this comment now and just call it a wrapper. :)

core/websock.js
```diff
@@ -22,9 +22,11 @@ import * as Log from './util/logging.js';
 const ENABLE_COPYWITHIN = false;
 const MAX_RQ_GROW_SIZE = 40 * 1024 * 1024;  // 40 MiB
 
-export default class Websock {
+export default class WebChannel {
     constructor() {
-        this._websocket = null;  // WebSocket object
+        this._webChannel = null;     // WebSocket or RTCDataChannel object
```
CendioOssman: Now that we have our own name for the wrapper I think we should start clearing up the naming. So can we come up with something here that doesn't copy the name of the wrapper?

core/websock.js
```diff
@@ -173,49 +175,92 @@ export default class Websock {
...
+            // Constants pulled from RTCDataChannelState enum
+            // https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/readyState#RTCDataChannelState_enum
+            return {
+                CONNECTING: "connecting",
+                OPEN: "open",
+                CLOSING: "closing",
+                CLOSED: "closed"
+            };
+        }
+        throw new Error(`channelType: ${channelType} not recognized`);
```
CendioOssman: Can't we just check both possible values in the if clauses? I think that makes things easier to follow as you don't have to keep track of this extra mapping table.

core/websock.js
```diff
@@ -173,49 +175,92 @@ export default class Websock {
...
+
     init() {
         this._allocate_buffers();
         this._rQi = 0;
-        this._websocket = null;
+        this._webChannel = null;
     }
 
-    open(uri, protocols) {
+    open({ uri, protocols, webChannel, channelType }) {
```
CendioOssman: Can't we do "uri or object" here as well?

core/websock.js
```diff
@@ -173,49 +175,92 @@ export default class Websock {
...
             this._eventHandlers.open();
-            Log.Debug("<< WebSock.onopen");
+            Log.Debug(`<< ${this._channelType}.onopen`);
         };
-        this._websocket.onclose = (e) => {
-            Log.Debug(">> WebSock.onclose");
+
+        this._webChannel.binaryType = "arraybuffer";
+
+        this._webChannel.onmessage = this._recv_message.bind(this);
```
CendioOssman: Since we're now given an existing object I think it's time we start being more well behaved and use `addEventListener()` instead.

core/websock.js
```diff
@@ -173,49 +175,92 @@ export default class Websock {
...
+            );
+        }
 
-        this._websocket.onmessage = this._recv_message.bind(this);
-        this._websocket.onopen = () => {
-            Log.Debug('>> WebSock.onopen');
-            if (this._websocket.protocol) {
-                Log.Info("Server choose sub-protocol: " + this._websocket.protocol);
+        const onOpen = () => {
+            Log.Debug(`>> ${this._channelType}.onopen`);
```
CendioOssman: The text here referred to the class name, so it doesn't need to be changed.

core/websock.js
audiolion: `WebChannel: high-performance wrapper over a WebSocket or RTCDataChannel`?

core/websock.js
audiolion: any ideas? naming things is hard >:)

core/websock.js
audiolion: from your earlier comments, if I isolate whether it is a url or existing websocket/datachannel in the constructor of RFB, accepting `open(uriOrWebChannel, protocols)` here would be polluting again and require similar checks to determine if it is a string, or an object, if it is an object, is it a rtcdatachannel/websocket. I can make this change if you want, but this API is internal so we can break it without worry.

core/websock.js
CendioOssman: That or perhaps just "wrapper" or "buffering wrapper". The longer description further down seems to be more accurate.

core/websock.js
CendioOssman: JavaScript fortunately has a tendency for long names, so we don't have to be too afraid of that. Perhaps `_underlyingSocket`, `_rawChannel` or something along those lines?

core/websock.js
CendioOssman: So far we've been trying to keep this somewhat independent (we are supposed to share it with the websockify repo).

Since it isn't a constructor we could split it to multiple functions for clarity. E.g. an `open()` that takes an URL and a `attach()` that takes an already open object?

audiolion: Sure that works for me.

vtolstov:
very cool, thanks for this pr.
@audiolion @CendioOssman  when this can be merged? I have own use case for this (mostly i have some server that check user right via oauth2 token in websocket security). And if token fine - connects to vnc server (server does not need auth)

audiolion:
@vtolstov I have to address the feedback, I will take a look at this today.

vtolstov:
Hi, do you have time to fix last things?

audiolion:
@vtolstov ill add you to my fork repo so you can make commits as well. I will be working on this but it could be a few days before I have free time again.

vtolstov:
Thank you!

core/websock.js
audiolion: this is the one feedback I didn't do, wasn't sure exactly what you meant.

audiolion:
@CendioOssman I believe all feedback has been addressed except for one item, left a comment on it. All tests passing now, I hope I followed what you had in mind for how it would look (duck typing, naming, open/attach api).

core/websock.js
samhed: instead of using `onmessage` you should be able to do `addEventListener('message', this._recv_message.bind(this));`

audiolion: Both WebSocket and RTCDataChannel expose `onmessage` to attach a function handler to be called, shouldn't we stick to using the provided methods?

samhed: From what I've heard it seems the `onevent` functions are becoming deprecated in favor of `addEventListener()`. I haven't looked into this one in detail, but it seems the most future proof path.

If we would make this change we should remove those 4 functions from `rawChannelProps` as well.

audiolion: Such a refactor would touch multiple unrelated areas of the codebase and seems outside the scope of this PR, could we make an issue to track this instead?

CendioOssman: We don't need to go over the entire codebase right now, but I think it needs to be done for the WebSocket/RTCDataChannel object on this PR. The reason being that we no longer have exclusive control of this object. The caller might also want to set up event handlers, so we should use the mechanism that allows multiple handlers.

CendioOssman:
Sauce is acting up in general the last few days. I'll see if I can have a look at it and figure out what's going on.

CendioOssman:
Not sure why travis isn't printing the errors properly, but the issue is that IE doesn't support `includes`:

```
  WebChannel
    Queue methods
      flush
        ✖ should actually send on the websocket
          IE 11.0.0 (Windows 7.0.0)
        TypeError: Object doesn't support property or method 'includes'
           at flush (eval code:178:13)
           at Anonymous function (eval code:189:15)
```

(There's a whole bunch of these)

audiolion:
> Not sure why travis isn't printing the errors properly, but the issue is that IE doesn't support includes:



I didn't realize novnc supported IE. I will change to `indexOf`

core/websock.js
audiolion: I am not quite sure how to update the FakeWebSocket class to have an `addEventListener` and `dispatchEvent` method to test against.

audiolion:
Though TravisCI is not reporting it for some reason, the latest build from ae6db43 is passing https://travis-ci.org/github/novnc/noVNC/builds/672578102. 

The final TODO is refactoring to `addEventListener` for `open, close, message, error` events. I could use some help in doing this though, moving to `addEventListener` breaks many tests that mock out `onmessage, onopen, onclose, onerror`, either through the FakeWebSocket class, or through sinon spys. I took a stab at it, we have to use the old event dispatch pattern so that it works in PhantomJS and IE11:

```js
let evt = document.createEvent('Event');
evt.initEvent('open', true, true);

// if we want any data on the evt obj
for (let prop in props) {
  evt[prop] = props[prop]
}
```

Mocks like FakeWebSocket require `addEventListener` and `dispatchEvent` methods:

```js
    constructor() {
       // ..
      this.__listeners = []
    }

    addEventListener(type, listener, options) {
        this.__listeners.push({type, handler: listener, options});
    }

    dispatchEvent(evt) {
        for (let i = 0; i < this.__listeners.length; i++) {
            let listener = this.__listeners[i];
            if (evt.type === listener.type) {
                listener.handler(evt);
            }
        }
    }
```

This gets me closer but I am still left with a number of failing cases that I can't seem to fix. I have pushed  my work up for now, currently only working on the `close` event. I figured I would do them one at a time to reduce the number of failing test cases.

If anyone can give me some guidance or direction here it would be much appreciated! 😃

CendioOssman:
It should just be a matter of:

```diff
diff --git a/tests/fake.websocket.js b/tests/fake.websocket.js
index 5194685..12e646d 100644
--- a/tests/fake.websocket.js
+++ b/tests/fake.websocket.js
@@ -35,9 +35,7 @@ export default class FakeWebSocket {
 
     close(code, reason) {
         this.readyState = FakeWebSocket.CLOSED;
-        if (this.onclose) {
-            this.onclose(make_event("close", { 'code': code, 'reason': reason, 'wasClean': true }));
-        }
+        this.dispatchEvent(make_event("close", { 'code': code, 'reason': reason, 'wasClean': true }));
     }
 
     send(data) {
@@ -71,9 +69,7 @@ export default class FakeWebSocket {
 
     _open() {
         this.readyState = FakeWebSocket.OPEN;
-        if (this.onopen) {
-            this.onopen(make_event('open'));
-        }
+        this.dispatchEvent(make_event('open'));
     }
 
     _receive_data(data) {
@@ -81,7 +77,7 @@ export default class FakeWebSocket {
         // neatly packaged
         for (let i = 0;i < data.length;i++) {
             let buf = data.subarray(i, i+1);
-            this.onmessage(make_event("message", { 'data': buf }));
+            this.dispatchEvent(make_event("message", { 'data': buf }));
         }
     }
 }
```

(it seems Travis has gotten utterly confused by this PR, always showing the latest entry as ongoing)

CendioOssman:
Also note that some test are in need of cleanup. They should not be poking about in the `WebChannel`/`WebSock` object but should manipulate the underlying `FakeWebSocket`.

audiolion:
> It should just be a matter of:

Thanks for the help, I will give that a shot

> Also note that some test are in need of cleanup.

Ok I will look out for this.

samhed:
@audiolion what's the status on this?

audiolion:
hey I just got to a point where I have time to circle back to this again, early next week I should be able to work on it again. The PR is working currently, but I was asked to refactor some of the event listener code as well so that is where we are at now. Also for some reason travis never updates the build status for this Github PR, so don't pay attention to that.

samhed:
Sounds good, thanks for the update!

ppoffice:
Great work! Just tried it using WebRTC DataChannel and it works great with STUN. However, there is one thing I added to make it work:

```diff
diff --git a/core/rfb.js b/core/rfb.js
index f6553cd..84ed8c8 100644
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -502,6 +502,7 @@ export default class RFB extends EventTargetMixin {
         } else {
             try {
                 Log.Info(`attaching ${this._rawChannel} to WebChannel`);
+                this._webchannel.init();
                 this._webchannel.attach(this._rawChannel, true);
             } catch (e) {
                 this._fail("Error attaching channel (" + e + ")");
```

Otherwise the buffers will be null and RFB will just crash.

audiolion:
@ppoffice thanks for that, I missed that when refactoring.

samhed:
ping @audiolion

audiolion:
Hey we ended up not using novnc at work so I am not able to continue work on this. As ststed before this was all working besides a refactor that I think is a bit out of scope for this CL. If anyone wants to take this up they can.

vtolstov:
@audiolion what you use instead of novnc? 
@CendioOssman can we merge this as-is and create issue for feature changes to add some sugar and beautify it?

CendioOssman:
We'll see if one of the core team can finish it up, given that they can modify PRs.

audiolion:
Grateful for the help. I normally would love to finish this up but I am so stretched thin right now. Appreciate your understanding. I think whoever takes this over will have a better vision for how to fit it into the existing codebase as well 👍🏻

jcpunk:
I'll confess this is something I'd love to see in novnc.

leonardospeca:
Waiting for updates, it will be a really cool update, specially for the situation where you want to use noVNC over https to control a device in the local network (http) due to this limitation of websockets

CendioOssman:
#1523 has now been merged instead. If there any other changes from this PR you'd like to get included then please submit a new PR with those changes.

Issue #1364, opened 2020-02-06
Detect parent element resize to redraw canvas (instead of window.onresize)


baleeds:
**Is your feature request related to a problem? Please describe.**
Currently, the canvas element resizes perfectly to its parent container when it detects a `window.onresize`.  However, when the size of the container changes without a window change, like a user resizable `div`, the canvas fails to redraw.

[See this youtube video](https://youtu.be/k9yhnoncuJ8)

**Describe the solution you'd like**
Could we consider switching the implementation of using `window.onresize` to instead use a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) polyfill?  This would work when the container size changes, whether it's from a window event or not.

**Describe alternatives you've considered**
Another option is to expose a function that will allow the consumer to indicate that the canvas should be redrawn.  Something like `rfb.redrawCanvas()`?

CendioOssman:
The problem with that is that neither Edge nor IE supports that. So it would have to be changed to somehow support both.

Should be doable, but might require some extra work.

baleeds:
[This polyfill](https://github.com/que-etc/resize-observer-polyfill) is supported back to IE8.  I've used it successfully in the past.

CendioOssman:
Nice. Worth having a look at then I guess.

Out of curiosity, what is you are integrating noVNC in?

baleeds:
My company is creating a remote monitoring and management platform.  https://level.tech/

I see you have a `FixMe` in there.  
![image](https://user-images.githubusercontent.com/9682964/74038003-6b70ea80-498d-11ea-807e-52641825e7d2.png)

If you'd like, I can open a PR for it.

CendioOssman:
A PR would be very welcome. So go right ahead.

baleeds:
https://github.com/novnc/noVNC/pull/1365

PR #1365, opened 2020-02-07
feature: Detect parent resize

baleeds:
Fixes an issue where if the screen div resizes for a reason other than window resize, the canvas wouldn't redraw.

https://github.com/novnc/noVNC/issues/1364

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ noVNC uses many modern web technologies so a formal requirement list is
 not available. However these are the minimum versions we are currently
 aware of:
 
-* Chrome 49, Firefox 44, Safari 11, Opera 36, Edge 79
+* Chrome 64, Firefox 79, Safari 13.4, Opera 51, Edge 79
 
 
 ### Server Requirements
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -134,6 +134,7 @@ export default class RFB extends EventTargetMixin {
         this._flushing = false;         // Display flushing state
         this._keyboard = null;          // Keyboard input handler object
         this._gestures = null;          // Gesture input handler object
+        this._resizeObserver = null;    // Resize observer object
 
         // Timers
         this._disconnTimer = null;      // disconnection timer
@@ -171,7 +172,7 @@ export default class RFB extends EventTargetMixin {
         // Bound event handlers
         this._eventHandlers = {
             focusCanvas: this._focusCanvas.bind(this),
-            windowResize: this._windowResize.bind(this),
+            handleResize: this._handleResize.bind(this),
             handleMouse: this._handleMouse.bind(this),
             handleWheel: this._handleWheel.bind(this),
             handleGesture: this._handleGesture.bind(this),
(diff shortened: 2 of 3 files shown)


baleeds:
It appears that TravisCI doesn't install dependencies before running tests?

CendioOssman:
Yeah, we don't want any external dependencies for the project. So please put a copy of what you need in the `vendor/` directory. We also want things to be ES6 modules, so some tweaking might be required.

baleeds:
> Could you run 'npm run lint' and fix the whitespace issues?
This didn't seem to do anything for me.  Is there a guide on this project's linting preferences?  That may be helpful, in the absence of a tool like Prettier.

LICENSE.txt
```diff
@@ -48,6 +48,8 @@ licenses (all MPL 2.0 compatible):
 
     vendor/promise.js       : MIT
 
+    vendor/resize-observer-polyfill : MIT
+
```
samhed: You should probably also include the path in the section below "The following license texts are included".

CendioOssman:
This looks like it is pretty much ready to be merged. What browsers have been tested?

baleeds:
> This looks like it is pretty much ready to be merged. What browsers have been tested?

I've tested in Chrome, Firefox, and Safari.  I haven't looked at IE or Edge, but according to the [polyfill docs](https://github.com/juggle/resize-observer#desktop), IE 11 is supported, along with all versions of Edge.

samhed:
I'll help out with some testing on IE and Edge tomorrow.

baleeds:
Hi @samhed, did you get a chance to test on IE?

samhed:
Shit.. sorry @baleeds, I totally forgot about this.

When I checkout your branch I can't even get things going however. First I get:

> core/rfb.js:23   /vendor/resize-observer-polyfill/dist/ResizeObserver net::ERR_ABORTED 404 (File not found)

If I add ".js" to the filename in the import I get the following error:

> SyntaxError: The requested module '../vendor/resize-observer-polyfill/dist/ResizeObserver.js' does not provide an export named 'default'

samhed:
Ping @baleeds

baleeds:
Hi sorry!  I don't know how to do it differently.  Is there a process that I should be taking for importing the library as an ES6 module?

CendioOssman:
Your import right now seems to be geared towards NodeJS. Are you testing this code directly in a browser without any modifiction? I.e. no browserify, babel or anything like it.

First you need to have an ES6 version of the code, which seems to already exist in the form of `ResizeObserver.es.js `. Then you need to import by filename. Lastly, let's not ship things we don't need to clean out any files that aren't used.

samhed:
Do you think you will get a chance to fix this @baleeds?

baleeds:
> Do you think you will get a chance to fix this @baleeds?

Hi Samhed, sorry I've been non-responsive.  Our project went in a different direction in terms of technology.  I do think this change is worthwhile, but I don't have time to look at it now.  It was only a few lines of code, and then adding in the library, if you wanted to make it happen.

CendioOssman:
We've dropped support for Internet Explorer so this should be possible to merge without the polyfill now.

samhed:
I have removed the polyfill, removed unrelated changes and rebased on master. I have manually tested on:

**Windows 10**
- [x] Edge 89
- [x] Chrome 89
- [x] Firefox 87

**macOS 11**
- [x] Safari 14
- [x] Chrome 89
- [x] Firefox 86

**iOS 14.4**
- [x] Safari

**Android 11**
- [x] Chrome 89
- [x] Firefox 87

What remains is to modify the unit tests, they rely on the `resize` event at the moment.

samhed:
Unit tests have now been added and the change is merged.

Issue #1288, opened 2019-09-03
How to configure image quality settings


samgoud:
**Describe the bug**
We are using tigervnc 1.9 to launch vnc server with -xstartup to launch a qt gui app
We were comparing the display quality of this app when viewed on novnc vs when the app was directly launched on the remote desktop.
What we noticed is in novnc the image quality is a little distorted in comparison
Below is a comparison:  
![image](https://user-images.githubusercontent.com/32926439/64209077-35e74300-cebd-11e9-9e8f-c91363b564b9.png)

Are there any settings in novnc we need to configure to make the picture better.

samhed:
No, I'm afraid there is no such setting at the moment, see #396 for adding this feature.

Until then you will have to manually modify the code for QualityLevel and CompressLevel here:
https://github.com/novnc/noVNC/blob/master/core/rfb.js#L1230

samhed:
I'd guess what you'd want is to increase the quality level and decrease the compress level. The code om noVNC uses quality level 6 and compress level 2, try increasing quality to 9 (which is max) and turnn of compression by changing the number to 0.

samgoud:
Thanks samhed, 
I will try the suggested change

samhed:
You'd be using more CPU on the client and more bandwidth over the network. Not sure about the memory impact, but I'd guess it wouldn't change much.

samgoud:
We are using a old version of novnc. Not sure which version exactly it is. From package.json I see
  "name": "noVNC",
  "version": "0.6.1",

In the version we have, I could not see the exact variables as suggested by you in rfb.js 

I found similar variables in our rfb.js with same encode values as in the latest noVnc
![image](https://user-images.githubusercontent.com/32926439/65240929-00b73380-db00-11e9-974b-6642278a7289.png)

I modified the settings to below
![image](https://user-images.githubusercontent.com/32926439/65241015-30fed200-db00-11e9-8a63-8a2ca936ce90.png)

image quality looks better now.

Does the modification looks ok ? I have attached the rfb.js file for your reference
[rfb.zip](https://github.com/novnc/noVNC/files/3631006/rfb.zip)

samhed:
The modifications in your post look good, you can see the details in the RFB specification here:

https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#encodings

I didn't look at your zip

CendioOssman:
[...]

Issue #1150, opened 2018-10-05
[Feature request]compression level and quality options


al3x609:
It would be good to allow users to modify the compression level from noVNC, to adapt it to the characteristics of their connection.
not only default Compression Level 2 and Quality Level 6, maybe...
https://github.com/novnc/noVNC/blob/d1050405813412a684636522f8daddb5ad913fe2/core/rfb.js#L1242-L1243

CendioOssman:
Please see #396 for tracking this feature request.

PR #1373, opened 2020-02-23
Add possibility to set quality ("JPEG Quality Level Pseudo-encoding")

### LINKED ISSUE #396: Change encoding and pixel format on the fly

trebler:
This PR adds possibility to change the quality (aka [JPEG Quality Level Pseudo-encoding](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#jpeg-quality-level-pseudo-encoding)) on the fly. Previously, the value was hardcoded to be `6` (more precisely: `-32 + 6 = -26`).

`RFB` class got a new property `qualityLevel` which can be set immediately after the `RFB` constructor (and then encodings are sent once as part of `ServerInitialisation` state). But it can also be set in `connected` state (and then encodings are sent again).

Values passed to `qualityLevel` setter are strictly verified to be integers in range `[0, 9]`. Hence, for IE11, I have added `Number.isInteger()` polyfill.

This has only been tested against (latest) TigerVNC.

Partially resolves #396, #1150, #1288 and similar.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -64,6 +64,11 @@ protocol stream.
     to the element containing the remote session screen. The default value is `rgb(40, 40, 40)`
     (solid gray color).
 
+`qualityLevel`
+  - Is an `int` in range `[0-9]` controlling the desired JPEG quality.
+    Value `0` implies low quality and `9` implies high quality.
+    Default value is `6`.
+
 `capabilities` *Read only*
   - Is an `Object` indicating which optional extensions are available
     on the server. Some methods may only be called if the corresponding
--- a/tests/test.rfb.js
+++ b/tests/test.rfb.js
@@ -2865,6 +2865,108 @@ describe('Remote Frame Buffer Protocol Client', function () {
             // error events do nothing
         });
     });
+
+    describe('Quality level setting', function () {
+        const defaultQuality = 6;
+
+        let client;
+
+        beforeEach(function () {
+            client = make_rfb();
+            sinon.spy(RFB.messages, "clientEncodings");
+        });
+
+        afterEach(function () {
+            RFB.messages.clientEncodings.restore();
+        });
+
+        it(`should equal ${defaultQuality} by default`, function () {
+            expect(client._qualityLevel).to.equal(defaultQuality);
+        });
+
+        it('should ignore non-integers when set', function () {
+            client.qualityLevel = '1';
+            expect(RFB.messages.clientEncodings).to.not.have.been.called;
+
+            RFB.messages.clientEncodings.resetHistory();
+
+            client.qualityLevel = 1.5;
+            expect(RFB.messages.clientEncodings).to.not.have.been.called;
+
+            RFB.messages.clientEncodings.resetHistory();
+
+            client.qualityLevel = null;
+            expect(RFB.messages.clientEncodings).to.not.have.been.called;
+
+            RFB.messages.clientEncodings.resetHistory();
+
+            client.qualityLevel = undefined;
+            expect(RFB.messages.clientEncodings).to.not.have.been.called;
+
+            RFB.messages.clientEncodings.resetHistory();
+
+            client.qualityLevel = {};
+            expect(RFB.messages.clientEncodings).to.not.have.been.called;
+        });
+
+        it('should ignore integers out of range [0, 9]', function () {
+            client.qualityLevel = -1;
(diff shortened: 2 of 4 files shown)


CendioOssman:
Awesome, thanks! We'll try to get this reviewed this week.

I am curious about the testing though. TigerVNC has its lossless refresh feature, meaning changes to the quality level aren't that visible. What was the scenario where you needed this?

trebler:
We use noVNC/TigerVNC to "stream" a 3D application with a dynamic scene. So, it is very obvious (both visually and also when comparing amount of data received per sec) that varying quality with `0` (`-32`) and `9` (`-23`) makes a big difference. With `0` one notices a lot of compression artefacts, not so much with values `>= 5` (Of course, this lossless feature of TigerVNC guarantees that static parts of a view, e.g. GUI controls are always rendered with high quality.)

So, for your test I could suggest playing a video over noVNC :)

Our client using satellite Internet has even reported big improvements between using setting `4` and default `6` in terms of latency.

Locally, I have used Apple's `Network Link Conditioner` utility to simulate poor internet connection.

On a side note, we have implemented a feature allowing for automatic adaptive quality setting using `Network Information API`, but this is for Chromium-based browsers only, and hence probably cannot make its way into noVNC.

samhed:
Very nice! Good job on the unit tests as well!

docs/API.md
```diff
@@ -64,6 +64,13 @@ protocol stream.
     to the element containing the remote session screen. The default value is `rgb(40, 40, 40)`
     (solid gray color).
 
+`qualityLevel`
+  - Is an `int` in range `[0-9]` controlling [`JPEG Quality Level Pseudo-encoding`](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#jpeg-quality-level-pseudo-encoding). Value `0` is equivalent to
```
CendioOssman: Can we describe this more in terms a user might understand? They shouldn't have to know the details of the protocol.

tests/test.rfb.js
```diff
@@ -2865,6 +2865,126 @@ describe('Remote Frame Buffer Protocol Client', function () {
...
+        it(`should equal ${defaultQuality} by default`, function () {
+            expect(client._qualityLevel).to.equal(defaultQuality);
+        });
+
+        it('should ignore non-integers when set', function () {
+            let newQuality;
+
+            newQuality = 3;
+            client.qualityLevel = newQuality;
+            expect(client._qualityLevel).to.equal(newQuality);
```
CendioOssman: Examining internal variables generally results in fragile tests, so we try to avoid that. Do you think you could examine the arguments to `RFB.messages.clientEncodings()` instead?

tests/test.rfb.js
```diff
@@ -2865,6 +2865,126 @@ describe('Remote Frame Buffer Protocol Client', function () {
...
+            expect(RFB.messages.clientEncodings).to.have.been.calledOnce;
+
+            client.qualityLevel = '1';
+            expect(client._qualityLevel).to.equal(newQuality);
+            expect(RFB.messages.clientEncodings).to.have.been.calledOnce;
```
CendioOssman: You can reset the spy between calls, which can make this more readable.

docs/API.md
trebler: Would this be too much of a simplification?

`qualityLevel`
  - Is an `int` in range `[0-9]` controlling the desired JPEG quality. Value `0` implies low quality and `9` implies high quality.
    Default value is `6`.

CendioOssman: That sounds like a reasonable level to me.

tests/test.rfb.js
trebler: Fixed

tests/test.rfb.js
trebler: Fixed

docs/API.md
trebler: Ok, fixed then.

trebler:
@CendioOssman, pushed updates based on your suggestions.

trebler:
I have added a line break in the documentation, and squashed commits.

PR #1374, opened 2020-02-24
Hide clipboard side bar button when view only mode

### LINKED ISSUE #1368: The clipboard button in the sidebar is visible when using view only mode
When using "View Only" the clipboard option in the sidebar should be hidden for clarity due to the clipboard not having any functionality at that time.

1. Turn on 'View Only' in the sidebar.
2. Note that the clipboard button does not disappear.

CendioNiko:
The clipboard side bar button serves no purpose if user uses 'View Only' mode. This commit hides this button in those instances.

Tested on Fedora 31 in Firefox 73 and Chrome 80.

Fixes #1368


CendioOssman:
Ehm? I merged this, yet this PR isn't updating?

PR #1375, opened 2020-02-25
Make logger output configurable

jalfd:
init_logging can now optionally be called with an object containing
custom handlers for log messages.

This allows clients to use their own logging frameworks instead of log
messages always just being written to the console.

### DIFF
--- a/core/util/logging.js
+++ b/core/util/logging.js
@@ -17,18 +17,28 @@ let Info = () => {};
 let Warn = () => {};
 let Error = () => {};
 
-export function init_logging(level) {
-    if (typeof level === 'undefined') {
-        level = _log_level;
-    } else {
-        _log_level = level;
+export function init_logging(param) {
+    Debug = Info = Warn = Error = () => {};
+    let has_custom_logger = false;
+    switch (typeof param) {
+        case 'string':
+            _log_level = param;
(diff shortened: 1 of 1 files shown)


samhed:
We have previously discussed this in #243 and #1206, and at a glance, this doesn't seem to suffer from the same problems that caused #1206 to be rejected. @CendioOssman, thoughts?

CendioOssman:
It's definitely along the right path. It doesn't seem to properly support full objects though, so some tweaks are needed.

Basically it should be possible to call it with `init_logging(window.console)`. And level should be kept as a separate thing.

jalfd:
> Basically it should be possible to call it with init_logging(window.console).

Sounds reasonable

> And level should be kept as a separate thing

My thinking was that if you install custom handlers for your log messages, they would do the level filtering, so noVNC would not need to track a separate level. However, I can definitely leave the level parameter in there as well if you prefer. :)

CendioOssman:
Since anyone using this still has to call `init_logging()` themselves they already have control over the logging. And if they want to handle it in their wrapper then they can just specify full logging to get all the calls.

samhed:
ping @jalfd

samhed:
No response.

PR #1376, opened 2020-02-25
Add scaleFactor option to RFB options

jalfd:
This allows the viewport to be scaled to compensate for DPI scaling on
high-DPI monitors. Previously, if DPI scaling was set to 200% in Windows, for example,
noVNC would request a viewport half the size of the actual screen area,
and scale it up. Leading to the VNC session looking grainy and low-res.

The caller can now set scaleFactor to 1 / window.devicePixelRatio to
compensate for this, and get a canvas scaled to the actual physical
screen DPI.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -337,6 +337,16 @@ export default class RFB extends EventTargetMixin {
         }
     }
 
+    get useNativeDpi() { return this._nativeDpi; }
+    set useNativeDpi(nativeDpi) {
+        this._nativeDpi = nativeDpi;
+        this._updateScale();
+        this._updateClip();
+        if (this._resizeSession) {
+            this._requestRemoteResize();
+        }
+    }
+
     get showDotCursor() { return this._showDotCursor; }
     set showDotCursor(show) {
         this._showDotCursor = show;
@@ -604,7 +614,14 @@ export default class RFB extends EventTargetMixin {
 
     _updateScale() {
         if (!this._scaleViewport) {
-            this._display.scale = 1.0;
+            if (this._nativeDpi) {
+                const dpr = window.devicePixelRatio || 1;
+                const scaledWidth = Math.floor(window.innerWidth * dpr);
+                // Compute the exact scale factor to match the floored width/height
+                this._display.scale = window.innerWidth / scaledWidth;
+            } else {
+                this._display.scale = 1;
+            }
         } else {
             const size = this._screenSize();
             this._display.autoscale(size.w, size.h);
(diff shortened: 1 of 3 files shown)


samhed:
What are you using on the VNC serverside to achieve good up-scaling? In my experience, support for this is lacking in Linux.

If thinking about merging this I'd like to see it used in the UI settings of `vnc.html` as well.

CendioOssman:
If the use case is HiDPI, then a simple boolean of "native" resolution or not should be sufficient. Let's try to keep things simple.

samhed:
ping @jalfd

samhed:
No response

samhed:
Go ahead and update this one 👍

jalfd:
> Go ahead and update this one 👍

Done!

core/rfb.js
```diff
@@ -604,7 +614,14 @@ export default class RFB extends EventTargetMixin {
 
     _updateScale() {
         if (!this._scaleViewport) {
-            this._display.scale = 1.0;
+            if (this._nativeDpi) {
+                const dpr = window.devicePixelRatio || 1;
+                const scaledWidth = Math.floor(window.innerWidth * dpr);
+                // Compute the exact scale factor to match the floored width/height
+                this._display.scale = window.innerWidth / scaledWidth;
```
CendioOssman, line 621: I don't understand the purpose of this calculation. Why not just `1 / window.devicePixelRatio`?

core/rfb.js
```diff
@@ -624,8 +641,10 @@ export default class RFB extends EventTargetMixin {
         }
 
         const size = this._screenSize();
+        size.w = Math.floor(size.w / this._display.scale);
+        size.h = Math.floor(size.h / this._display.scale);
```
CendioOssman, line 645: I'm not sure we want to tie this in to the display scaling right now. I would prefer if you just use `window.devicePixelRatio` and `this._nativeDpi` here.

core/rfb.js
```diff
@@ -337,6 +337,16 @@ export default class RFB extends EventTargetMixin {
         }
     }
 
+    get useNativeDpi() { return this._nativeDpi; }
+    set useNativeDpi(nativeDpi) {
```
CendioOssman, line 341: Should this be `useNativeDPI`? What does JavaScript APIs usually do with abbreviations?

samhed:
ping @jalfd

jalfd:
Sorry, I'm completely swamped with work at the moment, so it'll be a little while before I get back to this. I'll update the PR once I'm able to make the requested changes!

CendioOssman:
Any luck getting the final parts fixed up? :)

alectrocute:
@samhed Maybe we should re-add the “noresponse” label?

CendioOssman:
We try to close them now instead. The label doesn't really help out in getting some action on an issue. :/

PR #1384, opened 2020-03-23
rfb: VeNCrypt Plain SecurityType support

regnarg:
This allows using TigerVNC server with PAM authentication (e.g. against
LDAP or other extensible authentication mechanisms)

Tested with TigerVNC server (Xvnc -SecurityTypes Plain -PlainUsers '*')

Should not break anything else, this method is tried last when all
other fail.

Tested in Firefox 74 and Chromium 80

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -84,6 +84,7 @@ export default class RFB extends EventTargetMixin {
         this._rfb_version = 0;
         this._rfb_max_version = 3.8;
         this._rfb_tightvnc = false;
+        this._rfb_vencrypt_state = 0;
         this._rfb_xvp_ver = 0;
 
         this._fb_width = 0;
@@ -943,6 +944,8 @@ export default class RFB extends EventTargetMixin {
                 this._rfb_auth_scheme = 16; // Tight
             } else if (includes(2, types)) {
                 this._rfb_auth_scheme = 2; // VNC Auth
+            } else if (includes(19, types)) {
+                this._rfb_auth_scheme = 19; // VeNCrypt Auth
             } else {
                 return this._fail("Unsupported security types (types: " + types + ")");
             }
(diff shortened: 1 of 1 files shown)


regnarg:
NB: I have no idea why CI check fails, does not seem related to my changes.

regnarg:
I tried to improve the formatting and add blank lines for clarity.

core/rfb.js
```diff
@@ -1018,6 +1021,93 @@ export default class RFB extends EventTargetMixin {
...
+    // protocol: https://github.com/rfbproto/rfbproto/blob/617d72b4ced0da7d90d4a70cc6d6e1a30819f597/rfbproto.rst#vencrypt
+    _negotiate_vencrypt_auth() {
+
+        if (!this._rfb_credentials.username ||
+            !this._rfb_credentials.password) {
+            this.dispatchEvent(new CustomEvent(
+                "credentialsrequired",
+                { detail: { types: ["username", "password", "target"] } }));
+            return false;
+        }
```
CendioOssman: You don't know that you'll need this yet. The server might not support Plain.

core/rfb.js
```diff
@@ -1018,6 +1021,93 @@ export default class RFB extends EventTargetMixin {
...
+                // 0x100 = 256
+                this._sock.send([0, 0, 1, 0]);
+                const user = this._rfb_credentials.username;
+                const pass = this._rfb_credentials.password;
+
+                // XXX we assume lengths are <= 255 (should not be an issue in the real world)
+                this._sock.send([0, 0, 0, user.length]);
+                this._sock.send([0, 0, 0, pass.length]);
+                this._sock.send_string(user);
+                this._sock.send_string(pass);
```
CendioOssman: These two need some encoding here. Most likely UTF-8.

core/rfb.js
```diff
@@ -1018,6 +1021,93 @@ export default class RFB extends EventTargetMixin {
...
+        }
+
+        // waiting for subtypes list
+        if (this._rfb_vencrypt_state == 3) {
+            if (this._sock.rQwait("vencrypt subtypes", 4 * this._rfb_vencrypt_subtypes_length)) { return false; }
+
+            const subtypes = [];
+            for (let i = 0; i < this._rfb_vencrypt_subtypes_length; i++) {
+                subtypes.push(this._sock.rQshift32());
+            }
```
CendioOssman: Can we restructure things so this is shared with the normal handling? VeNCrypt is a super set of the normal authentication after all.

core/rfb.js
```diff
@@ -1018,6 +1021,93 @@ export default class RFB extends EventTargetMixin {
         return this._negotiate_authentication();
     }
 
+    // VeNCrypt authentication, currently only supports version 0.2 and only Plain subtype
+    // protocol: https://github.com/rfbproto/rfbproto/blob/617d72b4ced0da7d90d4a70cc6d6e1a30819f597/rfbproto.rst#vencrypt
```
CendioOssman: We don't need this as all of this file uses that specification as the reference.

regnarg:
> Out of curiosity, what does your setup look like?
I use TigerVNC with Plain auth against an LDAP server using pam_ldap.

> I figured Plain was rarely used in TigerVNC as it requires running the server as root.
Only when authenticating against local user database (PAM needs root to access /etc/shadow). With LDAP you can do it without root, pam_ldap just verifies the credentials with LDAP server.

core/rfb.js
regnarg: I am not sure how one would do that. VeNCrypt is one security type, which has subtypes. You first negotiate the VeNCcrypt security type and then negotiate the subtype. The subtype negotiation protocol is different from the type negotiation protocol. For example, subtype numbers are 32 bit but type numbers are 8 bit.

samhed:
@regnarg do you think you will get a chance to make the requested changes?

regnarg:
> @regnarg do you think you will get a chance to make the requested changes?

Yes, I will update it soon.

regnarg:
Updated according to requested changes.

regnarg:
On Tue, Apr 07, 2020 at 11:55:33PM -0700, Pierre Ossman (Work account) wrote:
> LGTM too. Could you squash the commits in to one?

Squashed.

PR #1390, opened 2020-04-06
fix of error - 'code 400, message Client must support 'binary' or 'base64' protocol

vitaly-zverev:
In accordance with rfc6455 (page 5) https://tools.ietf.org/html/rfc6455  :
>>> The handshake from the client looks as follows:
bla bla bla
>>>Sec-WebSocket-Protocol: chat, superchat
we need Sec-WebSocket-Protocol: binary

Without fix we could get 'code 400', for example:
websockify --web=. --cert=/etc/ssl/novnc.pem 8888 localhost:5900

WebSocket server settings:
  - Listen on :8888
  - Flash security policy server
  - Web server. Web root: /home/user/noVNC
  - SSL/TLS support
  - proxying from :8888 to localhost:5900
Home.local - - [06/Apr/2020 13:43:24] code 400, message Client must support 'binary' or 'base64' protocol
Home.local - - [06/Apr/2020 13:43:24] code 404, message File not found

and we can see regular handshake in the right place, just after fix:
user@ubuntu:~/noVNC$ sudo strace -f websockify --web=. --cert=/etc/ssl/novnc.pem 8888 localhost:5900  2>&1 | grep Sec-WebSocket-Protocol
[pid 13746] sendto(4, "Sec-WebSocket-Protocol: binary\r\n", 32, 0, NULL, 0) = 32

PS
tested with clients:
FireFox 74.0.1 (32-bit), Chrome 80.0.3987.149 (32-bit)

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -72,7 +72,7 @@ export default class RFB extends EventTargetMixin {
         this._rfb_credentials = options.credentials || {};
         this._shared = 'shared' in options ? !!options.shared : true;
         this._repeaterID = options.repeaterID || '';
-        this._wsProtocols = options.wsProtocols || [];
+        this._wsProtocols = options.wsProtocols || ["binary"];
 
         // Internal state
         this._rfb_connection_state = '';


vitaly-zverev:
Perhaps it should be PR not for master. Advise me right branch, please.

samhed:
No. Specifying 'binary' as a subprotocol is not in conformity with the standard.

It was intentionally removed from noVNC in c912230309806aacbae4295faf7ad6406da97617, and support for specifying this was removed from websockify in https://github.com/novnc/websockify/commit/4099949984eb80ef33c2d0dd216991124975a5d2

Your problem is that your Websockify is too old, you will need to use a more recent version.

Old websockify versions fallback on using base64 when no protocol is specified - as is the case with recent noVNC versions. If you upgrade websockify it will correctly fallback on using binary instead, like expected.

samhed:
Related #1276

vitaly-zverev:
Thanks so much, samhed. 
I had to look at IANA sub-protocols list (https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name) instead of head rfc6455 document.
Indeed, registration of new sub-protocol looks like obsession (side effect of 2-weeks selfisolation from covid). My bad. Pardon for turbulence. noVNC  v1.1.0 works like a charm with websockify 0.9.0. 
Combination from websockify 0.8.0 &  noVNC master was real root of problem.

user@ubuntu:~/websockify$ (cd ../noVNC/ && git  checkout master )
Already on 'master'
Your branch is up to date with 'origin/master'.
user@ubuntu:~/websockify$
user@ubuntu:~/websockify$ git checkout v0.8.0
Previous HEAD position was 654cbac websockify 0.9.0
HEAD is now at 1f132f9 Update to 0.8.0
user@ubuntu:~/websockify$
user@ubuntu:~/websockify$ ./websockify.py  --web=../noVNC/  --cert=/etc/ssl/novnc.pem 8888 localhost:5900
WebSocket server settings:
  - Listen on :8888
  - Flash security policy server
  - Web server. Web root: /home/user/noVNC
  - SSL/TLS support
  - proxying from :8888 to localhost:5900
Home.local - - [07/Apr/2020 04:42:07] code 400, message Client must support 'binary' or 'base64' protocol
Home.local - - [07/Apr/2020 04:42:07] code 404, message File not found

samhed:
> noVNC v1.1.0 works like a charm with websockify 0.9.0.

Good to hear!

PR #1394, opened 2020-04-18
update the qemu websocket uri

Akilesh1597:


CendioOssman:
I'm afraid noVNC is not specific to QEMU, so we do not follow their default WebSocket path. You will need to configure noVNC with the path that QEMU requires.

PR #1411, opened 2020-05-31
Enforce camelCase

samhed:
Most of our code is written with camelCase naming. Let's standardize on that and enforce it using our linter.


samhed:
Note that you can't trust Github's commit-order. The commit that actually adds the eslint rules is the last commit.

PR #1415, opened 2020-06-03
Test browsers directly on Travis

CendioOssman:
Travis provides some browsers at least, so let's test those directly without having to go through Sauce Labs.


CendioOssman:
Replaced by #1423.

PR #1417, opened 2020-06-08
Switch to RGBx pixel format

CendioOssman:
This is what the browser wants so it avoids having to spend time converting everything. Unfortunately it usually means the server instead needs to convert it for us, but we assume it has more power than we do.

This was previously how the code worked, see cf67217cbf4be43c05ef3efbd88a3ad8b47744b4. However it was changed in ac99a1f791e71f0b02de748419c21af0f45350b5 for not entirely clear reasons. @kanaka, do you remember the reasoning?


CendioOssman:
I haven't heard anything back from you, @kanaka. So I'm going to go ahead and merge this.

kanaka:
Sorry, I didn't see this until just now. My memory is vague, but I believe the reason was because VMWare VNC server at the time only supported the default ordering. Not sure if that's changed or not.

PR #1425, opened 2020-06-25
Update babel for transpiling

CendioOssman:
Some tweaks were needed for the latest version. Do it know so we don't have to deal with it when we want to change something in the future.


samhed:
:+1:

CendioOssman:
I don't understand what's going on with IE. Why is one build failing and one succeeding?

samhed:
> IE 11.0 (Windows 10) LOG: 'No module support detected. Loading fallback...'
26 06 2020 14:18:11.770:WARN [IE 11.0 (Windows 10)]: Disconnected (0 times), because no message in 30000 ms.
IE 11.0 (Windows 10) ERROR
  Disconnected, because no message in 30000 ms.

Could it be because the transpile takes too long?

CendioOssman:
Looks like it just randomly fails to start the web workers. Not really sure what's going on. IE isn't very good at reporting errors.

Might have to abandon web workers for IE.

CendioOssman:
We've now dropped support for IE, so this PR is no longer useful.

PR #1428, opened 2020-06-26
Make it easier for downstream to modify settings

### LINKED ISSUE #1159: make it easier to customize default host, password, port in the url via code
Is there anyway to pre-load / load the information (host, port, password) into the vnc.html page via PHP ?

This is possible with the vnc_light.html file but seems quite tricky with the vnc.html as it loads all the information from ui.js 

Thanks!

CendioOssman:
Expose a simple and stable API to override default settings, and force settings that users shouldn't be able to change.

Fixes #1159.


CendioOssman:
I just realised this isn't going to work with `use_require.js`. So either that needs to be fixed somehow, or we have to wait until we can drop Internet Explorer and require module support.

Putting this on hold for now.

CendioOssman:
So #1477 is now up, so we just need to wait for that.

tculp:
Bump because this would be very useful and #1477 is now merged

CendioOssman:
It's still on my todo, but I'm unsure when I'll have time to finish this.

vnc.html
```diff
@@ -49,11 +49,40 @@
...
+            if (!response.ok) {
+                throw Error("" + response.status + " " + response.statusText);
+            }
+
+            mandatory = await response.json();
+        } catch (err) {
+            Log.Error("Couldn't fetch mandatory.json: " + err);
+        }
+
+        // You can also override any defaults you need here:
```
samhed, line 85: Neat!

Could we add a comment about how to find names of settings?

defaults.json
```diff
@@ -0,0 +1 @@
+{}
```
samhed, line 1: Should these include a commented-out example?

samhed (changes requested):
Works great, this is a nice improvement!

I tested both setting stuff in vnc.html and in the defaults.json/mandatory.json, things work as expected as long as you clear the web storage inbetween, or use a private-browsing tab.

app/ui.js
```diff
@@ -175,8 +187,6 @@ const UI = {
         UI.initSetting('repeaterID', '');
         UI.initSetting('reconnect', false);
         UI.initSetting('reconnect_delay', 5000);
-
-        UI.setupSettingLabels();
```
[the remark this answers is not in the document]
CendioOssman: The new code disables the UI for mandatory settings. The mapping between controls and labels need to be set up before we go through the settings for the labels to also be disabled properly.

app/ui.js
```diff
@@ -24,6 +24,8 @@ const LINGUAS = ["cs", "de", "el", "es", "fr", "it", "ja", "ko", "nl", "pl", "pt
 
 const UI = {
 
+    userSettings: {},
```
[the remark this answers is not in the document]
CendioOssman: Sure, I understand the confusion. I think "custom" might be even better descriptor.

And the target audience in my mind are mainly sysadmins, not developers.

vnc.html
CendioOssman: Sure. Don't know how to best refer to it, though. Just `"docs/EMBEDDING.md"`?

defaults.json
CendioOssman: JSON doesn't support comments 🤬

CendioOssman:
Hmm... It looks like the tests are broken on `master`. Something with the new H.264 code. Open a new issue and ping the author(s) of that code?

vnc.html
samhed: Sounds good!

samhed (approved):
Looks good!

> Hmm... It looks like the tests are broken on `master`. Something with the new H.264 code. Open a new issue and ping the author(s) of that code?

I'll open a new issue.

Issue #1443, opened 2020-07-14
Failed to start WebSockets proxy


ricardoreis:
I had novnc installed and working normally with root.
But after stopping it, and making some user changes, I can't get it to start anymore.
I am getting the following error:

**Describe the bug**
Warning: could not find self.pem
Using installed websockify at /snap/novnc/5/usr/bin/websockify
Starting webserver and Websockets proxy on port 6081
Traceback (most recent call last):
   File "/snap/novnc/5/usr/bin/websockify", line 5, in <module>
      from pkg_resources import load_entry_point
ImportError: No module named pkg_resources
Failed to start WebSockets proxy

Could you tell me what I can do to try to correct?

**To Reproduce**
1. cd /snap/bin/
2. novnc --listen 6081 --vnc localhost:5901
3. See error

**Expected behavior**
Run novnc

**Screenshots**
1. error: https://prnt.sc/tht2sx
2. folders and files https://prnt.sc/tht3xz
3. vnc: https://prnt.sc/tht4y2
4. vncserver: https://prnt.sc/tht5md
5. Python:  https://prnt.sc/ththp9

**Client (please complete the following information):**
 - OS: Linux Ubuntu 18.04.4 LTS (Bionic Beaver) https://prnt.sc/tht6rp

**Server (please complete the following information):**
 - noVNC version: Which command to get the version? I think it's version 1.2.0. But I'm not sure.
 - VNC server: tightvncserver
 - WebSocket proxy:  I don't know, I didn't install any.

tkedwards:
Unfortunately there have been some issues with the Snap package in the 1.2.0 release. Paths weren't set properly and there was a need to get a latest version of websockify into the snap package. I've created https://github.com/novnc/noVNC/pull/1444 to fix this, but I don't have write access so it's up to the devs here to check it and approve.

In the meantime can you revert back the snap to the previous version?

ricardoreis:
@tkedwards Thank for your response. 

But I successfully ran novnc as follows:

`cd $HOME`
`git clone https://github.com/novnc/noVNC.git`
`cd noVNC/utils/`
`./launch.sh --vnc localhost:5901`

That way the novnc worked.

But I also find it interesting to have a snap solution.
Could you point me in the direction of how to revert back the snap to the previous version? 
I've never done this before.
 
Thank you

tkedwards:
Use _snap revert_ to revert snap package updates. Good to hear you got it working anyway, even if not using snap.

[...]

PR #1444, opened 2020-07-15
Fixing snap package that was broken in the 1.2.0 update:

tkedwards:
- Fix path to launch.sh - it's in the utils/ directory, not the top-level
- Version should be '1.2.0', previously it was some variable which looks like it comes from an external build program (this results in the snap being unable to build)
- Source of the novnc part has been fixed to reference the correct release file on github, not a 'build' directory on local machine (not present in the novnc sources)
- Websockify is now built direct from the 0.9.0 release on github. It appears that the version previously used (from the websockify package on Ubuntu) is too old to support 1.2.0 release of novnc (errors: "code 400, message Client must support 'binary' or 'base64' protocol")
- paths to launch.sh and svc_wrapper.sh have been corrected to where they actually are in the novnc sourc from github

### DIFF
--- a/snap/local/svc_wrapper.sh
+++ b/snap/local/svc_wrapper.sh
@@ -21,7 +21,7 @@ snapctl get services | jq -c '.[]' | while read service; do # for each service t
     listen_port_valid=$?
     if [ ! $listen_port_valid ] || [ -z "$vnc_host_port" ]; then
         # invalid values mean the service is disabled, do nothing except for printing a message (logged in /var/log/system or systemd journal)
-        echo "novnc: not starting service ${service} with listen_port ${listen_port} and vnc_host_port ${vnc_host_port}"
+        echo "novnc-unofficial: not starting service ${service} with listen_port ${listen_port} and vnc_host_port ${vnc_host_port}"
     else
         # start (and fork with '&') the service using the specified listen port and VNC host:port
         $SNAP/utils/launch.sh --listen $listen_port --vnc $vnc_host_port &
--- a/snap/local/svc_wrapper.sh
+++ b/snap/local/svc_wrapper.sh
@@ -24,6 +24,6 @@ snapctl get services | jq -c '.[]' | while read service; do # for each service t
         echo "novnc: not starting service ${service} with listen_port ${listen_port} and vnc_host_port ${vnc_host_port}"
     else
         # start (and fork with '&') the service using the specified listen port and VNC host:port
-        $SNAP/launch.sh --listen $listen_port --vnc $vnc_host_port &
+        $SNAP/utils/launch.sh --listen $listen_port --vnc $vnc_host_port &
     fi
 done
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,6 @@
 name: novnc
 base: core18 # the base snap is the execution environment for this snap
-version: '@VERSION@'
+version: 1.2.0
 summary: Open Source VNC client using HTML5 (WebSockets, Canvas)
 description: |
   Open Source VNC client using HTML5 (WebSockets, Canvas).
@@ -13,13 +13,17 @@ confinement: strict
 
 parts:
     novnc:
-        source: build/
+        source: https://github.com/novnc/noVNC/archive/v1.2.0.tar.gz
         plugin: dump
         stage-packages:
-            - websockify
             - bash
             - jq
-            - python-numpy
+            - python3-numpy
+            
+    websockify:
+        source: https://github.com/novnc/websockify/archive/v0.9.0.tar.gz
+        plugin: python
+        stage-packages:
             - python3-numpy
 
 hooks:
@@ -28,9 +32,9 @@ hooks:
 
 apps:
     novnc:
-        command: ./launch.sh
+        command: utils/launch.sh
         plugs: [network, network-bind]
     novncsvc:
-        command: ./svc_wrapper.sh
+        command: snap/local/svc_wrapper.sh
         daemon: forking
         plugs: [network, network-bind]
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -14,7 +14,8 @@ confinement: strict
 parts:
     novnc:
         source: https://github.com/tkedwards/noVNC.git
-        plugin: git
+        source-type: git
+        plugin: dump
         stage-packages:
             - bash
             - jq
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,4 +1,4 @@
-name: novnc
+name: novnc-unofficial
 base: core18 # the base snap is the execution environment for this snap
 version: 1.2.0
 summary: Open Source VNC client using HTML5 (WebSockets, Canvas)
@@ -13,15 +13,15 @@ confinement: strict
 
 parts:
     novnc:
-        source: https://github.com/novnc/noVNC/archive/v1.2.0.tar.gz
-        plugin: dump
+        source: https://github.com/tkedwards/noVNC.git
+        plugin: git
         stage-packages:
             - bash
             - jq
-            - python3-numpy
             
     websockify:
-        source: https://github.com/novnc/websockify/archive/v0.9.0.tar.gz
+        source: https://github.com/novnc/websockify.git
+        source-type: git
         plugin: python
         stage-packages:
             - python3-numpy
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -1,3 +1,3 @@
 #!/bin/sh -e
 
-snapctl restart novnc.novncsvc
+snapctl restart novnc-unofficial.novncsvc


snap/snapcraft.yaml
```diff
@@ -1,6 +1,6 @@
 name: novnc
 base: core18 # the base snap is the execution environment for this snap
-version: '@VERSION@'
+version: 1.2.0
```
CendioOssman, line 3: We don't want to have to keep updating this, so please keep it as is. It is handled by the deploy job.

snap/snapcraft.yaml
```diff
@@ -13,13 +13,17 @@ confinement: strict
 
 parts:
     novnc:
-        source: build/
+        source: https://github.com/novnc/noVNC/archive/v1.2.0.tar.gz
```
CendioOssman, line 16: We want it to build the working directory, not something else.

snap/local/svc_wrapper.sh
```diff
@@ -24,6 +24,6 @@ snapctl get services | jq -c '.[]' | while read service; do # for each service t
         echo "novnc: not starting service ${service} with listen_port ${listen_port} and vnc_host_port ${vnc_host_port}"
     else
         # start (and fork with '&') the service using the specified listen port and VNC host:port
-        $SNAP/launch.sh --listen $listen_port --vnc $vnc_host_port &
+        $SNAP/utils/launch.sh --listen $listen_port --vnc $vnc_host_port &
```
CendioOssman, line 27: This is also handled by the deploy job.

snap/snapcraft.yaml
```diff
@@ -13,13 +13,17 @@ confinement: strict
...
         plugin: dump
         stage-packages:
-            - websockify
             - bash
             - jq
-            - python-numpy
+            - python3-numpy
```
CendioOssman, line 21: Why do we need this change all of a sudden?

snap/snapcraft.yaml
```diff
@@ -13,13 +13,17 @@ confinement: strict
...
+            
+    websockify:
+        source: https://github.com/novnc/websockify/archive/v0.9.0.tar.gz
```
CendioOssman, line 24: Sounds good. But is there some special URL that always gives us the latest release? We'd like to avoid having to update this.

snap/snapcraft.yaml
```diff
@@ -13,13 +13,17 @@ confinement: strict
...
+        plugin: python
+        stage-packages:
             - python3-numpy
```
CendioOssman, line 27: I assume this means numpy should be removed earlier?

snap/snapcraft.yaml
```diff
@@ -13,13 +13,17 @@ confinement: strict
 
 parts:
     novnc:
-        source: build/
+        source: https://github.com/novnc/noVNC/archive/v1.2.0.tar.gz
         plugin: dump
         stage-packages:
-            - websockify
```
CendioOssman, line 19: Is this a snap package? Perhaps it should be updated in that case?

snap/snapcraft.yaml
tkedwards: python-numpy is the Python 2 package (Ubuntu apt package). I realised it's not necessary, only python3-numpy is.

snap/snapcraft.yaml
tkedwards: Yes, you can easily do this:
`    websockify:
        source: https://github.com/novnc/websockify.git
        source-type: git
`

The only thing is then you're potentially getting a different version of websockify each time you build the snap, and that version may be a development version instead of a tested released version like 0.9.0.
You probably know more about the websockify project than me and can say whether this is a risk or not

snap/snapcraft.yaml
tkedwards: Hmm, does novnc use numpy at all? If not then it should be removed, I had it in there previously as it was needed for the Debian/Ubuntu websockify package

snap/snapcraft.yaml
tkedwards: That was the Debian/Ubuntu websockify apt package, and is a bit out of date (in 18.04 at least). Not needed anymore now we're grabbing and building websockify directly in the snap

snap/snapcraft.yaml
tkedwards: Actually python3-numpy can probably go too, if it's not something novnc itself uses. I had those python*-numpy packages in there as they seemed to be dependencies of the websockify apt package, which is now no longer used

snap/snapcraft.yaml
tkedwards: I always thought the snap be built from what's released, i.e 1.2.0 release in this case. Is that 'build' a working directory on your computer? I guess either way, as long as it works then all good

zhogov:
@CendioOssman please take a look at this, snap package is completely broken (#1443) and all channels only have `1.2.0` so I just can't install older version anymore

snap/snapcraft.yaml
CendioOssman: On the computer building the snap, yes. I'd say that is more expected than the snap pulling down something else and packaging.

snap/snapcraft.yaml
CendioOssman: Indeed. We have no python in noVNC, so it's strictly a websockify thing.

snap/snapcraft.yaml
CendioOssman: That's probably not a good idea. I was thinking the latest **release**, not the latest **commit**. So something like https://github.com/novnc/websockify/archive/latest.tar.gz. But if that doesn't exists I guess we'll have to live with updating that number.

CendioOssman:
For reference, the deploy job is here:

https://github.com/novnc/noVNC/blob/master/.github/workflows/deploy.yml

If you want to test things the same way the real job would.

snap/snapcraft.yaml
tkedwards: From your comment down below I see you're using Github workflows to do the build, so that build/ directory is something created on the VM running on github's servers when the publish workflow is running.
That's fine, but it means no one else can build the snap, so it'd be up to you to re-run that workflow to fix it when the changes have been made

snap/snapcraft.yaml
tkedwards: In that case, python3-numpy can be removed too

tkedwards:
@CendioOssman 
Now I understand you're using that github workflow thing to build the snap it really comes down to 2 changes:

1. Adding a 'part' for websockify to bring in 0.9.0
2. Removing the unneeded stage-packages from the novnc part (since websockify is now included)

I've added below what should be the final snapcraft.yaml file. I can't do anything further because I can't build the snap (I can't run github workflows that are in your repo). Are you right to continue further with this - build, test the snap and submit to snap store?

```
name: novnc
base: core18 # the base snap is the execution environment for this snap
version: '@VERSION@'
summary: Open Source VNC client using HTML5 (WebSockets, Canvas)
description: |
  Open Source VNC client using HTML5 (WebSockets, Canvas).
  noVNC is both a VNC client JavaScript library as well as an
  application built on top of that library. noVNC runs well in any
  modern browser including mobile browsers (iOS and Android).

grade: stable
confinement: strict

parts:
    novnc:
        source: build/
        plugin: dump
        stage-packages:
            - bash
            - jq

    websockify:
        source: https://github.com/novnc/websockify/archive/v0.9.0.tar.gz
        plugin: python
        stage-packages:
            - python3-numpy

hooks:
    configure:
        plugs: [network, network-bind]

apps:
    novnc:
        command: ./launch.sh
        plugs: [network, network-bind]
    novncsvc:
        command: ./svc_wrapper.sh
        daemon: forking
        plugs: [network, network-bind]
```

CendioOssman:
I'm afraid I'm currently on vacation so I'm a bit limited in what I can do from here. But I'll get back to this when I can.

> I can't do anything further because I can't build the snap (I can't run github workflows that are in your repo).

You can test the same steps manually. The github workflow is basically just a script, so you can run the same commands by hand and see if they produce a good snap for you.

tkedwards:
I've added a Snap 'novnc-unofficial' to the store, so that we have a working Snap until the official one can be fixed.

The installation/use instructions are the same as already documented at https://github.com/novnc/noVNC#installation-from-snap-package, except for these slight changes:

`sudo snap install novnc-unofficial`

`novnc-unofficial --listen 6081 --vnc localhost:5901 `

`sudo snap set novnc-unofficial services.n6082.listen=6082 services.n6082.vnc=localhost:5902`

CendioOssman:
I'm back now so I was planning to see if I can finish this work. However I see that someone has already pushed an updated 1.2.0 version to snapcraft. Was that you, @tkedwards? And what build script was used for that?

tkedwards:
Welcome back!
I just created a different Snap called 'novnc-unofficial', so we'd had something that works until the 'official' snap is fixed. I did that in my fork of this project, so I think the changes come though in this pull request, but they can be ignored.

I set out how the snap should work in https://github.com/novnc/noVNC/pull/1444#issuecomment-663498867
Can you try putting that through the Github build system and see if it produces a working snap?

CendioOssman:
Sorry, things have been (and are) hectic here.

What I meant was that when this thread started the official novnc snap was at version 5. However someone has posted a version 6 that works fine for me. Could you have a look at it and see if it works fine for you as well?

Also, I'm not familiar with snap. Will users get an upgrade from 5 to 6 even if both have the same version id? (i.e. "1.2.0")?

(I will also get your changes tested and committed once I have some more time)

tkedwards:
I'm not sure who uploaded that version 6 but I just tested it and it all works. It wasn't me who did that, I don't have access to the project to do it. Good to see it fixed though.

Yes I think snap will automatically upgrade people to version 6 of the snap.

CendioOssman:
The upload logs claim that it was your account at least. :)

I've committed the necessary fix for the next release as 499eee4d069c34bb986324771ff95d391550ac31. But we should be gone with version 6 until then.

tkedwards:
Which upload logs are you looking at??

I don't have any access to the novnc snap since I handed over control to you. That's why I created a different snap novnc-unofficial snap as a temporary fix (now withdrawn since the main snap has been fixed).

CendioOssman:
The entry is here:

https://dashboard.snapcraft.io/snaps/novnc/revisions/6/

For me it states "Submitted by tkedwards". I would guess you pushed to the wrong repo. :)

And yes, you still have access. I left you there since you have a better grasp on these things than me. :)

tkedwards:
I must have done it half asleep one day, probably thought I was updating the novnc-unofficial snap!

Anyway, all good, glad I still have access too :-)

PR #1445, opened 2020-07-20
fix chinese garbled code

gongym12138:
change clipboardPasteFrom unicode encode

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -493,8 +493,8 @@ export default class RFB extends EventTargetMixin {
             for (let i = 0; i < text.length; i++) {
                 // FIXME: text can have values outside of Latin1/Uint8
                 let idx = text.charCodeAt(i);
-                data[i] = idx & 0xff;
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm afraid we cannot accept this code as it would violate the VNC protocol specification. Which server are you using this with?

gongym12138:
> I'm afraid we cannot accept this code as it would violate the VNC protocol specification. Which server are you using this with?

gongym12138:
> I'm afraid we cannot accept this code as it would violate the VNC protocol specification. Which server are you using this with?

I think it's also a server problem 

use https://www.tightvnc.com/download.php vncserver

gongym12138:
> 比如说，你这个修改是修复没有VNC剪切板无法复制中文的问题么？
> 你最终有没有解决剪切板中文乱码问题？如果解决了，能否分享一下你的解决方案么？

我提交的代码就是修复了这个问题

PR #1451, opened 2020-08-10
launch.sh --listen option handle bind IP

t-matsuo:
Websockify can handle bind IP but launch.sh cannot handle it.
See https://github.com/novnc/websockify/issues/411.

### DIFF
--- a/utils/launch.sh
+++ b/utils/launch.sh
@@ -13,7 +13,7 @@ usage() {
     echo "Starts the WebSockets proxy and a mini-webserver and "
     echo "provides a cut-and-paste URL to go to."
     echo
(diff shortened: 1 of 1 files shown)


samhed:
I'm fine with adding this, but I suspect it was made like this intentionally. The launch script is supposed to be simpler.

t-matsuo:
I run noVNC behind reverse proxy on same host, so noVNC just need to listen at 127.0.0.1.
It's useful to run noVNC in my script.

Sould I run noVNC and websockify manually instead of using launch script?
If yes, I will close this PR.

samhed:
Yeah, I guess we should draw the line somewhere. The launch script is for quick and easy testing. It doesn't have to support all kinds of setups.

Issue #1436, opened 2020-07-08
Failed while connected: Error decoding rect: TypeError


brathering82:
The connection works. For a split second I can see the remote screen. Then the connection closes.

The browser log is as follows:

> User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/83.0.4103.116 Safari/537.36
> display.js:69 << Display.constructor
> rfb.js:288 << RFB.constructor
> rfb.js:760 New state 'connecting', was ''.
> rfb.js:505 >> RFB.connect
> rfb.js:507 connecting to ws://192.168.15.89:6080/websockify
> rfb.js:554 << RFB.connect
> websock.js:190 >> WebSock.onopen
> rfb.js:242 Starting VNC handshake
> websock.js:196 << WebSock.onopen
> rfb.js:1220 Server ProtocolVersion: 003.003
> rfb.js:1260 Sent ProtocolVersion: 003.003
> rfb.js:1322 Authenticating using scheme: 2
> ui.js:1180 Server asked for credentials
> credentials @ ui.js:1180
> (anonymous) @ eventtarget.js:32
> dispatchEvent @ eventtarget.js:32
> _negotiateStdVNCAuth @ rfb.js:1470
> _negotiateAuthentication @ rfb.js:1621
> _initMsg @ rfb.js:1804
> _negotiateSecurity @ rfb.js:1324
> _initMsg @ rfb.js:1801
> _handleMessage @ rfb.js:856
> (anonymous) @ rfb.js:236
> _recvMessage @ websock.js:283
> rfb.js:1645 Authentication OK
> rfb.js:1719 Screen: 1024x768, bpp: 32, depth: 32, bigEndian: 0, trueColor: 1, redMax: 255, greenMax: 255, blueMax: 255, redShift: 16, greenShift: 8, blueShift: 0
> rfb.js:760 New state 'connected', was 'connecting'.
> **rfb.js:808 Failed while connected: Error decoding rect: TypeError: Failed to execute 'createImageData' on 'CanvasRenderingContext2D': Value is not of type 'long'.**
> _fail @ rfb.js:808
> _handleDataRect @ rfb.js:2473
> _handleRect @ rfb.js:2208
> _framebufferUpdate @ rfb.js:2160
> _normalMsg @ rfb.js:2076
> _handleMessage @ rfb.js:847
> (anonymous) @ rfb.js:236
[...]

PR #1460, opened 2020-08-31
Ignore rects with zero width or height

samhed:
Avoids having to handle these cases in the decoders. Should solve #1436


CendioOssman:
I'm afraid this code is incorrect as it assumes an empty rect is encoded as zero bytes, which isn't true for most encodings. If we wish to have a resilent protection we need to add unit tests for this corner case.

PR #1477, opened 2020-10-15
Drop support for Internet Explorer and legacy Edge

CendioOssman:
Now that IE is finally being abandoned by everyone we can start removing all that legacy handling we have in place.

This is a start which removes the most obvious bits, but we have lots of more things that can probably be removed. It needs testing with current browsers first though.


CendioOssman:
Should be ready now. Needs some proper testing on the major browsers though to make sure something wasn't missed.

CendioOssman:
Tested on Chrome on Linux and Android, Firefox on Linux, Edge on Windows and Safari on macOS and iOS. Good enough for me.

PR #1483, opened 2020-10-28
Update ui.js

adrianopedro:
Add wsProtocols options array for better compatibility, specially with x11vnc server.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1024,9 +1024,13 @@ const UI = {
         url += '/' + path;
 
         UI.rfb = new RFB(document.getElementById('noVNC_container'), url,
-                         { shared: UI.getSetting('shared'),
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm afraid this is not something we want. We explicitly removed those as they are nonstandard. Please use a current version of x11vnc. There are other bugfixes that are needed in its WebSocket handling so this is not the only issue.

adrianopedro:
I need to make it compatible with Raspbian buster that only supports x11vnc 0.9.13.
At least this x11vnc version has this problem and this is the way to solve it.
Anyway... is up to you..

Regards...

CendioOssman:
If you need to connect to an old buggy version of x11vnc then I would recommend using a separate websocket proxy in front of it.

PR #1500, opened 2020-12-16
guesture that activates the keyboard view

sportegioco:
I want to sincerely thank all of you for the addition of the guestures that have allowed the use of novnc in all touch devices much easier.
It would be useful to have a guesture that activates the keyboard view, or even through an icon that appears near the cursor as in the image that I am attaching.
Thanks
Gianpaolo

guesture


CendioOssman:
Not a PR, so please see #1501.

PR #1523, opened 2021-02-18
Added support for RTCDataChannel

TimSBSquare:
This work is originally by Ryan Castner <castner.rr@gmail.com> and submitted as a PR here https://github.com/novnc/noVNC/pull/1362

Architecturally it is much the same except it doesn't rename a lot of variables to make this more reviewable. It also avoids unrelated changes such as replacing .onclose with an event listener, which caused numerous test failures.

It also adds in ppoffice's fix to initialise the buffers.

Like the original author I don't have enough time available to refactor this project to the new style event listeners.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -66,20 +66,25 @@ const extendedClipboardActionPeek    = 1 << 26;
 const extendedClipboardActionNotify  = 1 << 27;
 const extendedClipboardActionProvide = 1 << 28;
 
-
 export default class RFB extends EventTargetMixin {
-    constructor(target, url, options) {
+    constructor(target, urlOrChannel, options) {
         if (!target) {
             throw new Error("Must specify target");
         }
-        if (!url) {
-            throw new Error("Must specify URL");
+        if (!urlOrChannel) {
+            throw new Error("Must specify URL, WebSocket or RTCDataChannel");
         }
 
         super();
 
         this._target = target;
-        this._url = url;
+
+        if (typeof urlOrChannel === "string") {
+            this._url = urlOrChannel;
+        } else {
+            this._url = null;
+            this._rawChannel = urlOrChannel;
+        }
 
         // Connection details
         options = options || {};
@@ -275,6 +280,8 @@ export default class RFB extends EventTargetMixin {
                     break;
             }
             this._sock.off('close');
+            // Delete reference to raw channel to allow cleanup.
+            this._rawChannel = null;
         });
         this._sock.on('error', e => Log.Warn("WebSocket on-error event"));
 
@@ -501,16 +508,23 @@ export default class RFB extends EventTargetMixin {
     _connect() {
         Log.Debug(">> RFB.connect");
 
-        Log.Info("connecting to " + this._url);
-
-        try {
-            // WebSocket.onopen transitions to the RFB init states
-            this._sock.open(this._url, this._wsProtocols);
-        } catch (e) {
-            if (e.name === 'SyntaxError') {
-                this._fail("Invalid host or port (" + e + ")");
-            } else {
-                this._fail("Error when opening socket (" + e + ")");
+        if (this._url) {
+            try {
+                Log.Info(`connecting to ${this._url}`);
+                this._sock.open(this._url, this._wsProtocols);
+            } catch (e) {
+                if (e.name === 'SyntaxError') {
+                    this._fail("Invalid host or port (" + e + ")");
+                } else {
+                    this._fail("Error when opening socket (" + e + ")");
+                }
+            }
+        } else {
+            try {
+                Log.Info(`attaching ${this._rawChannel} to Websock`);
+                this._sock.attach(this._rawChannel);
+            } catch (e) {
+                this._fail("Error attaching channel (" + e + ")");
             }
         }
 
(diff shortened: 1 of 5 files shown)


core/rfb.js
```diff
@@ -66,20 +66,50 @@ const extendedClipboardActionPeek    = 1 << 26;
...
 export default class RFB extends EventTargetMixin {
-    constructor(target, url, options) {
+    constructor(target, urlOrChannel, options) {
         if (!target) {
             throw new Error("Must specify target");
         }
-        if (!url) {
-            throw new Error("Must specify URL");
+        if (!urlOrChannel) {
+            throw new Error("Must specify URL or raw channel");
```
CendioOssman, line 88: Can we be more explicit and state "WebSocket or RTCDataChannel" rather than "raw channel"?

core/rfb.js
```diff
@@ -66,20 +66,50 @@ const extendedClipboardActionPeek    = 1 << 26;
...
+            this._url = null;
+            this._rawChannel = urlOrChannel;
+            for (let i = 0; i < rawChannelProps.length; i++) {
+                const prop = rawChannelProps[i];
+                if (urlOrChannel[prop] === undefined) {
+                    console.error('Raw channel missing property', prop);
+                    this._rawChannel = null;
+                    break;
+                }
+            }
```
CendioOssman, line 107: I guess it's the WebSock object that is mostly concerned about these things, so can we put the check there? That way all of the code is in the same place.

core/rfb.js
```diff
@@ -66,20 +66,50 @@ const extendedClipboardActionPeek    = 1 << 26;
...
+        }
+
+        if (!this._url && !this._rawChannel) {
+            throw new Error("Must specify URL or raw channel");
+        }
```
CendioOssman, line 112: Can we really get here?

core/rfb.js
```diff
@@ -124,7 +154,7 @@ export default class RFB extends EventTargetMixin {
         this._clipboardServerCapabilitiesFormats = {};
 
         // Internal objects
-        this._sock = null;              // Websock object
+        this._sock = null;              // WebSocket or RTCDataChannel object
```
CendioOssman, line 157: I don't think this changed?

core/rfb.js
```diff
@@ -66,20 +66,50 @@ const extendedClipboardActionPeek    = 1 << 26;
...
         super();
 
         this._target = target;
-        this._url = url;
+
+        if (typeof urlOrChannel === "string") {
+            this._url = urlOrChannel;
+        } else {
+            this._url = null;
+            this._rawChannel = urlOrChannel;
```
CendioOssman, line 99: Should we also drop this reference on disconnect to allow the gc to clean up?

core/websock.js
```diff
@@ -177,13 +193,18 @@ export default class Websock {
...
 
     open(uri, protocols) {
-        this.init();
+        const rawChannel = new WebSocket(uri, protocols);
+        this.attach(rawChannel, false);
+    }
 
-        this._websocket = new WebSocket(uri, protocols);
-        this._websocket.binaryType = 'arraybuffer';
+    attach(rawChannel, isOpen = false) {
```
CendioOssman, line 200: Can't we check readyState instead of having to specify the isOpen variable?

CendioOssman (changes requested):
Nice, thanks. Some minor things which I've added comments on.

Could you also update the API documentation?

core/rfb.js
TimSBSquare: At least not anymore

core/rfb.js
TimSBSquare: I've put it in the _sock close listener, please let me know if that is not what you intended

core/rfb.js
TimSBSquare: This turned out to have unexpected repercussions. The tests weren't always using the FakeWebsocket and the FakeWebsocket didn't have some properties set to null that were expected to be.

TimSBSquare:
I've left a couple of points open for your confirmation otherwise everything is resolved.

core/rfb.js
CendioOssman: I think this works. I assume we should always end up there?

core/rfb.js
CendioOssman: But I guess you get it working?

core/rfb.js
TimSBSquare: It works and as far as I'm aware yes

core/rfb.js
TimSBSquare: Yep. Just wanted you to be aware of the additional changes

CendioOssman (approved):
Looks good. Thanks.

Please squash your fixes together to a single commit and do a force push and we can go ahead and merge this.

TimSBSquare:
Build error appears to be transient and I have no way to retrigger:
```
snapd is not logged in, snap install commands will use sudo
sudo: unable to resolve host ubuntu: Resource temporarily unavailable
error: cannot refresh "core18": unexpectedly empty response from the server
       (try again later)
Error while refreshing snap 'core18' to channel 'latest/stable'
```

CendioOssman:
That's something unrelated, so it's not a blocker for merging this. :)

PR #1529, opened 2021-03-04
Allow longer passwords in Plain authentication

lhchavez:
Some people have longer passwords than 256 characters (hooray for
password managers!). Server implementations also allow longer passwords:
TigerVNC allows up to 1024 characters.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1438,9 +1438,18 @@ export default class RFB extends EventTargetMixin {
             const user = encodeUTF8(this._rfbCredentials.username);
             const pass = encodeUTF8(this._rfbCredentials.password);
 
-            // XXX we assume lengths are <= 255 (should not be an issue in the real world)
-            this._sock.send([0, 0, 0, user.length]);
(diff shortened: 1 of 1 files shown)


core/rfb.js
```diff
@@ -1438,9 +1438,18 @@ export default class RFB extends EventTargetMixin {
...
 
-            // XXX we assume lengths are <= 255 (should not be an issue in the real world)
-            this._sock.send([0, 0, 0, user.length]);
-            this._sock.send([0, 0, 0, pass.length]);
+            this._sock.send([
+                (user.length >> 24) & 0xFF,
+                (user.length >> 16) & 0xFF,
+                (user.legnth >> 8) & 0xFF,
+                user.length & 0xFF
+            ]);
```
lhchavez, line 1446: i could also go with adding something that would allow us to write `this._sock.sQput32(user.length)`. let me know what flavor is preferred.

CendioOssman: This is how we do it elsewhere, so it's fine for now. :)

PR #1532, opened 2021-03-10
Added details for no-store caching to EMBEDDING.md

ltomes:
Hi there, I saw an update in unraid around `no-cache` headers, and thought the configuration could be improved, and found my way to the EMBEDDING.md file.

# Thinking
If users are using NOVNC across WAN or a reverse proxy/vpn, there may be caching at play beyond browser caching.
In particular this may be useful for Unraids configuration.
Adding a `no-store` Cache-Control header can prevent unintended caching at the ISP level/levels above the browser cache.

Let me know if that sounds reasonable, or if I'm overlooking something. Or if opening a pull request for this is unreasonable/not your process. Thanks for the cool and useful project.

### DIFF
--- a/docs/EMBEDDING.md
+++ b/docs/EMBEDDING.md
@@ -83,7 +83,8 @@ browser reuses from its cache). To avoid this issue, the browser must be told
 to always revalidate cached files using conditional requests. The correct
 semantics are achieved via the (confusingly named) `Cache-Control: no-cache`
 header that needs to be provided in the web server responses.
-
+In additon to the `no-cache` header, the `no-store` header can be used to
+ prevent similar caching issues at caching layers above the browser level.
 ### Example Server Configurations
 
 Apache:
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm afraid I don't see the effectiveness of this change. If intermediate caches are ignoring `no-cache`, why do we expect to respect `no-store`?

ltomes:
> I'm afraid I don't see the effectiveness of this change. If intermediate caches are ignoring `no-cache`, why do we expect to respect `no-store`?

Hi @CendioOssman , `no-cache` was never intended to be respected in caches above the browser. Its name is a misnomer. 

> no-cache
> The response may be stored by any cache, even if the response is normally non-cacheable. However, the stored response MUST always go through validation with the origin server first before using it, therefore, you cannot use no-cache in-conjunction with immutable. If you mean to not store the response in any cache, use no-store instead. This directive is not effective in preventing caches from storing your response.

From: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#cacheability

CendioOssman:
We like caching. We have loads of files, so `no-cache` is not really desirable either and we'd prefer something lighter than that. :)

As to the matter at hand, I'm afraid I don't see anything in that link that suggests that `no-cache` only affects browsers?

ltomes:
> We like caching. We have loads of files, so `no-cache` is not really desirable either and we'd prefer something lighter than that. :)

Understood, I think you have what you want then/don't need these notes added.

> As to the matter at hand, I'm afraid I don't see anything in that link that suggests that `no-cache` only affects browsers?

This was probably poor communication on my part, by 'only affects browsers' I was intending to communicate that the files _can_ be cached in any caches above the browser. Which noVNC wants based on the above, but 3rd parties may not/may assume `no-cache` will not cache files. In any case though, it's not noVNCs responsibility to repeat documentation on headers across all your docs, and you want the caching; I think this can be closed.

ltomes:
Closing:

@CendioOssman  Unrelated to the above changes, potentially adding `max-age=0` only for vnc.html file is desired? That leaves reasonable caching on all the other files in your bundles, but guarantees immediate update of your root file.
It would be a more challenging configuration to communicate to users though.

Also, thanks for the discussion.

PR #1533, opened 2021-03-11
Normalize the credentials presence check

lhchavez:
Most places that check for the presence / absence of credentials compare
them against `undefined`, except the one for Plain authentication.

This change makes the very last place to use the same pattern (instead
of checking for falsiness) for consistency. Additionally, there are ways
to configure PAM to accept empty passwords, so it's possible for a user
to legitimately send an empty string as password.


CendioOssman:
Seems reasonable enough. Some unit tests for this would be nice, but I see we don't have any at all for the Plain handling so I won't let that be a blocker.

PR #1535, opened 2021-03-13
Add VeNCrypt Plain authentication tests

lhchavez:
This change adds tests for the VeNCrypt Plain authentication. In doing
that, this also fixes a typo that was introduced in a recent change.


CendioOssman:
Nice, thanks.

PR #1542, opened 2021-04-18
Fix connection race with existing WebSocket/RTCDatachannel

CendioOssman:
New attempt at #1537. Instead of bypassing `setTimeout()` when using an existing object, we instead get rid of the reason for that timeout.

@TimSBSquare , could you test this? I don't have any setup here that uses existing objects.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -234,56 +234,10 @@ export default class RFB extends EventTargetMixin {
         this._gestures = new GestureHandler();
 
         this._sock = new Websock();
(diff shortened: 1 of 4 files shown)


TimSBSquare:
Thanks for the alternate fix. I'll give it a test tomorrow

TimSBSquare:
@CendioOssman I've verified this works as expected

CendioOssman:
Great. Thanks for testing.

PR #1544, opened 2021-04-23
Update Spanish Translation

ascillato:
Update Spanish Translation.

Added missing translations. Fixed some typos.
Also, some changes were made to use a more formal and international spanish.


ascillato:
I have saw the requirement that the translations should be done in the PO files...

PR #1572, opened 2021-08-24
Adding Security & Privacy feature

yatru:
The changes add more privacy and security to noVNC, it allows the user to use the novnc client without logging in the server sensitive parameters like (password/path etc..) by using a hastag # instead of a question mark. It keeps the old method of indicating the parameters but add this extra option.

Examples:
 A URL with a query parameter can look like this (But will most probably get logged on the http server):
 `https://www.example.com?myqueryparam=myvalue`
 
  For privacy (Using a hastag #, the parameters will not be sent to the server)
 the url can be requested in the following way:
 `https://www.example.com#myqueryparam=myvalue&password=secreatvalue`

Even Mixing public and non public parameters will work:
 `https://www.example.com?nonsecretparam=example.com#password=secreatvalue`

thanks

### DIFF
--- a/app/webutil.js
+++ b/app/webutil.js
@@ -26,13 +26,13 @@ export function initLogging(level) {
 // For privacy (Using a hastag #, the parameters will not be sent to the server)
 // the url can be requested in the following way:
 // https://www.example.com#myqueryparam=myvalue&password=secreatvalue
-// 
+//
 // Even Mixing public and non public parameters will work:
 // https://www.example.com?nonsecretparam=example.com#password=secreatvalue
 export function getQueryVar(name, defVal) {
     "use strict";
     const re = new RegExp('.*[?&]' + name + '=([^&#]*)'),
-        match = ''.concat(document.location.href," ", window.location.hash).match(re);
+        match = ''.concat(document.location.href, " ", window.location.hash).match(re);
     if (typeof defVal === 'undefined') { defVal = null; }
 
     if (match) {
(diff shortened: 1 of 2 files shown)


CendioOssman:
Looks like a good addition, thanks. Could you squash your commits and force push things so everything is in a single commit?

yatru:
> Looks like a good addition, thanks. Could you squash your commits and force push things so everything is in a single commit?

Thanks to you, please, tell me if the new Merge branch is ok ? This is my first commit

CendioOssman:
Not quite. It seems we got some duplication. Let's try this:

1. Make sure we only have your merged commit: `git reset --hard 194af1bc324c47cd97486f4a49dfbaa0cd3e624e`
2. Fix up the commit message to avoid the duplication there: `git commit --amend`
3. Force an update to this PR: `git push --force`

yatru:
Thanks , please , tell me now if its fine

yatru (commented):
I fixed the extra spaces problem

samhed:
The issue I mentioned in my last comment is still there. The commit message for the first commit looks like this:

> Added security and privacy feature, use private parameters after hastag to avoid server logging
> 
> Added security and privacy feature, use private parameters after hastag to avoid server logging
> 
> Added security and privacy feature, use private parameters after hastag to avoid server logging

Remove 2 of those lines please.

Also, the two commits you added for fixing the extra spaces should be squashed into your original commit. We have to be picky in a project of this size.

yatru:
I am not versed on git, I will fork again and commit the changes

yatru:
Hi Samhed,
Please check this one 
https://github.com/novnc/noVNC/pull/1573

PR #1573, opened 2021-08-31
Security privacy to url parameters

yatru:
The changes add more privacy and security to noVNC, it allows the user to use the novnc client without logging in the server sensitive parameters like (password/path etc..) by using a hastag # instead of a question mark. It keeps the old method of indicating the parameters but add this extra option.

Examples:
A URL with a query parameter can look like this (But will most probably get logged on the http server):
https://www.example.com?myqueryparam=myvalue

For privacy (Using a hastag #, the parameters will not be sent to the server)
the url can be requested in the following way:
https://www.example.com#myqueryparam=myvalue&password=secreatvalue

Even Mixing public and non public parameters will work:
https://www.example.com?nonsecretparam=example.com#password=secreatvalue

thanks


yatru:
Thanks Samhed and all the contributors for this great software, I hope to contribute more

Issue #1502, opened 2020-12-18
two-step 'disconnect' option with custom message


tlcarpenter:
**Is your feature request related to a problem? Please describe.**
The request is related to a security issue I submitted to Apple. I'm wondering if there is a way to communicate privately via email, if necessary, about the issue.

**Describe the solution you'd like**
An option to display a custom message when noVNC's 'disconnect' button is clicked; the message would be displayed prior to noVNC being disconnected and would give the user the option to cancel the 'disconnect' action or to proceed with disconnecting.

**Describe alternatives you've considered**
Not aware of any.

**Additional context**

samhed:
You'd like a confirmation popup when disconnecting? Why would this have to be custom? And what does it have to do with Apple?

CendioOssman:
I'm not so sure about the ability to customize this. But a general message could be interesting.

One issue though is how to hide it for users that understand how this works. E.g. a "Don't show this again" check box. We don't currently have any mechanism for that.

samhed:
Couldn't such a "don't show this again"-choice be saved alongside the rest of the UI settings in the web storage?

CendioOssman:
Yes, sorry. I got this confused with TigerVNC where we don't have as flexible storage. :)

yatru:
In the same line, could be added an option to send some key combination to logout the remote user then you click "disconnect" being those key combinations configurable in the parameters, for example for windows & ubuntu (windows + L) to lock the user or (Ctrl+Alt+Del) to logout in ubuntu

PR #1575, opened 2021-09-03
Adding disconnect_message Confirmation dialog

yatru:
When parameter disconnect_message is set, it will show a message to confirm the disconnection being its value the text on the confirm dialog.

This is related to the issue #1502 "two-step 'disconnect' option with custom message #1502"

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1047,6 +1047,10 @@ const UI = {
     },
 
     disconnect() {
+        // When parameter disconnect_message is set, it will show a message to confirm the disconection being its value the text on the confirm dialog
+        if (WebUtil.getConfigVar('disconnect_message', false) && confirm( WebUtil.getConfigVar('disconnect_message') ) == false ) {
+            return;
+        }
         UI.rfb.disconnect();
 
         UI.connected = false;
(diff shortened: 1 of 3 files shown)


CendioOssman:
I would prefer if we start a bit more restrictive and have our own message. That way we can have it enabled by default, and we can make sure it is properly translated.

yatru:
if that the case, we can reuse the "Disconnect" string that is already translated and add a question mark

` if (WebUtil.getConfigVar('disconnect_message', false) && confirm(_("Disconnect") + " ?") == false ) {
            return;
 }`

What do you think about it ?

samhed:
Seeing as question marks work differently in different languages (for example spanish) we would need a different string anyway. No need to reuse something existing.

I suggest "Are you sure you want to disconnect?"

samhed (commented):
It would be nice if we could have a "Don't show this again" checkbox. For that we would probably have to create our own dialog for this question instead of using `window.confirm()`. An alternative would be to have it in the settings somewhere.

Either way we would need to save a new setting for it.

yatru:
I have the code for the custom dialog, I will upload it

CendioOssman:
Have you been able to have a look at this, @yatru?

CendioOssman:
No response. Closing.

PR #1283, opened 2019-08-18
zrle and zywrle support

Ekstazi:
Support for ZRLE encoding
https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#zrle-encoding

I didn't benchmark it. But as my opinion it slightly slower than tight.

PR #1587, opened 2021-10-04
Added support for ZRLE encoding

rosmo:
Supports all ZRLE encodings. I would love some feedback if I should be using the display queue and/or buffer sizing. I had to touch the inflator since ZRLE doesn't give me a clear indication of how much uncompressed bytes are incoming and there is some problem retrieving 1 byte at a time from the inflator.

Also wondering if it'd be ok to add 16-bit pixel format? (keeping the framebuffer at 24 bits, but passing the true bpp value to the decoders)

### DIFF
--- a/core/decoders/zrle.js
+++ b/core/decoders/zrle.js
@@ -1,0 +1,242 @@
+/*
+ * noVNC: HTML5 VNC client
+ * Copyright (C) 2021 The noVNC Authors
+ * Licensed under MPL 2.0 (see LICENSE.txt)
+ *
+ * See README.md for usage and integration instructions.
+ *
+ */
+import Inflator from "../inflator.js";
+
+export default class ZRLEDecoder {
+    constructor() {
+        this._data = false;
+        this._compressedLength = null;
+        this._uncompressed = null;
+        this._tileBuffer = new Uint8ClampedArray(64 * 64 * 4);
+        this._zlib = new Inflator();
+        this._clearDataBuffer();
+    }
+
+    _clearDataBuffer() {
+        this._dataBuffer = null;
+        this._dataBufferPtr = 0;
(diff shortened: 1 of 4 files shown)


rosmo:
Here's the version with 16-bit support (activated only for `reMarkable rfb`): https://github.com/rosmo/noVNC/tree/zrle-decoder-16bit

samhed:
Nice! Did you check the other PR we had on the topic? #1283 Perhaps there's something useful there.

What's your use case for this encoding? Did you do any profiling or comparisons to existing encodings?

core/decoders/zrle.js
```diff
@@ -0,0 +1,246 @@
+/*
+ * noVNC: HTML5 VNC client
+ * Copyright (C) 2019 The noVNC Authors
+ * (c) 2012 Michael Tinglof, Joe Balaz, Les Piech (Mercuri.ca)
```
samhed, line 4: This doesn't look correct. I guess you copied this header from some other file?

rosmo: I did, and I fixed it too now :)

rosmo:
I did find about the other PR after I had finished this. I had a look at it and it seemed slightly unfinished. My use case is that I want to use noVNC with reMarkable's (a eInk note taking tablet) Screen Sharing feature, which uses ZRLE and 16-bit colorspace. I didn't really benchmark that much - that's why I added the encoding between hextile and tight (I assume tight is the best and ZRLE is better than hextile). 

I fixed all the eslint warnings now though.

CendioOssman:
I've merged #1589 now, so I'll go ahead and close this PR. If you have further improvements you'd like to make on the merged code then please go ahead and open a new PR.

Issue #545, opened 2015-10-20
ZRLE support


jas8522:
I believe this may be related to the issue here: https://github.com/kanaka/noVNC/issues/528

The Repeater appears to be connecting perfectly, but when it attempts to negotiate with the server (through the repeater), the server indicates that it's using protocol version 003.016. Just like the user in the link above, I added 003.016 (although they were using 003.010) to the supported protocols and received:

Disconnected: unsupported encoding 16

Is there still no information on the type of encoding required to make this protocol version work?

According to Wikipedia encoding type 16 is File Transfer Protocol: https://en.wikipedia.org/wiki/RFB_protocol

jas8522:
Actually it seems the encoding may be ZRLE. I was interpreting 16 incorrectly. 16 corresponds to 0x00000010 hex, which according to the Wikipedia article should be ZRLE

DirectXMan12:
Well, we have a fairly performant zlib library now, so we should be able to support ZRLE.  I probably will not have time to get to it anytime soon, but we'd certainly be willing to accept PRs.

jas8522:
What do you think the chances are that the existing zlib library and this GNU rle encode/decode code can be paired up to implement ZRLE encoding in noVNC? http://www.marklomas.net/ch-egg/articles/rlejs.htm

kanaka:
@jas8522 pretty high if you implement it. ;-) There isn't any chance that I will get to this any time in the reasonable future. I suspect the same is true for @DirectXMan12. However, encodings are independent of each other so there is a good chance we would review and merge it in if there is a clean implementation (follows current coding style, doesn't break other tests, etc)

jas8522:
Hey Kanaka,

I wasn't attempting to have it created for me with that request, but rather to ask if anyone familiar with how RFB encoding typically works could indicate the best way to integrate that RLE code with the existing zlib library in noVNC. In other words, do you think I'm better off trying to build my ZRLE implementation from scratch or attempt to integrate the existing RLE Javascript code found above?

Do you have any tips for debugging an encoding type? In other words, what's best to test whether a section is working or not? I'm familiar with Javascript, but not so much with working with image tiles or display pixels. 

I'm working with the RFB documentation that includes the details of ZRLE. Any tips for how best to interpret their docs? It seems to me like starting with the HEXTILE code's usage of the FBU then modifying as necessary may work well. Thoughts on that?

CendioOssman:
Fixed via #1589.

PR #1589, opened 2021-10-05
Added support for ZRLE encoding

pauldumais:
Added support for ZRLE encoding, work mostly based on work done by #1283, just had to fix up the ZRLE palette tile decoding.  Also used the same fix to the inflator that #1587 did as well.  Tested against TightVNC Server on Windows and iDRAC VNC server.


rosmo:
Wow, this is really weird. @pauldumais I think you took a few notes from my PR (`allowIncomplete` for the inflator)? I think while this PR might be more straightforward, mine probably uses quite a bit less memory (I was testing on PC with 4k screen and the compressed data was around 4MB in size which decompressed is probably a lot more - I implemented a buffer system to reduce the memory usage. I'm also looking at the palette RLE decompression in this - I'm not sure if it supports padding bits correctly?

pauldumais:
It just so happened that we worked on it at the same time, I was attempting to get vnc connections working to iDRAC 8 servers, which requires ZRLE as the only encoding type supported, I noticed the work done by @Ekstazi on #1283 and worked on that as a base.  I noticed that @rosmo had put in the same fix with the `allowIncomplete` so I used the same fix for that file.  I ran the #1587 PR to see if I could use it instead of my changes and it appears to be to leave some weird rendering tearing artifacts on the screen, and mine renders the screen with no artifacts.

See the attached screenshot images in comparison:

pr1589
pr1587

pauldumais:
> Ok I understand. Thanks for testing the code from #1587 in your use case. Ideally the merged code should work well in both cases.
> 
> Despite it not being required for the code to work, I think the new encoding should be added to `core/encodings.js`.
> 
> You think you could write some unit tests for this? That's something we'd like to see before moving on.

I pushed my changes with unit tests.

CendioOssman:
Thanks. If you squash these commits together I can go ahead and merge this.

Did you test implementing this without modifying `Inflator`? It should be possible by inflating as you go, which in some cases means just a single byte.

samhed (changes requested):
One of the new ZRLE unit tests are failing it seems.

samhed (approved):
Looks good, thanks!

samhed:
Do you think you could squash your changes into one commit?

pdlan:
It seems that encodings.encodingZRLE is missing.

CendioOssman:
Indeed. Apparently we don't have a proper test for that part. I'll have a look.

CendioOssman:
Fixed in 65d6357cdf37f6523bf04cccadc8188f4fee462e. Thanks for noticing!

Neustradamus:
It is linked to:
- https://github.com/novnc/noVNC/issues/25
- https://github.com/novnc/noVNC/issues/545

PR #1590, opened 2021-10-05
Added support for Apple Remote Desktop authentication

pauldumais:
Added the support to perform authentication using the Apple Remote Desktop authentication protocol.  This required me to add the dependancy on the forge js encryption library for mobPow, MD5 and AES encryption.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -32,6 +32,8 @@ import RREDecoder from "./decoders/rre.js";
 import HextileDecoder from "./decoders/hextile.js";
 import TightDecoder from "./decoders/tight.js";
 import TightPNGDecoder from "./decoders/tightpng.js";
+import {MD5} from "./util/md5.js";
+import {modPow} from "./util/bigint-mod-arith.js";
 
 // How many seconds to wait for a disconnect to finish
 const DISCONNECT_TIMEOUT = 3;
@@ -1242,13 +1244,13 @@ export default class RFB extends EventTargetMixin {
                 break;
             case "003.003":
             case "003.006":  // UltraVNC
-            case "003.889":  // Apple Remote Desktop
                 this._rfbVersion = 3.3;
                 break;
             case "003.007":
                 this._rfbVersion = 3.7;
                 break;
             case "003.008":
+            case "003.889":  // Apple Remote Desktop
             case "004.000":  // Intel AMT KVM
             case "004.001":  // RealVNC 4.6
             case "005.000":  // RealVNC 5.3
@@ -1304,6 +1306,8 @@ export default class RFB extends EventTargetMixin {
                 this._rfbAuthScheme = 16; // Tight
             } else if (types.includes(2)) {
                 this._rfbAuthScheme = 2; // VNC Auth
+            } else if (types.includes(30)) {
+                this._rfbAuthScheme = 30; // ARD Auth
             } else if (types.includes(19)) {
                 this._rfbAuthScheme = 19; // VeNCrypt Auth
             } else {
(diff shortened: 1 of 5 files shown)


CendioOssman:
There is the crypto api available in modern browsers. Can't we use that and avoid having to ship extra things?

core/rfb.js
```diff
@@ -1496,6 +1499,76 @@ export default class RFB extends EventTargetMixin {
...
+        let sharedKey = serverPubKey.modPow(clientPrivKey, prime);
+
+        let md5Shared = forge.md.md5.create().update(forge.util.hexToBytes(sharedKey.toString(16))).digest();
+
+        let userPassword = this._rfbCredentials.username + '\0'
+                + forge.util.createBuffer(forge.random.generate(32)).toHex().substring(0, 63-this._rfbCredentials.username.length) +
+                this._rfbCredentials.password + '\0'
+                + forge.util.createBuffer(forge.random.generate(32)).toHex().substring(0, 63-this._rfbCredentials.password.length);
+
+        let packedUserPassword = forge.util.encodeUtf8(userPassword);
```
CendioOssman, line 1550: This seems odd. Are you sure this shouldn't be done to the fields before they are concatenated? Right now the total structure will have varying length, and you never send the length to the server.

CendioOssman (changes requested):

core/rfb.js
```diff
@@ -1496,6 +1499,76 @@ export default class RFB extends EventTargetMixin {
...
+        cipher.mode.pad = false;
+        cipher.start();
+        cipher.update(forge.util.createBuffer(packedUserPassword));
+        cipher.finish();
+
+        let data = this._stringToAsciiByteArray(cipher.output.data);
+        this._sock.send(data);
+
+        let pubBytes = clientPubKey.toByteArray();
+        if (pubBytes.length==129) {
```
CendioOssman, line 1563: This seems odd to hard code. Doesn't it depend on the key length?

CendioOssman (changes requested):

pauldumais:
> There is the crypto api available in modern browsers. Can't we use that and avoid having to ship extra things?

The Apple Remote Desktop authentication mechanism needs the follow cryptographic functions:
1) Support for BigInteger (it looks like web crypto has this)
2) Support for modpow function on BigInteger, I did not find this in web crypto
3) MD5 hashing, web crypto does not support this
4) AES-ECB encryption, web crypto has some AES but not ECB from what I can tell

core/rfb.js
pauldumais: The packed user password is always, 128 bytes in length, but you are right if someone who has a user or password that is larger than 63 characters this would not work, I can change the code to handle that case.

core/rfb.js
pauldumais: > This seems odd to hard code. Doesn't it depend on the key length?

For some reason sometimes the client pub key comes out at 129 bytes with an extra 0 at the beginning, i'm not sure why it's happening but this fixes that case and makes it work all the time.

core/rfb.js
CendioOssman: My point was that encoding it to UTF8 **expands** the data. A 128 character string might end up being 140 bytes once you've encoded it. You should do the encoding as the **first** step, not the last.

core/rfb.js
CendioOssman: Right, that is common. However specifically `129` seems wrong. `clientPubKey` is `keyLength` bytes large, not always 128. I'm assuming in your test `keyLength` happens to be 128 bytes?

CendioOssman:
> * Support for BigInteger (it looks like web crypto has this)
> 

BigInt is a language feature that seems to be present on all modern browsers. It doesn't seem to have the nice conversion functions that forge has, but most things are there.

> * Support for modpow function on BigInteger, I did not find this in web crypto
> 

They don't seem to expose anything for this, no. But from what I can gather a modpow function is just a handful of lines.

> * MD5 hashing, web crypto does not support this
> 

Annoying. They apparently decided to explicitly drop older stuff. It may not be secure these days, but it is still common in older protocols. :/

> * AES-ECB encryption, web crypto has some AES but not ECB from what I can tell

ECB is the brain dead chaining so it doesn't have an explicit mode. It's just doing AES without modification repeatedly.

forge is rather large, so I think it would be nice if we could do something simpler since we only seem to need a few percent of it. It looks like what we need is some MD5 code, a modpow function, and some conversion to byte arrays. That should be possible in way less than the 1 MiB that forge takes up. This is a load time that's enforced on all users after all.

pauldumais:
> forge is rather large, so I think it would be nice if we could do something simpler since we only seem to need a few percent of it. It looks like what we need is some MD5 code, a modpow function, and some conversion to byte arrays. That should be possible in way less than the 1 MiB that forge takes up. This is a load time that's enforced on all users after all.

Yes I agree a solution that did not depend on forge would be great.  I will work on a solution now.

pdlan:
WebCrypto API only works on secure contexts so you can’t use it without TLS. Also, its API uses asynchronous Promise and I have no idea how to combine it with noVNC.

CendioOssman:
The browser has taken the stand that new APIs will require TLS, so that will be a de facto requirement for the modern web. So I see it as natural that noVNC will require it.

pauldumais:
> The browser has taken the stand that new APIs will require TLS, so that will be a de facto requirement for the modern web. So I see it as natural that noVNC will require it.

I pushed the changes to support web crypto instead of forge.  I needed to add support for BigInt and modpow, so to do that I had to add the es2020 requirement.  Another option would be to use BigInteger.js instead but that would inflate the size a bit bigger.  I also had to perform the web crypto in an async block, but it appears to connect and authenticate properly.

CendioOssman:
BigInt seems to be well supported by all the major browsers, so I don't see an issue raising the requirements.

core/rfb.js
```diff
@@ -1496,6 +1500,99 @@ export default class RFB extends EventTargetMixin {
...
+        let password = encodeUtf8(this._rfbCredentials.password).substring(0, 63);
+
+        let paddedUsername = username + '\0' + bytesToHex(window.crypto.getRandomValues(new Uint8Array(34))).substring(0, 63);
+        let paddedPassword = password + '\0' + bytesToHex(window.crypto.getRandomValues(new Uint8Array(34))).substring(0, 63);
+
+        let credentials = paddedUsername.substring(0, 64) + paddedPassword.substring(0, 64);
+
+        let sharedKeyString = Array.from(hexToBytes(sharedKey.toString(16)), byte => String.fromCharCode(byte)).join('');
+
+        (async () => {
```
CendioOssman, line 1575: Using async functions is okay since it is also well supported by all current browsers.

However this is probably not the correct way to do so. Calling an async function will schedule it to complete at a later time. So you will be changing `this._rfbInitState` before you've completed the encryption and sent the data.

This happens to work anyway since `RFB` will go back to waiting for more data from the server, which it won't get since the server is waiting for data from the async function. But this is fragile, so we need to get this a bit more robust.

The proper solution would be to rewrite all protocol handling to be async. But that's a massive change, so not suitable here.

What you need to do is to make sure that the caller knows you are waiting for something. So you return `false` once you've kicked off the async function. You then need to set up a couple of things:

1. Some tracking that you are waiting for the async function, so if `_negotiateARDAuth()` gets called again you can throw an error.
2. A resolve handler that updates `this._rfbInitState` and calls `this._initMsg()`. Compare to what happens when we need to wait for credentials, and then later `sendCredentials()` gets called to resume things.

All of this should ideally also have unit tests to make sure nothing breaks in the future.

karma.conf.js
```diff
@@ -34,7 +34,7 @@ module.exports = (config) => {
       { pattern: 'app/localization.js', included: false, type: 'module' },
       { pattern: 'app/webutil.js', included: false, type: 'module' },
       { pattern: 'core/**/*.js', included: false, type: 'module' },
-      { pattern: 'vendor/pako/**/*.js', included: false, type: 'module' },
+      { pattern: 'vendor/**/*.js', included: false, type: 'module' },
```
CendioOssman, line 37: This change is no longer needed.

package.json
```diff
@@ -16,7 +16,7 @@
     "docs/LIBRARY.md",
     "docs/LICENSE*",
     "core",
-    "vendor/pako"
+    "vendor"
```
CendioOssman, line 19: This change is no longer needed.

tests/test.rfb.js
```diff
@@ -1170,6 +1170,56 @@ describe('Remote Frame Buffer Protocol Client', function () {
...
+                    // 2 bytes generator
+                    // 2 bytes keylength
+                    // prime number of keylength bytes
+                    // public key of keylength bytes
+                    const challenge = [127, 252, 0, 16];
+                    // prime & pubkey
+                    for (let i = 0; i < 32; i++) {
+                        challenge[i+4] = i;
+                    }
+                    client._sock._websocket._receiveData(new Uint8Array(challenge));
```
CendioOssman, line 1215: You should check that the correct response is calculated and sent as well.

CendioOssman (changes requested):
The async stuff needs a bit more things, but otherwise I think this looks great!

Please also squash the commits together once you have a final version.

CendioOssman:
I think the new changes look good. So just remove those bits that are no longer needed and I can go ahead and merge this.

CendioOssman:
@pauldumais, what's the status on this? :)

.eslintignore
```diff
@@ -1 +1 @@
-**/xtscancodes.js
+**/xtscancodes.js
```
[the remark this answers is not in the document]
CendioOssman: This should still be taken out.

pauldumais:
> @pauldumais, what's the status on this? :)

I reverted extra the change to the .eslintignore file.  Please let me know if there is anything else you need to merge this change.

CendioOssman:
Nice, thanks. Everything is merged now. :)

PR #1609, opened 2021-11-20
Add RealVNC's JPEG encoding

pdlan:
Add support for RealVNC's JPEG encoding.


CendioOssman:
Nice.

Do you think you could have a look at adding unit tests for this as well? To make sure it doesn't break in the future. RealVNC is not a server we test with regularly.

This encoding also lacks documentation over at @rfbproto. Do you think you could have a go at writing a PR there as well?

CendioOssman:
Nice. Thanks!

Could you squash things together to a single commit? Then I'll go ahead and merge this.

PR #1611, opened 2021-11-24
Document binding to localhost in Quick Start

pykgi6:
Not desirable to bind to public Internet unless necessary, especially for unprotected VNC.


CendioOssman:
Seems reasonable. Thanks!

PR #1615, opened 2021-11-30
Rename use_require.js to js2common.js

samhed:
This script only has one purpose now, let the name reflect that.


CendioOssman:
I'm not going to object to the overall principle, but "js2common" sounds like the output is no longer javascript. :)

Something generic like "convert.js"?

samhed:
Sure :+1:

PR #1617, opened 2021-12-10
Fix issue with ResizeObservers and scrollbars

### LINKED ISSUE #1616: Server-initiated resizes don't work in Chrome when "Remote resize" is enabled
**Describe the bug**
When `scaling mode` is set to `remote resize` and resizing the remote screen from the remote side, you get the correct size for a short moment, and then noVNC resizes back to fit your browser viewport.

This should not happen, we should only request a new size from the client-side when the browser window changes.

**To Reproduce**
1. Resize your browser window to a size smaller than 1920x1080
2. Connect using noVNC from Google Chrome
3. Resize the remote screen using `xrandr -s 1920x1080` from a terminal
4. See that the noVNC screen is resized to fit the browser window

**Expected behavior**
I expect to see scrollbars and that the remote screen should be larger than the browser window. This is what happens in Firefox.

 - OS: Fedora 34
 - Browser: Google Chrome
 - Browser version: 96

 - noVNC version: 78eda3c0403f82e10a6fe1bcc5f473a9af571474 (latest master)
 - VNC server: TigerVNC
 - WebSocket proxy: websockify

**Additional context**
This is a regression from when we switched to using a `ResizeObserver` instead of listening to the `resize` event. I can reproduce the issue using a9c2ff30b636cf5bb0c5f183302646a5b6ec8481 but not using fcb95821b76e105ed4a7ce4c2c19549c8dbbeb44.

samhed:
Also includes some minor fixups of the unit tests and the removal of an old unnecessary workaround for buggy scrollbars in Google Chrome.

Closes #1616

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -703,12 +703,13 @@ export default class RFB extends EventTargetMixin {
     }
 
     _fixScrollbars() {
-        // This is a hack because Chrome screws up the calculation
-        // for when scrollbars are needed. So to fix it we temporarily
-        // toggle them off and on.
+        // This is a hack because Safari on macOS screws up the calculation
+        // for when scrollbars are needed. We get scrollbars when making the
+        // browser smaller, despite remote resize being enabled. So to fix it
+        // we temporarily toggle them off and on.
         const orig = this._screen.style.overflow;
         this._screen.style.overflow = 'hidden';
-        // Force Chrome to recalculate the layout by asking for
+        // Force Safari to recalculate the layout by asking for
         // an element's dimensions
         this._screen.getBoundingClientRect();
         this._screen.style.overflow = orig;
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -244,6 +244,8 @@ export default class RFB extends EventTargetMixin {
         this._sock.on('message', this._handleMessage.bind(this));
         this._sock.on('error', this._socketError.bind(this));
 
+        this._expectedClientWidth = null;
+        this._expectedClientHeight = null;
         this._resizeObserver = new ResizeObserver(this._eventHandlers.handleResize);
 
         // All prepared, kick off the connection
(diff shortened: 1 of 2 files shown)


core/rfb.js
```diff
@@ -688,6 +688,12 @@ export default class RFB extends EventTargetMixin {
         }
 
         const size = this._screenSize();
+
+        // Don't request the same size we already have
+        if (this._fbWidth === size.w && this._fbHeight === size.h) {
+            return;
+        }
```
CendioOssman, line 722: This could in theory be race:y, but I'm not sure it will ever happen in practice.

If you've sent a request but haven't gotten a reply yet, then the values you are checking are stale.

E.g.

1. Start at 100×100
2. Send request to 200×200
3. Before you've gotten a reply, try to resize back to 100×100

This scenario will now break since no request will be sent in 3. The session will be stuck at 200×200 even though 100×100 was the latest proper size.

samhed:
With regards to the "chrome scrollbars"-workaround, it seems to still be needed on Safari 15.1 on macOS 12. Instead of removing it, I have updated the comment now.

core/rfb.js
samhed: Good point, I did not consider this. We could save the previous request values and check against those I guess, but I don't think it's worth the added complexity.

I'll remove this change, the gain from it is minimal anyway.

samhed:
I have tested this PR on ..

**Fedora 34:**
- Firefox 95
- Chrome 96

**Android 12:**
- Firefox 95
- Chrome 96
- Edge 96

**Windows 10:**
- Firefox 95
- Chrome 96
- Edge 96

**iPad OS 15.1:**
- Safari 15.1

**macOS 12:**
- Firefox 95
- Chrome 96
- Edge 96
- Safari 15.1

The things I tested was:

- [x] Making size smaller (server initiated)
- [x] Making size larger (server initiated)
- [x] Making size smaller by resizing the browser window
- [x] Making size larger by resizing the browser window
- [x] Scrollbars work when session is too big
- [x] Panning works when session is too big
- [x] Reconnect with a different browser size causes resize

PR #1619, opened 2021-12-21
make fallback error dismissable

flumm:
sometimes, an error is not that fatal, and the user may want
to dismiss it, so add a button to do exactly that

Signed-off-by: Dominik Csapak <d.csapak@proxmox.com>

### DIFF
--- a/app/error-handler.js
+++ b/app/error-handler.js
@@ -17,9 +17,11 @@ function handleError(event, err) {
             return false;
         }
 
-        // Only show the initial error
+        // Clear last error
         if (msg.hasChildNodes()) {
-            return false;
+            while (msg.firstChild) {
+                msg.removeChild(msg.firstChild);
+            }
         }
 
         let div = document.createElement("div");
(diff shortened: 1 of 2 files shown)


CendioOssman:
I don't think we want to do this. Such errors are often in a way that the internal state is corrupted. As such users will easily continue working if they can dismiss it and we'll get really bad bug reports when things break later in unexpected ways.

Compare this dialog with the crash dialog of an application. The system doesn't let you continue after that, and for good reason.

If you wish to take that risk when using noVNC in proxmox, then feel free to have a local modification. But for us I prefer if we get a clear report with the initial crash.

flumm:
Ok makes sense, i just want to explain further why i did this (maybe we can tackle this differently)

We currently use a 'webview' in our flutter android app to show novnc (for vm/ct/etc. console), and in this webview the
"requestFullscreen" call throws an exception (not entirely clear why, but we think it's a security decision) which would be shown
as such an error. We work around now differently (we hide the fullscreen button and make the webview itself fullscreen at all times now).

Aside from that, i only ever saw customer reports with errors when they had some browser extension that produced some js errors, and none from novnc itself (good job :) ).

How could we better handle the fullscreen error in another way?

CendioOssman:
We should catch that error and probably log something instead. Normal user interaction should never result in stray exceptions.

Ideally though we should detect that full screen won't work and remove the button automatically.

What error do you get?

flumm:
i get:

```
noVNC encountered an error:

fullscreen error
TypeError: fullscreen error
    at HTMLInputElement.toggleFullscreen
```

CendioOssman:
That wasn't terribly descriptive. :/

I don't suppose there is more details in some other log? E.g. the browser console?

CendioOssman:
I checked MDN, and I don't see any indication that there should be any details included unfortunately.

It does however not specify anything about an exception being thrown, so I'm guessing this is a Promise that is being rejected. So we need some handling for that.

Issue #294, opened 2013-08-30
Disable Directory Browsing?


DrDavy:
By Default it looks like the launch.sh allows you to see the root files and folders under the running directory.

I checked out the web.py but can't see anything to flip.
I also see the nova_novncproxy (which looks like it using apache pieces) but don't know if actually is used.

Thanks

gaborszita:
I can't find the include folder. I know images is under apps but where is the include directory and the vnc_auto?

samhed:
> I can't find the include folder. I know images is under apps but where is the include directory and the vnc_auto.html?

The include folder hasn't existed for many years. It got split into /core and /app, vnc_auto.html was renamed to vnc_lite.html

PR #1630, opened 2022-03-04
Possible Work Around for noVNC Directory Listing

schneider-san:
noVNC utilizes [Websockify](https://github.com/novnc/websockify/) which uses the python http.server module. This module loads a **[do_GET() -->>send_head()](https://docs.python.org/3.10/library/http.server.html#http.server.SimpleHTTPRequestHandler.do_GET)** method from **[SimpleHTTPRequestHandler](https://docs.python.org/3.10/library/http.server.html#http.server.SimpleHTTPRequestHandler)** class which eventually results in the **directory listing of all items in the current directory from which its original caller was run** ( if no index.html or index,htm file exists in that directory ). Otherwise it serves a http 404 response on listdirectory failure.

A possible but unclear mitigation is to run websockify directly #294. Another but not encouraged solution is **commenting the "**else**" block in **send_head(self)** method in python **http.server.SimpleHTTPRequestHandler** class** to force a http 404 response.

Code:
`def send_head(self):`
`....`
`....if not parts.path.endswith('/'):`
`....`
`....for index in "index.html", "index.htm":`
`....`
`....#else:`
`    #return self.list_directory(path)`

A much simpler and preferred solution would be to include an "index.html" file in the root folder "~/noVNC". This possible mitigation is an added feature to using noVNC as it gives leverage to users preferred landing pages.


CendioOssman:
Listing a directory or not is a decision we prefer to leave to the web server. So thank you for your idea, but I'm afraid it's not something we'll be adding at this time.

If you'd like to change this behaviour in websockify then feel free to add a feature request or PR in that repository.

PR #1643, opened 2022-04-03
Add support for ARD Security Type

pdlan:
[This PR](https://github.com/novnc/noVNC/pull/1590) seems not active anymore, so I wrote this alternative version. Similar to my previous PR of RA2ne, no external dependencies are needed. The handling of async function is also similar.

Tested against Chrome with ARD on macOS Big Sur.


CendioOssman:
Since #1590 was pretty much done let's see if we can get some attention there first.

CendioOssman:
#1590 is merged now, so this would be a duplicate. If you'd like to suggest further improvements to the merged code then please open a new PR with the changes you'd like to see.

PR #1644, opened 2022-04-05
Remove bigint-mod-arith.js

pdlan:
This PR removes the dependency of bigint-mod-arith.js and uses a simple implementation.


samhed:
This looks like a nice improvement, thanks!

However, I don't have the possibility of testing this myself. Perhaps @pauldumais could be a second opinion in this case?

pauldumais:
I just tested the change and it works great, nice optimization!

samhed:
Thanks!

PR #1660, opened 2022-06-22
Experimental h264

linlin4426:
where is main.js?  It seems no main.js in project.


CendioOssman:
I guess this was misfiled and seems to have been handled as martin19/noVNC#1.

PR #1672, opened 2022-08-08
Allow continued reconnect tries

VibroAxe:
Currently novnc will only retry once (assuming the server is unavailable) and then stop (as the detail from is unclean, usually "failed to connect"). Minor change will continue to reconnect every reconnect_delay seconds until either reconnected or user intervention cancels the attempt.

Enhancement to #747


CendioOssman (approved):
Sounds reasonable. And since we have a cancel button, there is a way for the user to break this now infinite loop.

Issue #209, opened 2012-10-24
Create screenshots


wmertens:
It would be really great if this code could be used to create a screenshot of a vnc screen. So connect to a session, get a full screenshot and then disconnect.

Would it be easy to implement something like that? I have a Node.JS project that manages VNC sessions on VMWare and I would like to show people a screenshot of their session.

There is the vncsnapshot C-based project but it's very hard to compile nowadays and it doesn't work with the VMWare VNC server, producing a grey screen.

PR #1676, opened 2022-08-15
Provide access to content of framebuffer

derfred:
I would like to access the content of the framebuffer programatically. I have added two helper functions.

Adresses #209

### DIFF
--- a/core/display.js
+++ b/core/display.js
@@ -224,6 +224,18 @@ export default class Display {
         this.viewportChangePos(0, 0);
     }
 
+    getImageData() {
+        return this._drawCtx.getImageData(0, 0, this.width, this.height);
+    }
+
+    toDataURL(type, encoderOptions) {
+        return this._backbuffer.toDataURL(type, encoderOptions);
+    }
+
+    toBlob(callback, type, quality) {
+        return this._backbuffer.toBlob(callback, type, quality);
+    }
+
     // Track what parts of the visible canvas that need updating
     _damage(x, y, w, h) {
         if (x < this._damageBounds.left) {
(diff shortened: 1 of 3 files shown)


core/display.js
```diff
@@ -224,6 +224,14 @@ export default class Display {
         this.viewportChangePos(0, 0);
     }
 
+    copyImageData() {
+        return this._targetCtx.getImageData(0, 0, this.width, this.height)
```
CendioOssman, line 228: Why `_targetCtx`? It seems more reasonable to get the uncropped, full display.

core/display.js
```diff
@@ -224,6 +224,14 @@ export default class Display {
...
+    }
+
+    toDataURL(format) {
```
CendioOssman, line 231: Let's mimic the API fully and also include `encoderOptions`.

On the same theme, let's keep the same argument names as the canvas API and rename this to `type`.

CendioOssman, line 231: The modern way is to use blobs, so a `toBlob()` would be nice as well.

CendioOssman (changes requested):
Seems reasonable. We try to mimic existing JavaScript APIs as much as possible though, so there as some tweaks I'd like.

Also, I'd like `docs/API.md` to be updated with these new methods.

core/display.js
derfred: I've switched the calls to use the `_drawCtx` and the `_backbuffer`

derfred:
@CendioOssman I've addressed your comments. Let me know what you think

CendioOssman (changes requested):
Looks great! Thanks for addressing all the comments!

Please squash these to a single commit and I can go ahead and merge things.

Also, try to add a little bit more context to the commit message. The summary in this PR is a very good example.

derfred:
@CendioOssman I've squashed the commits

CendioOssman (approved):

PR #1682, opened 2022-09-03
Add UTF8 support for clipboard

vrolife:
Convert clipboard content to UTF8 before serialization.


CendioOssman:
I'm afraid this would violate the VNC protocol and make noVNC incompatible with all conforming VNC servers.

For full Unicode support, please use a server that supports the extended clipboard format extension. E.g. TigerVNC or UltraVNC.

Issue #974, opened 2017-12-11
View only should show the remote cursor


agronick:
The Readme lists under features "Local or remote cursor". As far as I can tell this option was disabled and Remote cursors will be used whenever they are available. However, on view only connections the users see no cursor with buttons being clicked by an invisible cursor.

I was able to fix this by changing these two lines

In _negotiate_server_init:
```
this._display.disableLocalCursor();
```
to 
```
if (!this._viewOnly)
    this._display.disableLocalCursor();
```
In _sendEncodings
```
            if ((0, _browsers.browserSupportsCursorURIs)() && !_browsers.isTouchDevice && this._fb_depth == 24) {
                encs.push(_encodings.encodings.pseudoEncodingCursor);
            }
```
to
```
            if ((0, _browsers.browserSupportsCursorURIs)() && !_browsers.isTouchDevice && this._fb_depth == 24) {
                if (!this._viewOnly)
                    encs.push(_encodings.encodings.pseudoEncodingCursor);
            }
```

It would be nice if some of the options were restored such as the ability to have local and remote cursor. So helpdesk people can see exactly what is happening on the users screen and what they are clicking.

DirectXMan12:
Seems reasonable.

@samhed  @CendioOssman WDYT

samhed:
Agreed, sounds good. I'll commit.

> The Readme lists under features "Local or remote cursor". As far as I can tell this option was disabled and Remote cursors will be used whenever they are available.

This documentation issue is addressed in #971

samhed:
@CendioOssman asked me to hold off with committing, said he has an objection.

CendioOssman:
Yeah, I don't think this is the best way to solve this. This is a server bug IMO as it should not be sending a local cursor if it wants to control the position. Even if we change this it will still be broken if "view only" is turned off. So extra complexity for us without a proper fix for the fundamental problem.

What server are you using?

samhed:
> The easiest fix is to just add the option back in, IMO.

What do you mean?

CendioOssman:
> I'm not certain there's a good way for the sever to know when the user wants to control cursor position.

It's actually rather simple. The server can compare the current cursor position with the last requested by the client. If they differ then the server needs to render the cursor. This is implemented in the RealVNC code base, so every server based on that includes this behaviour.

> In some view-only scenarios (e.g. a presentation), the user doesn't care about cursor position...

In that scenario the user wants no cursor at all, not a server rendered one.


[...]

PR #1684, opened 2022-09-07
implement remote cursor rendering

derfred:
I have a use case where the client is watching the screen of a different user. In this scenario I would like to use the server rendering of the cursor, as the client has no input.

Addresses #974


CendioOssman (commented):
I'm still opposed to this, but if I'm overruled then I think this should be clearly presented as a workaround for buggy servers and not the normal way of getting a server rendered cursor.

fidoriel:
also, would need a workaround like this one. Any chance to merge this PR after two years?

Issue #1696, opened 2022-09-26
Wrong MacOS meta key (⌘) keysym


m1k1o:
When pressing `⌘` key on mac os, `XK_Alt_L` and `XK_Super_L` are sent instead of `XK_Meta_L` and `XK_Meta_R`.
![image](https://user-images.githubusercontent.com/7534274/192354707-10b9dca0-0e65-43b9-95d2-6f75668d27ef.png)

They are registered as `XK_Super_L` and `XK_Super_R` despite the name says `Meta`.
https://github.com/novnc/noVNC/blob/c101a31520aba083a85edcb2f3269a826471007a/core/input/domkeytable.js#L46

There is also placeholder for the `Super` keys, where I would say they should be registered.
https://github.com/novnc/noVNC/blob/c101a31520aba083a85edcb2f3269a826471007a/core/input/domkeytable.js#L53

And they are here converted from `XK_Super_L` and `XK_Super_R` to XK_Alt_L` and `XK_Super_L`.
https://github.com/novnc/noVNC/blob/c101a31520aba083a85edcb2f3269a826471007a/core/input/keyboard.js#L135-L140

What is the reason for not seding correct `XK_Meta_` keys?

**Expected behavior** is to send `XK_Meta_L` and `XK_Meta_R`.

**Client:**
 - OS: MacOS 12.5
 - Browser: Chrome 105, Firefox 104, Safari 15.6

**Server:**
 - noVNC version: from master c101a31520aba083a85edcb2f3269a826471007a, but the part of the code that handles this was not touched for 2 years.
 - VNC server / WebSocket proxy: n/a, because client sends (not or wrong) information.

**Additional context**
Guacamole client sends correct `XK_Meta_L` and `XK_Meta_R`, tested on their [test page](https://guacamole.apache.org/pub/tests/guac/keyboard-test.html).
![image](https://user-images.githubusercontent.com/7534274/192354168-d367d818-6de5-4cf0-9e08-370c272248aa.png)

[w3 docs](https://www.w3.org/TR/uievents-key/#keys-modifier) states, that Meta is used for `Apple Command or ⌘ key` and Windows key.

<details>
<summary> Testing on other platforms (click for details). </summary>

**Windows 11**
Testing on Windows reveals, it sends `XK_Super_L` and `XK_Super_R` as expected.
![image](https://user-images.githubusercontent.com/7534274/192357038-c8f90837-1349-4483-a32b-e009f2d63996.png)

[...]

PR #1698, opened 2022-09-26
MacOS meta key (⌘) with a different key has unreliable keyup event

### LINKED ISSUE #1695: MacOS meta key (⌘) with a different key has unreliable keyup event
When pressing any combination on MacOS with `⌘` key, keyup event for the second key is not sent.

**To Reproduce**

1. Press `⌘` key.
2. Press and release `V` key.
3. Release `⌘` key.
4. You notice that `V` key is repeatedly written on the screen because keyup event had never been sent.

https://user-images.githubusercontent.com/7534274/192346291-1f4156aa-1a38-44af-8b7f-69b576376719.mov

In logs you can see, that keyup event was not sent, because browser did not trigger that.
image

**Expected behavior**
Since I released the key I would expect that keyup event is sent.

**Client:**
 - OS: MacOS 12.5
 - Browser: Chrome 105, Firefox 104, Safari 15.6

**Server:**
 - noVNC version: from master c101a31520aba083a85edcb2f3269a826471007a, but the part of the code that handles this was not touched for 2 years.
 - VNC server / WebSocket proxy: n/a, because client sends (not or wrong) information.

**Additional context**
Guacamole client fixes exacly this problem [here](https://github.com/apache/guacamole-client/blob/165bd413c0ab9c55b0c0e3614fcb829ef5f8ca95/guacamole-common-js/src/main/webapp/modules/Keyboard.js#L286-L289) by marking those events as unreliable and sends keyup together with keydown:

m1k1o:
Fixes #1695 by auto releasing keys while meta is held down, similar as [apache guacamole's solution](https://github.com/apache/guacamole-client/blob/165bd413c0ab9c55b0c0e3614fcb829ef5f8ca95/guacamole-common-js/src/main/webapp/modules/Keyboard.js#L286-L289).

Must be merged after #1697 or editted based on the outcme of #1696.

### DIFF
--- a/core/input/keyboard.js
+++ b/core/input/keyboard.js
@@ -153,6 +153,16 @@ export default class Keyboard {
             keysym = this._keyDownList[code];
         }
 
+        // macOS doesn't send proper key releases if a key is pressed
+        // while meta is held down
+        if ((browser.isMac() || browser.isIOS()) &&
+            (e.metaKey && code !== 'MetaLeft' && code !== 'MetaRight')) {
+            this._sendKeyEvent(keysym, code, true);
+            this._sendKeyEvent(keysym, code, false);
+            stopEvent(e);
+            return;
+        }
+
         // macOS doesn't send proper key events for modifiers, only
         // state change events. That gets extra confusing for CapsLock
         // which toggles on each press, but not on release. So pretend
--- a/tests/test.keyboard.js
+++ b/tests/test.keyboard.js
@@ -197,6 +197,67 @@ describe('Key Event Handling', function () {
         });
     });
 
+    describe('Meta key combination on iOS and macOS', function () {
+        let origNavigator;
+        beforeEach(function () {
+            // window.navigator is a protected read-only property in many
+            // environments, so we need to redefine it whilst running these
+            // tests.
+            origNavigator = Object.getOwnPropertyDescriptor(window, "navigator");
+
+            Object.defineProperty(window, "navigator", {value: {}});
+            if (window.navigator.platform !== undefined) {
+                // Object.defineProperty() doesn't work properly in old
+                // versions of Chrome
+                this.skip();
+            }
+        });
+
+        afterEach(function () {
+            if (origNavigator !== undefined) {
+                Object.defineProperty(window, "navigator", origNavigator);
+            }
+        });
+
+        it('should send keyup when meta key is pressed on iOS', function () {
+            window.navigator.platform = "iPad";
+            const kbd = new Keyboard(document);
+            kbd.onkeyevent = sinon.spy();
+
+            kbd._handleKeyDown(keyevent('keydown', {code: 'MetaRight', key: 'Meta', location: 2, metaKey: true}));
+            expect(kbd.onkeyevent).to.have.been.calledOnce;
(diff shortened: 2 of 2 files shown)


core/input/keyboard.js
```diff
@@ -126,6 +126,15 @@ export default class Keyboard {
             return;
         }
 
+        // If a key is pressed while meta is held down, the keyup will
+        // never be sent in Chrome (bug #108404) and possibly others
+        if (e.metaKey && keysym !== KeyTable.XK_Meta_L && keysym !== KeyTable.XK_Meta_R) {
```
CendioOssman, line 131: This is a macOS specific bug, so we want to avoid punishing other platforms.

CendioOssman (changes requested):
Do you think you could look at unit tests for this as well, to make sure this doesn't break in the future? We already have tests for Caps Lock, so you should be able to do something similar.

core/input/keyboard.js
m1k1o: Don't know if it could also happen on IOS keyboard. But since the rewrite for both Super-key modifiers is guarded by `browser.isMac() || browser.isIOS()` I'll add it there.

m1k1o:
Moved to `browser.isMac() || browser.isIOS()`, fixed Meta -> Super and added unit tests.

core/input/keyboard.js
```diff
@@ -126,11 +126,20 @@ export default class Keyboard {
...
-        // possibly others).
         if (browser.isMac() || browser.isIOS()) {
+            // If a key is pressed while meta is held down, the keyup will
+            // never be sent in Chrome (bug #108404) and possibly others
+            if (e.metaKey && keysym !== KeyTable.XK_Super_L && keysym !== KeyTable.XK_Super_R) {
+                this._sendKeyEvent(keysym, code, true);
+                this._sendKeyEvent(keysym, code, false);
+                stopEvent(e);
+                return;
+            }
```
CendioOssman, line 137: I think it makes more sense to have this next to the other cases where we don't get a keyup event, further down.

CendioOssman (changes requested):
Looks good, thanks! Just a minor code move and we should be good.

Have you had a look at reporting this to the browsers? It would be nice to get the long-term fix rolling as well.

core/input/keyboard.js
m1k1o: It must be before shuffling the keys around, because later in code Super_L will be Alt_R and using that in the condition would be even more confusing, I'd say.

m1k1o: Or I could change the condition from `keysym` to `code === 'Meta'`. I try that.

m1k1o:
Moved code to the right place.

CendioOssman:
I think you need to look for `MetaLeft` and `MetaRight` instead. `Meta` is what's stored in `key`, not `code`. Otherwise, I think it looks fine.

CendioOssman (approved):
Great! Can you just squash the commits together, and then I can go ahead and merge this.

m1k1o:
Ready to merge.

CendioOssman:
Merged now. Thanks for your work!

PR #1706, opened 2022-10-20
Allow custom scale values

ax-jmckenzie:
I have a use case for allowing my end user to set a custom scale/zoom level.  Rather than providing only 'fit to window / auto-scale', or 'actual size' (i.e. clipping mode), I'd like to offer them the ability to scale at 1.5x, for example.

The scenario this would help address is that of multiple screens on the remote device. When there are two or more screens being shared simultaneously via VNC, a custom scale would allow the end user to fit just one of those screens in their viewport, by manually matching the height of their targeted remote screen to their own viewport height, via custom scaling.

When using fit-to-window/auto-scale, the multiple screens tend to appear much too small.  Switching to actual size _might_ work out just fine, but depending on the resolution of the remote screens, they may end up much too tall.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -287,6 +287,7 @@ export default class RFB extends EventTargetMixin {
         this._viewOnly = false;
         this._clipViewport = false;
         this._scaleViewport = false;
+        this._customScale = null;
         this._resizeSession = false;
 
         this._showDotCursor = false;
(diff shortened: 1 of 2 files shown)


CendioOssman:
We want to be very careful adding new APIs, so I prefer if we can do this in a more high level and general manner.

It sounds like you are trying to work around the fact that noVNC doesn't support multi-head. Is the server in this case one with proper multi-head support? E.g., TigerVNC. If so, then there is metadata available about the monitors that we might use to make more intelligent decisions to help the user. E.g., a choice of which monitor to show.

ax-jmckenzie:
Thank you for your response Pierre! In my case we are using the TightVNC server without multi-head. I certainly would love to get to a final state in the future where the user can select a monitor. In the interim I'm hoping to at least allow for custom scale. I should be able to achieve it outside of the NoVNC context if necessary but would have preferred to using the scaling that is already built-in. After realizing there was no public API to achieve it, I figured I'd propose adding it.

CendioOssman:
We really want to avoid removing APIs, so I'm afraid temporary solutions are generally forbidden.

I don't suppose switching to TigerVNC is an option?

I'm also thinking if we can find some inspiration from document or image viewers, and they scale options they provide. E.g. "best fit" and such. Perhaps we can have a mode that does a compromise between scaling and clipping for when the aspect ratio difference is too large...

Issue #1577, opened 2021-09-09
vnc.html crashes when running inside an <iframe> and "third-party cookies" are blocked


samhed:
**Describe the bug**
When blocking "third-party cookies" and trying to use vnc.html inside an <iframe> it will try to use `window.localStorage` to read or save settings. This causes a crash and an error message saying:

```
noVNC encountered an error:

Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
Error: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.
    at Module.readSetting (http://samuel:6080/app/webutil.js:159:9)
    at Object.initSetting (http://samuel:6080/app/ui.js:731:27)
    at Object.initSettings (http://samuel:6080/app/ui.js:150:12)
    at Object.start (http://samuel:6080/app/ui.js:59:12)
    at http://samuel:6080/app/ui.js:47:27
```

**To Reproduce**
1. Create a html page with an iframe pointing to your noVNC:
```
<html>
    <body>
        <iframe style="width: 1200px; height: 800px;" src="http://myhost:6080/vnc.html"></iframe>
    </body>
</html>
```
2. Start Incognito-mode in Google Chrome
3. Surf to your html page with the iframe
4. Try to connect
5. See error

**Expected behavior**
noVNC should not crash even if the user changes the default settings in their browser or runs noVNC inside an <iframe> in incognito mode.

**Screenshots**
![third-party-cookie-crash](https://user-images.githubusercontent.com/4117162/132701204-844a30f7-d016-41cc-a603-fdd7b4c67b0e.png)

**Client (please complete the following information):**
 - OS: Fedora 34
 - Browser: Google Chrome
 - Browser version: 92

**Server (please complete the following information):**
 - noVNC version: 1.3.0-beta
 - VNC server: TigerVNC
 - WebSocket proxy: websockify

**Additional context**
It's most important to not crash in this case, and perhaps log a warning saying we can't read or write settings to localStorage, but we should also investigate if we could work around this issue somehow and still be able to use localStorage.

duckduckstab1:
Maybe the beta's causing it?

samhed:
Same issue exists with v1.2.0, this is not a regression.

duckduckstab1:
Then maybe try 1.1.9?

samhed:
We have noticed this issue in normal use cases, that did not involve iframes, on two of our devices.

[...]

PR #1710, opened 2022-11-02
Don't crash if we can't use localStorage

samhed:
Our settings are not a fatal requirement, we can fall back on the default values if they can't be accessed. A scenario where we've seen this happen is when cookies are disabled in the browser. It seems `localStorage` is disabled along with cookies in these settings.

Fixes issue #1577.

### DIFF
--- a/app/webutil.js
+++ b/app/webutil.js
@@ -6,16 +6,16 @@
  * See README.md for usage and integration instructions.
  */
 
-import { initLogging as mainInitLogging } from '../core/util/logging.js';
+import * as Log from '../core/util/logging.js';
 
 // init log level reading the logging HTTP param
 export function initLogging(level) {
     "use strict";
     if (typeof level !== "undefined") {
-        mainInitLogging(level);
+        Log.initLogging(level);
     } else {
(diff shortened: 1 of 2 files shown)


app/webutil.js
```diff
@@ -146,7 +146,7 @@ export function writeSetting(name, value) {
     if (window.chrome && window.chrome.storage) {
         window.chrome.storage.sync.set(settings);
     } else {
-        localStorage.setItem(name, value);
+        localStorageWrapper("setItem", [name, value]);
```
CendioOssman, line 149: This became a bit meta. Calling methods by giving their names as a string looks very odd.

It also makes it more difficult to follow, as it is no longer apparent what happens here.

I think some duplication in the name of readability can be acceptable here. You could always add a `logOnce()` helper if you want to encapsulate that magic. Could even make it generic based on the message. ;)

farrukh-taqveem:
Can we merge this ?

samhed:
Yeah, I'll just rework it a bit according to @CendioOssman's comment

app/webutil.js
samhed: Good point, fixed now.

samhed:
Added some simple unit tests as well.

PR #1712, opened 2022-11-07
Fix unrecognized option "p" in busybox ps

nggit:
This will fix
`ps: unrecognized option: p`

on busybox ps (Alpine Linux).

I see that they already made a patch:
https://git.alpinelinux.org/aports/tree/testing/novnc/alpine-specific-launch.js.patch?id=5a37614f7bd6e97952c56e6d221b838fe46fbdaf

Changes in this PR will have both GNU/non-GNU compatibility.

Notes:
The "-e" option is not needed and will be ignored in busybox ps.
Hence it will not cause errors like the "-p" option.
https://github.com/mirror/busybox/blob/master/procps/ps.c#L585


CendioOssman:
Looks good. Thanks!

PR #1729, opened 2022-12-12
Add styling for checkboxes and range-sliders

samhed:
These commits also adds styling for radio-buttons and file choosers, but those are not used at the moment. They were added to make further cleanup in the CSS code possible.


app/styles/input.css
```diff
@@ -83,7 +83,16 @@ select {
...
+select:active {
+  /* Rotated arrow */
+  --select-arrow: url('data:image/svg+xml;utf8, \
+      <svg width="8" height="6" version="1.1" viewBox="0 0 8 6" \
+           xmlns="http://www.w3.org/2000/svg" transform="rotate(180)" > \
+          <path d="m6.5 1.5 -2.5 3 -2.5 -3 5 0" stroke-width="3" \
+                stroke="rgb(31,31,31)" fill="none" \
+                stroke-linecap="round" stroke-linejoin="round" /> \
+      </svg>');
+}
```
CendioOssman, line 73: This doesn't work in Firefox. :/

Not sure if it has a state for "open".

app/styles/input.css
```diff
@@ -96,6 +96,7 @@ input[type=checkbox] {
   width: 13px;
   height: 13px;
   vertical-align: bottom;
+  transition: 0.2s background-color linear;
```
CendioOssman, line 92: Other inputs? :)

samhed: Let's save that for later. The :hover, and selected states use `background-image: linear-gradient()` which seems tricky to animate:

https://stackoverflow.com/questions/6542212/use-css3-transitions-with-gradient-backgrounds/63848864#63848864

PR #1730, opened 2022-12-14
Replace JavaScript .noVNC_touch with CSS @media (any-pointer: coarse)

samhed:
Here's some more CSS cleanup. Many parts of the code were relying on the class `root.noVNC_touch`, which was set by the JavaScript. Firstly, this dependency isn't ideal, and secondly, it made many CSS selectors very complex.

Instead, we can use `@media (any-pointer: coarse)` to detect if there is a coarse (not accurate) pointing device available. This basically translates to; if there is a touch screen available. Our testing shows that this method is comparable.

Moreover, here's a 3rd party source where testing results for this media query can be found:

https://patrickhlauke.github.io/touch/pointer-hover-any-pointer-any-hover/results/

Note that this approach still isn't perfect. For example, on Linux, even if a touch screen is plugged in, we still get `False` on `any-pointer: coarse`. Also note that `any-pointer: coarse` isn't the perfect selector to decide if we need to show the on-screen keyboard button or not, but it's the best we got at the moment.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -319,7 +319,8 @@ const UI = {
     addConnectionControlHandlers() {
         document.getElementById("noVNC_disconnect_button")
             .addEventListener('click', UI.disconnect);
-        document.getElementById("noVNC_connect_button")
+        document.getElementById("noVNC_connect_box")
+            .getElementsByTagName('button')[0]
             .addEventListener('click', UI.connect);
         document.getElementById("noVNC_cancel_reconnect_button")
             .addEventListener('click', UI.cancelReconnect);
--- a/app/ui.js
+++ b/app/ui.js
@@ -88,7 +88,6 @@ const UI = {
 
         // Adapt the interface for touch screen devices
         if (isTouchDevice) {
-            document.documentElement.classList.add("noVNC_touch");
             // Remove the address bar
             setTimeout(() => window.scrollTo(0, 1), 100);
         }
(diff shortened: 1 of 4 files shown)


app/ui.js
```diff
@@ -319,7 +319,8 @@ const UI = {
     addConnectionControlHandlers() {
         document.getElementById("noVNC_disconnect_button")
             .addEventListener('click', UI.disconnect);
-        document.getElementById("noVNC_connect_button")
+        document.getElementById("noVNC_connect_box")
+            .getElementsByTagName('button')[0]
```
CendioOssman, line 322: Finding specific elements should really be done based on id. Both for clarity and robustness.

vnc.html
```diff
@@ -274,9 +274,9 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br>VNC</h1>
     <div class="noVNC_center">
         <div id="noVNC_connect_dlg">
             <div class="noVNC_logo" translate="no"><span>no</span>VNC</div>
-            <div id="noVNC_connect_button"><div>
+            <div id="noVNC_connect_box"><button>
```
CendioOssman, line 277: Is the outer div even needed?

app/ui.js
samhed: Fixed in 05baf14256a4640e15cda63cc2ac576f44482775

vnc.html
samhed: No, fixed in 05baf14256a4640e15cda63cc2ac576f44482775

PR #1732, opened 2022-12-23
Favicon cleanups and fixes

samhed:
The main changes here are:

 * We now use an ICO file for the favicons. This works better in almost all browsers and displays the proper icon in most places.
 * We have a different set of apple-touch-icons. A few old sizes were removed, and some new sizes were added according to Apple's documentation.

Note that when generating new icons with the Makefile, you can't use the ImageMagick packaged by Fedora. It's buggy:
https://bugzilla.redhat.com/show_bug.cgi?id=2140018


app/images/icons/Makefile
```diff
@@ -1,5 +1,5 @@
 BROWSER_SIZES := 16 24 32 48 64
-ANDROID_SIZES := 48 72 96 144 192
+ANDROID_SIZES := 72 96 144 192
```
CendioOssman: Does Android want 48×48, though? I think it would be best to encode this fact here and deal with the duplication elsewhere. Otherwise, this will look like 48×48 isn't used on Android and might get removed.

PR #1733, opened 2022-12-25
Refactor hasScrollbarGutter into a function

Ron-Lavi:
the code to determine hasScrollbarGutter can run while the document.body is not loaded yet. setting it as a function will trigger it when the actual UI is loaded, and document.body should exist.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1335,7 +1335,7 @@ const UI = {
 
         let brokenScrollbars = false;
 
-        if (!hasScrollbarGutter) {
+        if (!hasScrollbarGutter()) {
             if (isIOS() || isAndroid() || isMac() || isChromeOS()) {
                 brokenScrollbars = true;
             }
--- a/core/util/browser.js
+++ b/core/util/browser.js
@@ -45,30 +45,32 @@ try {
 
 export const supportsCursorURIs = _supportsCursorURIs;
 
-let _hasScrollbarGutter = true;
-try {
-    // Create invisible container
-    const container = document.createElement('div');
-    container.style.visibility = 'hidden';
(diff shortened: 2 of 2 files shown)


Ron-Lavi:
@samhed could you take a look? :)

CendioOssman:
I'm not sure about this change, as this might have a lot of overhead (since it does layout stuff).

In which scenario have you seen this running early enough to crash?

Ron-Lavi:
Hey @CendioOssman! when we import the library into our project, we always see the error in the browser console,
that `document.body` is undefined.
I think it's because we load the scripts in the page `head` and immediately the scrollbar test here is starting to run while the body wasn't finished to load yet.

CendioOssman:
We do the same here, so there seems to be something more to it.

Are you mangling the code in some way? E.g. converting it from modules to classical JavaScript? That would affect how the files are loaded.

Ron-Lavi:
yeah, I think that's what happens,
basically due to our JS architecture with plugins, IIRC we first bundle them together in:
https://github.com/theforeman/foreman-js/blob/master/packages/vendor-core/lib/modules.js#L117

and then using it in:
https://github.com/theforeman/foreman/blob/develop/webpack/assets/javascripts/bundle_novnc.js#L4

which is later used as vanilla JS in https://github.com/theforeman/foreman/blob/develop/app/views/hosts/console/vnc.html.erb

Ron-Lavi:
@CendioOssman rebased :)

Ron-Lavi:
Closing as it doesn't seem to be resolved here,
and on our side we indeed solved it by passing the script to the html body.

PR #1735, opened 2022-12-27
Handle overlay scrollbars better for dragging/clipping

### LINKED ISSUE #1678: Clipping forced on for desktop Firefox
**Describe the bug**
When using Firefox, clipping is forced on, and you have to use the drag panning move instead of using scroll bars when the session is larger than the browser.

**To Reproduce**
1. Open settings

**Expected behavior**
It should be possible to change "Clip to Window".

 - OS: Fedora 35
 - Browser: Firefox
 - Browser version: 102

 - noVNC version: 832937292ef09ede7da69923398ba79e69ca278d
 - VNC server: TigerVNC
 - WebSocket proxy: websockify

**Additional context**
This is likely fallout from #1341.

It also seems that this is new behaviour for Firefox 100, where scroll bars now hide when not used. There is a new setting for Firefox to get the old behaviour back. If you enable this and reload, then clipping is no longer forced.

CendioOssman:
This is another round of adjusting the automatic logic for when we need to force clipping and dragging. Previous attempts have been made in 0342e4f4891a0c29f10b25a53258bf40890bc9fd, 11715f20927f0d6667231f8f4c072737b5fab339, ef64917a9091a9a12e8245099d6b0b0fbac2f6e5 and e52a278ed7aca2f2c0831e3ceabefcdb39581620.

Firefox has now started doing useful overlay scrollbars, which means we can no longer assume that all overlay scrollbars are crap. Instead, fall back to an earlier approach of a black list.

Also note that the method here breaks on mobile platforms when they request a "desktop version" of a page. The reason being that they fake being some other platform. E.g. Android browsers will pretend to be regular Linux.

Long term, we should probably try to get away from the dichotomy of dragging versus scrolling. Many devices have both touch and mouse input, so the user might want to switch.

Fixes #1678.


CendioOssman:
This PR also adds some disable/enable functionality to the drag button. Many users can be confused about what the button does when activating it for a session that isn't being clipped. So now we simply disable it when it isn't useful.

PR #1739, opened 2022-12-30
Cleanup for cryptographic algorithms

pdlan:
noVNC has introduced many cryptographic algorithms including AES, DES, SHA1, MD5 and RSA for multiple security types. However, the code for these algorithms are distributed in different files. Also, some code, such as the one for the conversion between BigInt and Uint8Array, modPow and AES block cipher with AES-CBC, is duplicated. This PR does the following things with two commits:
The first commit 1. puts all code for cryptographic algorithms to a new directory `core/crypto` ; 2. adds wrappers for the algorithms including SHA1, AES and RSA keygen to prevent directly using the SubtleCrypto API.

The second commit adds the fallback implementations for AES, SHA1 and RSA keygen when SubtleCrypto is not available, which is the case for insecure contexts. With these changes, RSA-AES, ARD should be available without HTTPS.

I am not sure whether it is a good idea to introduce the fallback implementations though. If you think it is unnecessary, I may remove the second commit.

### DIFF
--- a/core/crypto/bigint.js
+++ b/core/crypto/bigint.js
@@ -1,0 +1,34 @@
+export function modPow(b, e, m) {
+    let r = 1n;
+    b = b % m;
+    while (e > 0n) {
+        if ((e & 1n) === 1n) {
+            r = (r * b) % m;
+        }
+        e = e >> 1n;
+        b = (b * b) % m;
+    }
+    return r;
+}
+
+export function bigIntToU8Array(bigint, padLength=0) {
+    let hex = bigint.toString(16);
+    if (padLength === 0) {
+        padLength = Math.ceil(hex.length / 2);
+    }
+    hex = hex.padStart(padLength * 2, '0');
+    const length = hex.length / 2;
+    const arr = new Uint8Array(length);
+    for (let i = 0; i < length; i++) {
+        arr[i] = parseInt(hex.slice(i * 2, i * 2 + 2), 16);
+    }
+    return arr;
+}
+
+export function u8ArrayToBigInt(arr) {
+    let hex = '0x';
+    for (let i = 0; i < arr.length; i++) {
+        hex += arr[i].toString(16).padStart(2, '0');
+    }
+    return BigInt(hex);
+}
--- a/core/crypto/crypto.js
+++ b/core/crypto/crypto.js
@@ -1,0 +1,90 @@
+import { AESECBCipher, AESEAXCipher } from "./aes.js";
+import { DESCBCCipher, DESECBCipher } from "./des.js";
+import { RSACipher } from "./rsa.js";
+import { DHCipher } from "./dh.js";
+import { MD5 } from "./md5.js";
+
+// A single interface for the cryptographic algorithms not supported by SubtleCrypto.
+// Both synchronous and asynchronous implmentations are allowed.
+class LegacyCrypto {
+    constructor() {
+        this._algorithms = {
+            "AES-ECB": AESECBCipher,
+            "AES-EAX": AESEAXCipher,
+            "DES-ECB": DESECBCipher,
+            "DES-CBC": DESCBCCipher,
+            "RSA-PKCS1-v1_5": RSACipher,
+            "DH": DHCipher,
+            "MD5": MD5,
+        };
+    }
+
+    encrypt(algorithm, key, data) {
+        if (key.algorithm.name !== algorithm.name) {
+            throw new Error("algorithm does not match");
+        }
+        if (typeof key.encrypt !== "function") {
+            throw new Error("key does not support encryption");
+        }
+        return key.encrypt(algorithm, data);
+    }
+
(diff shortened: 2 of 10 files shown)


CendioOssman (changes requested):
I'm afraid I think most of this is the wrong direction. SubtleCrypto is how crypto is supposed to be handled on the web, so we want to avoid obfuscating that by having wrappers or otherwise do things differently.

The browsers have also been very clear that unencrypted HTTP is deprecated, so missing SubtleCrypto is just the first of many features we will need. That means I don't see any point in trying to provide fallbacks for that scenario. TLS is simply a base requirement for any modern web application.

What we do still have a need for are implementations for things that SubtleCrypto doesn't have. Like DES, MD5 or raw RSA. We can always look at how to clean those up.

Ideally, though, they should behave similar to how SubtleCrypto does things, so the code looks similar, regardless of where the crypto implementation happens to be.

I'm afraid that means a no to most of the suggestions in this PR. If you would like to clean up the parts that SubtleCrypto doesn't provide, then please update the PR, and we can have a close look.

core/crypto/crypto.js
```diff
@@ -0,0 +1,77 @@
...
+
+    importKey(format, keyData, algorithm, extractable, keyUsages) {
+        if (format !== "raw") {
+            throw new Error("key format is not supported");
+        }
+        const alg = this._algorithms[algorithm.name];
+        if (typeof alg === "undefined") {
+            throw new Error("algorithm is not supported");
+        }
+        return alg.importKey(keyData, algorithm, extractable, keyUsages);
```
CendioOssman, line 43: We probably don't need it yet, but ideally the keys should mimic `CryptoKey` objects.

core/crypto/bigint.js
```diff
@@ -0,0 +1,217 @@
...
+
+export function u8ArrayToBigInt(arr) {
+    let hex = '0x';
+    for (let i = 0; i < arr.length; i++) {
+        hex += arr[i].toString(16).padStart(2, '0');
+    }
+    return BigInt(hex);
+}
+
+export class RSACipher {
```
CendioOssman, line 38: Isn't it a bit confusing that the RSA and DH algorithms are hidden here? Why not give them their own files?

core/crypto/crypto.js
```diff
@@ -0,0 +1,77 @@
...
+    constructor() {
+        this._algorithms = {
+            "AES-ECB": AESECBCipher,
+            "AES-EAX": AESEAXCipher,
+            "DES-ECB": DESECBCipher,
+            "DES-CBC": DESCBCCipher,
+            "RSA-PKCS1-v1_5": RSACipher,
+            "DH": DHCipher,
+            "MD5": MD5,
+        };
```
CendioOssman, line 18: Shouldn't this be split based on the type of algorithm? Or the check that a proper algorithm is specified might miss that it won't actually work.

core/crypto/des.js
```diff
@@ -258,9 +258,74 @@ export default class DES {
         }
         return b;
     }
+}
+
+export class DESECBCipher {
```
CendioOssman, line 263: We should probably stop exporting the old interface so that we use the new one everywhere.

tests/test.rfb.js
```diff
@@ -1298,35 +1312,30 @@ describe('Remote Frame Buffer Protocol Client', function () {
...
-                        243, 81, 112, 182, 64, 62, 149, 7, 151, 28, 36, 161, 247, 247, 36, 96,
-                        230, 95, 58, 207, 46, 183, 100, 139, 143, 155, 224, 43, 219, 3, 71, 139]);
+                        199, 39, 204, 95, 190, 70, 127, 66, 5, 106, 153, 228, 123, 236, 150, 206,
+                        62, 107, 11, 4, 21, 242, 92, 184, 9, 81, 35, 125, 56, 167, 1, 215,
+                        182, 145, 183, 75, 245, 197, 47, 19, 122, 94, 64, 76, 77, 163, 222, 143,
+                        186, 174, 84, 39, 244, 179, 227, 114, 83, 231, 42, 106, 205, 43, 159, 110,
+                        209, 240, 157, 246, 237, 206, 134, 153, 195, 112, 92, 60, 28, 234, 91, 66,
+                        131, 38, 187, 195, 110, 167, 212, 241, 32, 250, 212, 213, 202, 89, 180, 21,
+                        71, 217, 209, 81, 42, 61, 118, 248, 65, 123, 98, 78, 139, 111, 202, 137,
+                        50, 185, 37, 173, 58, 99, 187, 53, 42, 125, 13, 165, 232, 163, 151, 42, 0]);
```
CendioOssman, line 1338: How come this needed to change? The encryption algorithms should still be the same, right?

CendioOssman (changes requested):
Nice! Great work!

There are some minor questions, but I think this is pretty much ready.

CendioOssman (approved):

PR #1740, opened 2022-12-31
Fix unescaped optargs in websockify command

tutacat:
When passing a path with spaces to websockify, it is incorrectly read as multiple arguments, because the variables are not enclosed in quotes.


tutacat (commented):
All args need fixing

PR #1741, opened 2022-12-31
Fix unescaped arguments in websockify command

tutacat:
When passing a path with spaces to websockify, it is incorrectly read as multiple arguments, because the args weren't enclosed in quotes.


utils/novnc_proxy
```diff
@@ -93,14 +93,14 @@ while [ "$*" ]; do
     --web)     WEB="${OPTARG}"; shift            ;;
     --ssl-only) SSLONLY="--ssl-only"             ;;
     --file-only) FILEONLY_ARG="--file-only"      ;;
-    --record) RECORD_ARG="--record ${OPTARG}"; shift ;;
-    --syslog) SYSLOG_ARG="--syslog ${OPTARG}"; shift ;;
+    --record) RECORD_ARG='--record "${OPTARG}"'; shift ;;
+    --syslog) SYSLOG_ARG='--syslog "${OPTARG}"'; shift ;;
```
CendioOssman, line 97: I'm afraid `'` disables variable expansion, so this won't work.

CendioOssman (changes requested):
Thank you for your fix!

I'm afraid it isn't entirely correct, though. Please have a look at the comment.

samhed:
@tutacat do you think you will get a chance to fix this?

CendioOssman:
This looks better, but unfortunately, it still doesn't work. It is still splitting the words for the record, syslog and auth arguments.

CendioOssman:
No response. Closing.

tutacat:
Yes. I am a "volounteer". Not 100% of my time is github. Not all minds work the same, either.

CendioOssman:
I understand, and we do try to give some time before we consider something abanonded.

I can gladly reopen it if it's not abandoned and you'd like to resume the work?

tutacat:
hiya sorry about this pull, it was easier to make [a new one](1891)

PR #1745, opened 2023-01-14

### LINKED ISSUE #1737: noVNC `/websockify` endpoint fails when web server is served with a path prefix
**Describe the bug**
When noVNC is served with a path prefix, say `http://localhost:5900/pathprefix/vnc.html`, the requests to websockets fail as URL is built without considering the path prefix. In my case, I am trying to integrate noVNC within JupyterLab environment using [jupyter server proxy](https://github.com/jupyterhub/jupyter-server-proxy). When running on localhost, the integrated noVNC server runs at root `http://localhost:8888/novnc_0/`. noVNC is able to fetch all js related files by taking the `/novnc_0/` prefix into account. But the endpoint `/websockify` does not take the path prefix into account and hence fails. I am attaching screenshot of browser console.

Looking at screenshot, the request that failed should be made to  `http://localhost:8888/novnc_0/websockify` instead of `http://localhost:8888/websockify`. 

**To Reproduce**

A simple reproducer can be serving noVNC via a nginx reverse proxy at the path prefix.

**Expected behavior**
noVNC taking base URL into account while building URL for `websockify` endpoint.

**Screenshots**
websockify

 - OS:  Ubuntu 20
 - Browser: Firefox
 - Browser version: 108

 - noVNC version: `3553a451d8b9cf566232b7de2e764861c57a0e9a`
 - VNC server: TigerVNC
 - WebSocket proxy: websockify

**Additional context**

By taking base URL into account in `vnc.html` and `vnc_lite.html` should fix the problem.

```
url += window.location.pathname + path;
```

vsilvar:
Fixes #1737 and it should work for all of the different scenarios where the relative assets also work. This includes, for example:
   - /
   - /vnc.html
   - /vnc/
   - /vnc/vnc.html

It won't work for:
 - /vnc (but the assets don't work either, so it's consistent)
 - /vnc.html/ (but this results in a 404 by default anyway)

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -180,7 +180,7 @@ const UI = {
         UI.initSetting('shared', true);
         UI.initSetting('view_only', false);
         UI.initSetting('show_dot', false);
-        UI.initSetting('path', window.location.pathname.substr(1) + '/../websockify');
+        UI.initSetting('path', window.location.pathname.substring(1) + '/../websockify');
         UI.initSetting('repeaterID', '');
         UI.initSetting('reconnect', false);
         UI.initSetting('reconnect_delay', 5000);
--- a/app/ui.js
+++ b/app/ui.js
@@ -180,7 +180,7 @@ const UI = {
         UI.initSetting('shared', true);
         UI.initSetting('view_only', false);
         UI.initSetting('show_dot', false);
-        UI.initSetting('path', 'websockify');
+        UI.initSetting('path', window.location.pathname.substr(1) + '/../websockify');
         UI.initSetting('repeaterID', '');
         UI.initSetting('reconnect', false);
         UI.initSetting('reconnect_delay', 5000);


app/ui.js
```diff
@@ -180,7 +180,7 @@ const UI = {
         UI.initSetting('shared', true);
         UI.initSetting('view_only', false);
         UI.initSetting('show_dot', false);
-        UI.initSetting('path', 'websockify');
+        UI.initSetting('path', window.location.pathname.substr(1) + '/../websockify');
```
CendioOssman, line 183: This is a bit roundabout way of getting to where we want¹. I'd like to see something where we construct the sane, final path and hand off here.

¹ And `substr()` is also deprecated. :)

CendioOssman (changes requested):

app/ui.js
vsilvar: I've updated the usage of the deprecated `substr()` method.
And I agree, not sure if this is the best method, but the main idea is to get the ball rolling because this issue seems to keep being brought up.

CendioOssman:
There were some suggestions on bug #1058 that could be explored further. We don't want to hard-code any document names, though. The ideal is to mimic what the browser does when fetching linked resources.

app/ui.js
jginternational: Completely agree with @vsilvar 
We are facing this issue since years ago

jginternational:
In my solution, I need several servers behind an nginx
The urls are: 
www.server.com/view/user1
www.server.com/view/user2
...
The only way to get there is by adding the path as querystring
www.server.com/view/user1?path=/view/user1/websockify
www.server.com/view/user2?path=/view/user2/websockify

dafoo:
> In my solution, I need several servers behind an nginx The urls are: [www.server.com/view/user1](http://www.server.com/view/user1) [www.server.com/view/user2](http://www.server.com/view/user2) ... The only way to get there is by adding the path as querystring [www.server.com/view/user1?path=/view/user1/websockify](http://www.server.com/view/user1?path=/view/user1/websockify) [www.server.com/view/user2?path=/view/user2/websockify](http://www.server.com/view/user2?path=/view/user2/websockify)

A little hacky, but this worked. Thanks

DeepCowProductions:
Could this get merged at some point? 
A lot of custom application proxies don't support websocket rewrites. (also requires more configuration effort...)

For now, I am using a rather dirty fix for my docker images after the novnc installation (also works for kasmvnc):
` RUN sed -i "s/UI.initSetting('path', 'websockify');/UI.initSetting('path', window.location.pathname.substring(1) + '\/..\/websockify');/" /path/to/novnc/main.bundle.js`

PR #1768, opened 2023-03-27
Add english to list of supported languages

mathis-marcotte:
Adds "en" to the list of supported languages so that it can at least get picked up as a language in that "Second Pass: fallback" when translating. 
This is so that it doesn't ignore english in cases where a browser doesn't have "English" or "English-US" as their set language, but have something like "English-Canada" or "English-UK" as their primary language set in their browser.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1762,7 +1762,7 @@ const UI = {
 };
 
 // Set up translations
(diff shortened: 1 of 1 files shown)


samhed:
Looks good. I tested your change in chrome with the following languages configured:

1. English (Canada)
2. French

Without the fix I see French in the noVNC UI, with the fix I see English.

Thank you!

PR #1769, opened 2023-04-06
Fixes for fragment parameters

CendioOssman:
I don't think #1573 actually fixed anything, as we already had code for fragments, so this PR reverts that.

It also fixes up the history handling in the tests for that code, as we don't want permanent history changes after running tests.


vnc_lite.html
```diff
@@ -120,7 +120,7 @@
             // Note that we use location.href instead of location.search
             // because Firefox < 53 has a bug w.r.t location.search
             const re = new RegExp('.*[?&]' + name + '=([^&#]*)'),
-                  match = ''.concat(document.location.href, window.location.hash).match(re);
+                  match = document.location.href.match(re);
```
samhed, line 123: That regex doesn't match `#`. Since vnc_lite.html doesn't use webutil.js, specifying fragment parameters doesn't work. We should remove that part of the comment here.

samhed (changes requested):
Yeah, the changes merged in #1573 are confusing indeed. Looks good to remove that.

I also saw some small typos in the comments (can't comment on those lines since they aren't part of the normal context):

```
   // https://www.example.com#myqueryparam=myvalue&password=secreatvalue
                                                                 ^
```

PR #1771, opened 2023-04-06
Set _rfbVeNCryptState = 4 not == 4

lewayotte:
Just a small bug I noticed...

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1593,7 +1593,7 @@ export default class RFB extends EventTargetMixin {
                              this._rfbAuthScheme >> 8,
                              this._rfbAuthScheme]);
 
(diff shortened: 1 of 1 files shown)


samhed:
Nice find!

`_rfbVeNCryptState` number 4 doesn't seem to mean anything, however. The code seems to have been introduced by df8d005de960e2133ded0610656af591e92fc6b7 as part of #1677. As the author of that commit, any input here @CendioOssman? Should we simply remove that line, or is it there for readability?

CendioOssman:
It is partly there because it used to be in the old code. But it still serves a purpose as a sanity check. If something tries to continue the VeNCrypt handshake, then the state machine will be in a clearly invalid state and things will crash more clearly. So I think we should keep it.

PR #1780, opened 2023-05-30
Various fixes for unit tests

CendioOssman:
I was playing around with using `async`, which unfortunately failed. But I had to make a bunch of fixes to the tests, which are still valuable.


tests/test.websock.js
```diff
@@ -185,61 +185,48 @@ describe('Websock', function () {
...
+            sock.attach(websock);
+        });
+
+        describe('flush', function () {
             it('should actually send on the websocket', function () {
-                sock._websocket.bufferedAmount = 8;
-                sock._websocket.readyState = WebSocket.OPEN;
                 sock._sQ = new Uint8Array([1, 2, 3]);
                 sock._sQlen = 3;
                 const encoded = sock._encodeMessage();
```
samhed, line 204: From the variable & function names, it seems we're still looking at internal stuff here? If we can't avoid that, perhaps we can add a comment, like you did in other commits.

tests/test.rfb.js
```diff
@@ -297,77 +297,102 @@ describe('Remote Frame Buffer Protocol Client', function () {
 
         describe('#sendCtrlAlDel', function () {
             it('should sent ctrl[down]-alt[down]-del[down] then del[up]-alt[up]-ctrl[up]', function () {
```
samhed, line 299: typo "sent"

samhed (commented):
Very nice cleanup, good work!

tests/test.rre.js
```diff
@@ -93,7 +93,10 @@ describe('RRE Decoder', function () {
         display.fillRect(2, 0, 2, 2, [ 0x00, 0xff, 0x00 ]);
         display.fillRect(0, 2, 2, 2, [ 0x00, 0xff, 0x00 ]);
 
-        testDecodeRect(decoder, 1, 2, 0, 0, [ 0x00, 0xff, 0xff, 0xff, 0xff ], display, 24);
+        testDecodeRect(decoder, 1, 2, 0, 0,
+                       [ 0x00, 0x00, 0x00, 0x00,
+                         0xff, 0xff, 0xff, 0xff ],
+                       display, 24);
```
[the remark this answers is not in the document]
CendioOssman: It's really a flaw in the tests. The success condition is that nothing happens. So this test succeeded because the decoder wanted more data, not because it was done processing and found nothing to render.

I'll have a look if that distinction can be detected in the tests.

tests/test.websock.js
CendioOssman: Are you referring to `_sQ` and `_sQlen`? Those are unfortunately the current API. :/

I have another branch soon incoming which will clean those up and add a more clean API on the send side as well.

tests/test.rfb.js
CendioOssman: Existing bug. ;)

PR #1782, opened 2023-06-04
Websock API cleanup

CendioOssman:
More clean-up as a result of playing around with async functions. Follow-up to #1780.


samhed (commented):
Nice cleanup. Looks good!

PR #1786, opened 2023-07-05
Add option to hide status bar for vnc_lite.html

damnthem:


CendioOssman:
Thank you for your contribution, but vnc_lite.html is just meant to be a minimal example. I'm afraid that we only accept features for it that focus on showing important parts of how to use the noVNC API.

PR #1795, opened 2023-07-27
Add support for QEMU LED State pseudo encoding

otthou:
QEMU supports using a Pseudo Encoding to send keyboard LED status to the VNC client:
https://www.qemu.org/docs/master/interop/vnc-ledstate-pseudo-encoding.html

This complements the QEMU extended key event no-vnc already supports. When you have a VNC server like QEMU that uses the extended key events to send the scan codes of the pressed keys, instead of the letter that is typed, the remote state of the keyboards CapsLock, NumLock and ScrollLock matters, instead of the local state.

This pull request just exposes these events, if supported by the VNC server, via an event. This way, it is possible for an integrator of no-vnc to show the remote keyboard led states in the client.

### DIFF
--- a/core/input/keyboard.js
+++ b/core/input/keyboard.js
@@ -36,7 +36,7 @@ export default class Keyboard {
 
     // ===== PRIVATE METHODS =====
 
-    _sendKeyEvent(keysym, code, down) {
+    _sendKeyEvent(keysym, code, down, numlock = null, capslock = null) {
         if (down) {
             this._keyDownList[code] = keysym;
         } else {
@@ -48,8 +48,8 @@ export default class Keyboard {
         }
 
         Log.Debug("onkeyevent " + (down ? "down" : "up") +
-                  ", keysym: " + keysym, ", code: " + code);
-        this.onkeyevent(keysym, code, down);
+                ", keysym: " + keysym, ", code: " + code, + ", numlock: " + numlock + ", capslock: " + capslock);
+        this.onkeyevent(keysym, code, down, numlock, capslock);
     }
 
     _getKeyCode(e) {
@@ -86,6 +86,14 @@ export default class Keyboard {
     _handleKeyDown(e) {
         const code = this._getKeyCode(e);
         let keysym = KeyboardUtil.getKeysym(e);
+        let numlock = e.getModifierState('NumLock');
+        let capslock = e.getModifierState('CapsLock');
+
+        // getModifierState for NumLock is not supported on mac and ios and always returns false.
+        // Set to null to indicate unknown/unsupported instead.
+        if (browser.isMac() || browser.isIOS()) {
+            numlock = null;
+        }
 
         // Windows doesn't have a proper AltGr, but handles it using
         // fake Ctrl+Alt. However the remote end might not be Windows,
@@ -107,7 +115,7 @@ export default class Keyboard {
                 //        key to "AltGraph".
                 keysym = KeyTable.XK_ISO_Level3_Shift;
             } else {
-                this._sendKeyEvent(KeyTable.XK_Control_L, "ControlLeft", true);
+                this._sendKeyEvent(KeyTable.XK_Control_L, "ControlLeft", true, numlock, capslock);
             }
         }
 
@@ -118,8 +126,8 @@ export default class Keyboard {
                 // If it's a virtual keyboard then it should be
                 // sufficient to just send press and release right
                 // after each other
-                this._sendKeyEvent(keysym, code, true);
-                this._sendKeyEvent(keysym, code, false);
+                this._sendKeyEvent(keysym, code, true, numlock, capslock);
+                this._sendKeyEvent(keysym, code, false, numlock, capslock);
             }
 
             stopEvent(e);
@@ -157,8 +165,8 @@ export default class Keyboard {
         // while meta is held down
         if ((browser.isMac() || browser.isIOS()) &&
             (e.metaKey && code !== 'MetaLeft' && code !== 'MetaRight')) {
-            this._sendKeyEvent(keysym, code, true);
-            this._sendKeyEvent(keysym, code, false);
+            this._sendKeyEvent(keysym, code, true, numlock, capslock);
+            this._sendKeyEvent(keysym, code, false, numlock, capslock);
             stopEvent(e);
             return;
         }
@@ -168,8 +176,8 @@ export default class Keyboard {
         // which toggles on each press, but not on release. So pretend
         // it was a quick press and release of the button.
         if ((browser.isMac() || browser.isIOS()) && (code === 'CapsLock')) {
-            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', true);
-            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', false);
+            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', true, numlock, capslock);
+            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', false, numlock, capslock);
             stopEvent(e);
             return;
         }
@@ -182,8 +190,8 @@ export default class Keyboard {
                             KeyTable.XK_Hiragana,
                             KeyTable.XK_Romaji ];
         if (browser.isWindows() && jpBadKeys.includes(keysym)) {
-            this._sendKeyEvent(keysym, code, true);
-            this._sendKeyEvent(keysym, code, false);
+            this._sendKeyEvent(keysym, code, true, numlock, capslock);
+            this._sendKeyEvent(keysym, code, false, numlock, capslock);
             stopEvent(e);
             return;
         }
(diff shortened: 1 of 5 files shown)


CendioOssman:
Unfortunately, this extension is about more than just showing the state of those LEDs. The server also expects the client to be able to synchronize the state if it supports this extension. Something we cannot do, as the browsers will not let us change the state.

Are you seeing issues with QEMU right now? Because last I checked, it is supposed to try to do this synchronisation on behalf of the client if the LED extension is missing. So things are likely to get worse if we include partial support.

otthou:
You are right, just blindly enabling this option would make things worse in some cases, so this (partial) support should be explicitly enabled on the client side, if we keep it just as a notification like this.

I looked into the QEMU code for this synchronization, (https://github.com/qemu/qemu/blob/master/ui/vnc.c#L1911), and it looks this synchronization is indeed supported, but only (in case of caps-lock) for keys that are letters a-z. So it doesn't work if the VNC client keyboard layout uses non-latin letters. It makes sense that it can't do that for all keyboard layouts, since it doesn't actually know the clients caps-lock state, only its own. So the best QEMU can do is guess it based on the shift state and the send sym.

And yes, we cannot actually toggle the local led state, due to javascript restrictions, but by supporting this extension we can actually do better than QEMU can, since we actually do know the led state on both sides with it. So apart from showing the remote led state, we can also send a scroll lock/capslock/numlock key if needed to make the remote state match our state, before sending a keypress. This is basically the same fixing behavior that QEMU tries to do if we don't support this extension, but with actually knowing the led state on both sides, it would work in all cases, not just the case where the clients keyboard layout only uses the latin letters.
And for scroll-lock, QEMU doesn't do a fixup at all (it can't obviously, there is no way to even guess the hosts scroll lock state).

CendioOssman:
Yeah, I suppose we could try doing QEMU's fallback in noVNC instead. Even if it isn't the ideal, it should hopefully not be worse than the current state of things.

Except if there is a network issue, I suppose. That's something that QEMU doesn't have to deal with. There might be an issue if we try to send fake CapsLock multiple times just because we haven't got an updated LED state back yet. So some care might be needed.

Could you update the patch with a suggestion for such logic?

core/input/keyboard.js
```diff
@@ -87,6 +89,37 @@ export default class Keyboard {
...
+            this._remoteCapsLock = null;
+        }
+
+        // Same goes for numlock is almost the same, except getModifierState for NumLock doesn't work on Mac,
+        // since Macs don't have a numlock key. So we just never try to fix numlock state on Mac.
+        // Also isn't supported for iOS.
+        if (code == 'NumLock') {
+            this._remoteNumLock = null;
+        }
+        if (!browser.isMac() && !browser.isIOS() && this._remoteNumLock !== null && this._remoteNumLock !== e.getModifierState('NumLock')) {
```
CendioOssman, line 116: It would be nice to avoid platform checks. I don't suppose macOS/iOS set anything on the event that makes it possible to detect that NumLock isn't supported?

Also, if we want to mimic current QEMU behaviour, shouldn't we try to always force NumLock on when on a macOS system?

CendioOssman (changes requested):
Looks mostly good, thanks!

I don't suppose the logic can be moved to rfb.js? So far we've kept keyboard.js rather generic and it would be nice to keep that.

core/input/keyboard.js
otthou: I haven't found a nicer way to detect lack of support no. Unfortunately passing an unsupported value to getModifierState returns false, exactly like numlock being off.

I could change it so numlock locally always counts as on instead for mac/ios yes. The reason I didn't, is that I don't think that behavior is better. Our getKeysym translates 'Clear' to 'NumLock' so that we specifically can toggle remote NumLock even on a mac. I think being able to toggle remote numlock, and use both states is better than not being able to, if the client is a mac, even if you can't see the state on your local keyboard. I am willing to change it though, if you disagree.

core/input/keyboard.js
```diff
@@ -212,17 +220,17 @@ export default class Keyboard {
...
-            this._sendKeyEvent(KeyTable.XK_Control_L, "ControlLeft", true);
+            this._sendKeyEvent(KeyTable.XK_Control_L, "ControlLeft", true, null, null);
         }
 
         // See comment in _handleKeyDown()
         if ((browser.isMac() || browser.isIOS()) && (code === 'CapsLock')) {
-            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', true);
-            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', false);
+            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', true, null, null);
+            this._sendKeyEvent(KeyTable.XK_Caps_Lock, 'CapsLock', false, null, null);
```
CendioOssman, line 229: Did you consider using `undefined` instead so that callers can just omit the lock arguments?

core/rfb.js
```diff
@@ -993,7 +995,35 @@ export default class RFB extends EventTargetMixin {
...
-    _handleKeyEvent(keysym, code, down) {
+    _handleKeyEvent(keysym, code, down, numlock, capslock) {
+        // If remote state of capslock is known, and it doesn't match the local led state of
+        // the keyboard, we send a capslock keypress first to bring it into sync.
+        // If we just pressed CapsLock, or we toggled it remotely due to it being out of sync
+        // we clear the remote state so that we don't send duplicate or spurious fixes,
+        // since it may take some time to receive the new remote CapsLock state.
+        if (code == 'CapsLock' && down) {
+            this._remoteCapsLock = null;
+        }
```
CendioOssman, line 1006: Nice safety logic. :)

core/rfb.js
```diff
@@ -993,7 +995,35 @@ export default class RFB extends EventTargetMixin {
...
+        if (this._remoteCapsLock !== null && capslock !== null && this._remoteCapsLock !== capslock && down) {
+            Log.Warn("Fixing remote caps lock");
```
CendioOssman, line 1008: These syncs should be somewhat common, and normal. So I think a debug log level is more appropriate. We don't want to scare the users with something expected.

otthou: Fair, I'll change it to debug.

core/input/keyboard.js
otthou: I can make null the default value for those parameters so that they can be omitted yes. I prefer that over using undefined directly, since it is more explicitly optional this way.

core/rfb.js
```diff
@@ -2716,6 +2750,25 @@ export default class RFB extends EventTargetMixin {
...
+        let data = this._sock.rQshift8();
+        let scrollLock = data & 1 ? true : false;
+        let numLock = data & 2 ? true : false;
+        let capsLock = data & 4 ? true : false;
+        this._remoteCapsLock = capsLock;
+        this._remoteNumLock = numLock;
+
+        this.dispatchEvent(new CustomEvent(
+            "ledstatus",
+            { detail: { scrollLock: scrollLock, numLock: numLock, capsLock: capsLock } }));
```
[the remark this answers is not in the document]
otthou: True, this event is no longer that useful. I'll remove it.

tests/test.rfb.js
```diff
@@ -3125,28 +3125,24 @@ describe('Remote Frame Buffer Protocol Client', function () {
...
-
-                    expect(spy.args[1][0].detail.capsLock).to.equal(false);
-                    expect(spy.args[1][0].detail.numLock).to.equal(true);
-                    expect(spy.args[1][0].detail.scrollLock).to.equal(true);
-
-                    expect(spy.args[2][0].detail.capsLock).to.equal(false);
-                    expect(spy.args[2][0].detail.numLock).to.equal(false);
-                    expect(spy.args[2][0].detail.scrollLock).to.equal(false);
+                    expect(client._remoteCapsLock).to.equal(false);
+                    expect(client._remoteNumLock).to.equal(false);
```
CendioOssman, line 3145: We want to avoid poking around in internal variables as that makes the tests very fragile.

Do you think you could adjust the tests so that they only look at external behaviour? I.e. that it only sends the fake key presses when supported or needed.

tests/test.rfb.js
```diff
@@ -517,6 +517,143 @@ describe('Remote Frame Buffer Protocol Client', function () {
...
 
+    describe('Caps Lock and Num Lock remote fixup', function () {
+        let client;
+        beforeEach(function () {
+            client = makeRFB();
+            sinon.stub(client, 'sendKey');
+        });
+
+        it('should toggle caps lock if remote caps lock is on and local is off', function () {
+            client._remoteCapsLock = true;
```
CendioOssman, line 528: Ideally the same thing here. Get the tests in to the correct state by faking the right messages.

CendioOssman (commented):
Looks nice, thanks. I just have a final request if you could see if you could get the new tests a bit more robust.

otthou:
@CendioOssman Can you take another look at this?

otthou:
Ok, thanks for letting me know. Also not sure why the linter is failing here. It passes just fine locally, and it looks like it is failing with the same error on master too.

CendioOssman:
Yeah, I don't know what the linter is up to. I can't reproduce the problem here either.

CendioOssman (changes requested):
Looks nice! Just squash everything to a single commit and I can go ahead and merge it.

otthou:
Squashed

CendioOssman (approved):

PR #1801, opened 2023-08-26
Save new expected client size on resize

und3f:
Hello,

`_handleResize()` doesn't save new expected size if window was resized which cause a bug if window was resized to different size, and then return back to original (e.g. go fullscreen and back to windowed).
Since the expected client size wasn't changed during fullscreen resize, the `_handleResize` would not resize window during resize to original size as it would exit because of `this._clientHasExpectedSize()`.

The PR add save of new expected client size to fix the issue

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -724,6 +724,7 @@ export default class RFB extends EventTargetMixin {
         window.requestAnimationFrame(() => {
             this._updateClip();
             this._updateScale();
+            this._saveExpectedClientSize();
         });
 
         if (this._resizeSession) {
--- a/tests/test.rfb.js
+++ b/tests/test.rfb.js
@@ -776,6 +776,25 @@ describe('Remote Frame Buffer Protocol Client', function () {
             expect(client._display.autoscale).to.have.been.calledWith(40, 50);
         });
 
+        it('should update the scaling when the container returns to previous size', function () {
+            client._saveExpectedClientSize();
+
+            sinon.spy(client._display, "autoscale");
+
+            container.style.width = '40px';
+            container.style.height = '50px';
+            fakeResizeObserver.fire();
+            clock.tick(1000);
+
+            container.style.width = '70px';
+            container.style.height = '80px';
+            fakeResizeObserver.fire();
+            clock.tick(1000);
+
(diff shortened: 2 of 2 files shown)


und3f:
Windows test failed because of windows machine itself, please rerun if needed

samhed:
Thanks for your contribution, good to see that you included a unit test. Could you please describe the issue you saw before applying your fix?

samhed:
Resize back and forth between full screen and regular seems to work fine for me. What am I looking for?

und3f:
> Resize back and forth between full screen and regular seems to work fine for me. What am I looking for?

Hi, I've rechecked functionality with `vnc.html` and `vnc_lite.html` files, and it seems that resize works correctly in those.
I am working on reproducible sample, I suspect that it depends on VNC container element display or other parameters.

What I've noticed: there are two resize handlers which are executed separately: `_resize ` and `_handleResize` the main difference is that `_resize` saves new ExpectedClientSize and `_handleResize` doesn't. For both `vnc.html` and `vnc_lite.html` the `_resize` is called on each container element resize.

samhed:
The two functions have very different uses. `_handleResize()` is called when the browser window resizes, and  `_resize()` is called when noVNC receives a "DesktopSize" or "ExtendedDesktopSize" message from the VNC server.

The fact that `_resize()` isn't called in your case indicates that you didn't receive the triggering messages. Are you using a different VNC server when testing vnc.html?

und3f:
I am using same VNC server (qemu's VNC with websocket) for all tests.

Than, it is kinda weird that `_resize()` is executed on html element resize. `_resize` is calling `_saveExpectedClientSize` which is why vnc.html doesn't have bug with resizing back to original size).

Is it correct to assume that my test case correctly emulates expected real world RFB behavior for resizing?

samhed:
> Is it correct to assume that my test case correctly emulates expected real world RFB behavior for resizing?

I'm not sure what you mean. I don't know the details of your test case. You said you were working on a reproducible sample, do you have something like that?

und3f:
Sorry I got busy and wasn't able to create clean example. But I've tried latest noVnc and it has the issue fixed, so we can close it

PR #1805, opened 2023-09-15
Use history API's replaceState() method to hide query for privacy

robertsspaceindustries:
For example; you are using autoconnect in your browser and you want to make a screenshot, it wouldn't be so handy-dandy if your VNC password was exposed. 
If my proposal for this change isn't welcome, or if there are any other factors, feel free to close the pull request.

### DIFF
--- a/app/webutil.js
+++ b/app/webutil.js
@@ -36,6 +36,7 @@ export function getQueryVar(name, defVal) {
     if (typeof defVal === 'undefined') { defVal = null; }
 
     if (match) {
+        history.replaceState(null, "", window.location.pathname.slice(1, 1));
         return decodeURIComponent(match[1]);
     }
 


samhed:
Hi,

Would you mind elaborating a bit, on what you are trying to achieve here? I'm not seeing any difference in either Chrome's history entries or in the URL bar with your code.

I connected to a VNC server using vnc.html, with the following query string `?autoconnect=true&password=123`

robertsspaceindustries:
Heya

The initial commit was broken and I now have moved the code to the end of the connect() function.
I hope this clarifies things, otherwise, feel free to ask questions.

samhed:
Okay, so I can now see that it strips the query part from the visible URL.

And the reason you want to do this is to hide the query-password in screenshots? That sounds like quite a niche usecase. Do you have other arguments for this? I don't know if it is common for web services to strip the query parameters once handled.

robertsspaceindustries:
Hi.

There's nothing more for me to add.
Maybe a togglable setting for this?
Main reason for this proposal stays as privacy.

Feel free to close if you have other ideas.

CendioOssman:
I'm afraid it's a -1 from me. This would be confusing/annoying to most users, and I'm worried it might encourage more use of putting sensitive information in the URL. Something I doubt is ever particularly safe.

PR #1809, opened 2023-10-03
Pinch zoom

sravyak:
Added **auto panning and pinch zoom functionality**. 

My requirement is to make pinch zoom functionality to work on mobile devices and give seamless panning.

I used the pinch gesture inside gesturehandler.js and added an event handlers to make novnc canvas zoom in/zoom out.


samhed (changes requested):

I'm not opposed to the idea of a pinch to zoom gesture. However, there are numerous issues here. First and foremost, noVNC already has a sort of pinch to zoom. Check `core/rfb.js` and the `_handleGesture()` method. It sends `Ctrl+Scroll` to the remote, i.e server-side "zoom". I see that we need to find a method where the user gets to choose between the proposed client-side zoom and the existing server-side variant.

I also tested your code, and while pinch-to-zoom works fine, the biggest user-facing issue is that drag-and-drop is completely disabled now.

Another issue is that you seem to have disregarded the existing architecture and style of the code. In order for this to be maintainable, we need to keep things uniform and coherent. For example, `core/input/gesturehandler.js` is only supposed to detect the actual gesture, not act upon it. That should be the responsibility of `_handleGesture()` in RFB.

bitbound:
@samhed  I've been testing noVNC in one of my prototype projects ([ControlR](https://github.com/bitbound/ControlR)) by replacing the WebRTC-based remote control with VNC.  I don't think I have it in me to fully implement yet another remote control app, and it'd be nice to use (and contribute back to) an existing project.

So far, noVNC is working really well.  I'm proxying it through ASP.NET Core, to my agent running on the remote device, then to the loopback-only VNC server.  However, I do find the mobile controls kind of challenging to use.  Pinch zoom would help, but I think it needs a little more than that.

Would you be open to my adding a "touchpad mode" button under the mobile buttons section?  It would only change behavior when explicitly enabled, so it shouldn't affect any existing features.  It would behave similarly to other mobile remote control clients (RealVNC, Microsoft RDP, etc.), with the following behaviors:

- Single-finger touch-move would move the cursor in the gesture direction.
  - The viewport will try to stay centered on the cursor.  If the viewport can't be scrolled further in that direction , the cursor moves toward the edge.
- Left-click would be single tap.
- Right-click could either be two-finger tap or long-press-then-release.
- Left-click-and-drag would be double-tap, hold, then drag.
- Pinch-zoom would zoom the viewport.
- Two-finger gesture up and down would do mouse wheel scroll.

Below is how I figure it would look.  Assuming I conform to contribution guidelines and existing code style, would this be okay?

image

app/ui.js
```diff
@@ -229,8 +229,9 @@ const UI = {
         document.getElementById("noVNC_control_bar")
             .addEventListener('keydown', UI.keepControlbar);
 
-        document.getElementById("noVNC_view_drag_button")
-            .addEventListener('click', UI.toggleViewDrag);
+        // // Agilicus Modified
+        // document.getElementById("noVNC_view_drag_button")
+        //     .addEventListener('click', UI.toggleViewDrag);
```
kalvdans, line 234: Just remove the code instead of uncommenting.

samhed:
@bitbound yes. Something like that would indeed be very nice. Is this a feature you have implemented in a fork?

bitbound:
@samhed  I haven't started implementing it yet.  I just threw the button there as a visual aid.

Since you're okay with the idea, I might work on it when I have some free time.  In my previous projects, I've always used tap-based controls with pinch-zoom, so doing a "touchpad mode" will be new territory for me.  It was requested a few times, though.  It should be an interesting exercise.  :)

samhed:
Ok, sounds good. I'm looking forward to your PR!

samhed:
> @samhed   This PR has been stuck on my homepage and notifications for over two years, and I haven't been able to get rid of it.  Could you please delete your mention of me so I can see if that will do it?

Deleted

bitbound:
@samhed   Thanks.  Unfortunately, that didn't do it.

PR #1810, opened 2023-10-11
Add the possibility to listen on a specific host

### LINKED ISSUE #1798: Specifying a bind address does not work
**Describe the bug**
The instructions given for binding to a specific address (--listen \<address\>:\<port\>) do not appear to work. 

From the [Quick Start instructions](https://github.com/novnc/noVNC):
- **"If you don't need to expose the web server to public internet, you can bind to localhost:"**
\
`./utils/novnc_proxy --vnc localhost:5901 --listen localhost:6081`"

It does not bind to that address, and the resulting URL given when it is used this way doesn't make sense.

**Server:**
 - noVNC version: 1.4, and git
 - WebSocket proxy: websockify

afbjorklund:
For instance, for listening only on "localhost"

That is, bind on 127.0.0.1 instead of 0.0.0.0

Closes #1798

### DIFF
--- a/docs/novnc_proxy.1
+++ b/docs/novnc_proxy.1
@@ -3,12 +3,12 @@
 .SH NAME
 novnc_proxy - noVNC proxy server
 .SH SYNOPSIS
-.B novnc_proxy [--listen PORT] [--vnc VNC_HOST:PORT] [--cert CERT] [--ssl-only]
+.B novnc_proxy [--listen [HOST:]PORT] [--vnc VNC_HOST:PORT] [--cert CERT] [--ssl-only]
 
 Starts the WebSockets proxy and a mini-webserver and
 provides a cut-and-paste URL to go to.
 
-    --listen PORT         Port for proxy/webserver to listen on
+    --listen [HOST:]PORT  Port for proxy/webserver to listen on
                           Default: 6080
     --vnc VNC_HOST:PORT   VNC server host:port proxy target
                           Default: localhost:5900
--- a/docs/novnc_proxy.1
+++ b/docs/novnc_proxy.1
@@ -9,7 +9,7 @@ Starts the WebSockets proxy and a mini-webserver and
 provides a cut-and-paste URL to go to.
 
     --listen [HOST:]PORT  Port for proxy/webserver to listen on
-                          Default: 6080
+                          Default: 6080 (on all interfaces)
     --vnc VNC_HOST:PORT   VNC server host:port proxy target
                           Default: localhost:5900
     --cert CERT           Path to combined cert/key file, or just
--- a/utils/novnc_proxy
+++ b/utils/novnc_proxy
@@ -14,7 +14,7 @@ usage() {
     echo "provides a cut-and-paste URL to go to."
     echo
     echo "    --listen [HOST:]PORT  Port for proxy/webserver to listen on"
-    echo "                          Default: 6080"
+    echo "                          Default: 6080 (on all interfaces)"
     echo "    --vnc VNC_HOST:PORT   VNC server host:port proxy target"
     echo "                          Default: localhost:5900"
     echo "    --cert CERT           Path to combined cert/key file, or just"
--- a/utils/novnc_proxy
+++ b/utils/novnc_proxy
@@ -8,12 +8,12 @@ usage() {
         echo "$*"
         echo
     fi
-    echo "Usage: ${NAME} [--listen PORT] [--vnc VNC_HOST:PORT] [--cert CERT] [--ssl-only]"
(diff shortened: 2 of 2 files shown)


docs/novnc_proxy.1
```diff
@@ -3,12 +3,12 @@
...
 .SH SYNOPSIS
-.B novnc_proxy [--listen PORT] [--vnc VNC_HOST:PORT] [--cert CERT] [--ssl-only]
+.B novnc_proxy [--listen [HOST:]PORT] [--vnc VNC_HOST:PORT] [--cert CERT] [--ssl-only]
 
 Starts the WebSockets proxy and a mini-webserver and
 provides a cut-and-paste URL to go to.
 
-    --listen PORT         Port for proxy/webserver to listen on
+    --listen [HOST:]PORT  Port for proxy/webserver to listen on
                           Default: 6080
```
kalvdans, line 12:
```suggestion
                          Default: 6080 on all interfaces
```

utils/novnc_proxy
```diff
@@ -191,9 +200,9 @@ else
     fi
 fi
 
-echo "Starting webserver and WebSockets proxy on port ${PORT}"
-#${HERE}/websockify --web ${WEB} ${CERT:+--cert ${CERT}} ${PORT} ${VNC_DEST} &
-${WEBSOCKIFY} ${SYSLOG_ARG} ${SSLONLY} ${FILEONLY_ARG} --web ${WEB} ${CERT:+--cert ${CERT}} ${KEY:+--key ${KEY}} ${PORT} ${VNC_DEST} ${HEARTBEAT_ARG} ${IDLETIMEOUT_ARG} ${RECORD_ARG} ${TIMEOUT_ARG} ${WEBAUTH_ARG} ${AUTHPLUGIN_ARG} ${AUTHSOURCE_ARG} &
+echo "Starting webserver and WebSockets proxy on host ${HOST} port ${PORT}"
```
kalvdans, line 203:
```suggestion
echo "Starting webserver and WebSockets proxy on port $LISTEN"
```

Your suggestion is slightly incorrect. The default setting listens to all interfaces but the message will tell you it listens on a single interface.

utils/novnc_proxy
```diff
@@ -107,8 +109,15 @@ while [ "$*" ]; do
...
+if [ "$LISTEN" != "$PORT" ]; then
+    HOST=${LISTEN%:*}
+    PORT=${LISTEN##*:}
+    # if no host was given, restore hostname
+    [ "$HOST" = "$PORT" ] && HOST=$(hostname)
+fi
+
 # Sanity checks
-if bash -c "exec 7<>/dev/tcp/localhost/${PORT}" &> /dev/null; then
+if bash -c "exec 7<>/dev/tcp/${HOST}/${PORT}" &> /dev/null; then
```
kalvdans, line 120: Just disable the sanity check if we give the `--listen` option. Then we can remove the $HOST variable altogether.

kalvdans:
It's more complicated that it needs to be, but it is fine that it just passes the argument unmodified to websockify, and the complications are only for pretty-printing.

LGTM besides my small remarks.

utils/novnc_proxy
```diff
@@ -191,9 +202,9 @@ else
     fi
 fi
 
-echo "Starting webserver and WebSockets proxy on port ${PORT}"
-#${HERE}/websockify --web ${WEB} ${CERT:+--cert ${CERT}} ${PORT} ${VNC_DEST} &
-${WEBSOCKIFY} ${SYSLOG_ARG} ${SSLONLY} ${FILEONLY_ARG} --web ${WEB} ${CERT:+--cert ${CERT}} ${KEY:+--key ${KEY}} ${PORT} ${VNC_DEST} ${HEARTBEAT_ARG} ${IDLETIMEOUT_ARG} ${RECORD_ARG} ${TIMEOUT_ARG} ${WEBAUTH_ARG} ${AUTHPLUGIN_ARG} ${AUTHSOURCE_ARG} &
+echo "Starting webserver and WebSockets proxy on${HOST:+ host ${HOST}} port ${PORT}"
+#${HERE}/websockify --web ${WEB} ${CERT:+--cert ${CERT}} ${LISTEN} ${VNC_DEST} &
```
kalvdans, line 206: Remove commented out lines

utils/novnc_proxy
afbjorklund: You will still get an error (exception):
```
OSError: [Errno 98] Address already in use
Failed to start WebSockets proxy
```

CendioOssman (approved):
Looks good to me. Nice work @afbjorklund!

And thanks for the help with reviewing, @kalvdans.

PR #1835, opened 2024-02-04
Add "touchpad mode" with pinch-zoom for mobile view.

### LINKED ISSUE #1764: Add simulated touchpad option for touchscreen devices

Option for a simulated touchpad when used on touchscreen;

Connecting a mouse

**Additional context**
It would make using noVNC on a touchscreen much easier. It's a complete pain to use on touchscreen devices otherwise.

bitbound:
This PR adds a "touchpad mode" option for input when in mobile view.  It includes pinch-zoom.

Screenshot_20240204-114726

In touchpad mode, it will function similar to other remote control apps (RealVNC, TeamViewer, etc.).  Touch movements will "push" the cursor in the same direction, and tap events will occur at the cursor's current location.

Long-press or two-tap will perform a right-click.  Double-tap-and-drag will perform a click-and-drag with the left mouse button.

https://github.com/novnc/noVNC/assets/20995508/99c448fe-d1b4-4123-863f-ae8564ae1833

### DIFF
--- a/app/images/touchpad.svg
+++ b/app/images/touchpad.svg
@@ -1,0 +1,62 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="25"
+   height="25"
+   viewBox="0 0 25 25"
+   version="1.1"
+   id="touchpad"
+   inkscape:version="1.3 (0e150ed6c4, 2023-07-21)"
+   sodipodi:docname="touchpad.svg"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <sodipodi:namedview
+     id="namedview1"
+     pagecolor="#505050"
+     bordercolor="#eeeeee"
+     borderopacity="1"
+     inkscape:showpageshadow="0"
+     inkscape:pageopacity="0"
+     inkscape:pagecheckerboard="0"
+     inkscape:deskcolor="#505050"
+     inkscape:document-units="mm"
+     inkscape:zoom="11.588216"
+     inkscape:cx="9.7081384"
+     inkscape:cy="15.921347"
+     inkscape:window-width="1920"
+     inkscape:window-height="1009"
+     inkscape:window-x="-8"
+     inkscape:window-y="-8"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="layer1" />
+  <defs
+     id="defs1" />
+  <g
+     inkscape:label="Layer 1"
+     inkscape:groupmode="layer"
+     id="layer1">
+    <g
+       id="g1">
+      <rect
+         style="fill:none;stroke:#ffffff;stroke-width:1.5;stroke-dasharray:none;stroke-opacity:1"
+         id="rect1"
+         width="20.896252"
+         height="20.896252"
+         x="1.8494672"
+         y="2.1561899"
+         rx="3.3434002"
+         ry="3.3434005" />
+      <path
+         style="fill:none;stroke:#ffffff;stroke-width:1.43009;stroke-dasharray:none;stroke-opacity:1"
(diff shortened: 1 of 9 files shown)


CendioOssman:
This looks promising, thanks!

I've not looked at things in detail yet, but there are two things I noticed so far:

 * An easier zoom for mobile devices is something that is generally needed, so I don't think it should be conditional on being in touchpad mode. 
 * Do we really need to be able to quickly toggle touchpad mode? I figured it was something that could be in the settings dialog, rather than taking up toolbar space.

bitbound:
> This looks promising, thanks!

Happy to contribute!  This is an awesome project.  Thanks for all your work on it.

> An easier zoom for mobile devices is something that is generally needed, so I don't think it should be conditional on being in touchpad mode.

Should I create a separate option for this in settings?  If so, any preference on what to call it?  "Free Scale" is the first term that comes to my mind.

> Do we really need to be able to quickly toggle touchpad mode? 

That's totally up to you (and other maintainers).  Since drag mode, default mode, and touchpad mode are mutually exclusive, my first thought was to put it here to easily see the relationship.

LMattsson:
If we don't want to populate the side bar too much, I see a few alternatives:

* Can this replace view-drag-mode (hand icon)? 
* Could view-drag-mode and touchpad mode be combined into one button that a small panel, like the extra keyboard button?
  - Alternatively, have a single button with 3 states.

samhed (changes requested):
Thank you for your patience.

We have now tested your changes and I must say the overall impression and feeling is very nice!

We have used a phone for testing, and the mapping between finger and mouse movements feels well calibrated. :1st_place_medal: 

We think this is a very impressive feature that will give our users a better
experience, especially if using small screens. :iphone: 

---

Synaptics is, in our minds, the most well-known developer of trackpads. For reference, here's a list of “gestures” for the Synaptics driver on Linux:

https://man.archlinux.org/man/synaptics.4

However, we don't think we must support all the features they support. The most important thing is that your trackpad implementation doesn't conflict with theirs, since that's what most users expect from a trackpad.

---

Aside from testing your implementation, we also compared it with Microsoft's RDP Android app, and RealVNC's Android app:

* In general, we think your implementation matches great with the “mouse mode” in Microsoft's RDP app. They both use the same gestures and actions. The feeling is very similar, that is an excellent thing.

   There are two small differences when compared to Microsoft's RDP; Firstly, they have some inertia that makes the cursor stop softly after lifting the finger (it continues moving a little bit, slower and slower until coming to a complete stop). We're not sure whether we like this effect. Secondly, their center area is larger, the area where the cursor can move freely without moving the viewport.
 
   Similar to noVNC, Microsoft's RDP also has a “touch mode”, which behaves like our standard mode. The gestures they use in the equivalent mode are listed here:

   https://learn.microsoft.com/en-us/azure/virtual-desktop/users/client-features-android-chrome-os#use-touch-gestures-and-mouse-modes-in-a-remote-session

   Screenshot_20240228_155737_Microsoft Remote Desktop

* RealVNC only has this trackpad mode, no mode where touch positions are directly mapped to mouse movements. Your implementation matches their trackpad mode well in terms of available gestures and how they function. One major difference is that the cursor is always centered in RealVNC's variant. We like your implementation better.

   Screenshot_20240228_155844_RVNC Viewer

   RealVNC has the option of showing a “mouse button overlay”, see below screenshot.

   Screenshot_20240228_160543_RVNC Viewer

   While this might be useful in some cases, we don't think it is something we need
to include in the first version.

---

After testing the other apps, I noticed that Microsoft refers to their “standard mode” as “touch mode”, and their trackpad mode as “mouse mode”. I'm a bit worried that your “touchpad mode” could easily be confused with “touch mode”, which would be incorrect. What are your thoughts on renaming your new mode to “trackpad mode”?

---

Lastly, have you considered unit tests for your code?

samhed:
We also have some future ideas for changes to the view drag mode (the “hand”). With trackpad mode in place, it would not be as critical anymore.

samhed:
@bitbound are you interested in continuing work on this?

bitbound:
> @bitbound are you interested in continuing working on this?

I'm sorry.  I forgot to update this PR.

I took my project in another direction (WebRTC) and won't be using noVNC anymore.  I won't have time to finish this.

Of course, you or anyone else is free to take over and finish this if they want.

Cheers!

marcnotenboom:
I created my own solution. Grab it if you need it. Feed it to your favorite LLM like Opus 4.7 to tailored it to your needs.

https://github.com/marcnotenboom/novnc-trackpad-overlay

PR #1845, opened 2024-03-25
Clipboard: handle multiple CR+LF

syrykh:
In case copied from the remote server text contains multiple new-lines, only the first occurence of "\r\n" is properly replaced with '\n'. If the received text is then directly written to clipboard with `navigator.clipboard.writeText(e.detail.text)`, empty lines will be duplicated on macOS.

image

This PR addresses this issue by replacing all the occurrences instead.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -2356,7 +2356,7 @@ export default class RFB extends EventTargetMixin {
                         textData = textData.slice(0, -1);
                     }
 
-                    textData = textData.replace("\r\n", "\n");
+                    textData = textData.replaceAll("\r\n", "\n");
 
                     this.dispatchEvent(new CustomEvent(
                         "clipboard",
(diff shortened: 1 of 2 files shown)


CendioOssman (approved):
Thanks! That's a bit embarrassing that it was overlooked. :)

samhed:
Thank you for the fix.

Out of interest, I attempted — and failed — to see the practical effects of this. You mentioned “empty lines will be duplicated on macOS”, what are the steps for this to happen?

What I did:

On the remote (RHEL 9 with MATE), I created a text file with `\r\n` linebreaks:

```
echo -ne "1\r\n" >> myfile
echo -ne "2\r\n" >> myfile
echo -ne "3" >> myfile
```

I then connected using noVNC from Safari on macOS. I opened `myfile` in both Pluma and vim and both apps displayed the file as containing 3 rows as expected. After copying the contents I still see it as 3 rows, both in the clipboard textarea in Safari, and in macOS notes when pasted.

Note that I can verify, using Safari's developer tools, that I see the bug. The `textData` in my JavaScript event contains `1\n2\r\n3`.

PR #1849, opened 2024-04-11
Resize the browser to match VNC session

XiaoXianNv-boot:
**I've added a new zoom mode that allows the browser to automatically adapt to the window .**
Just need to use window.open('vnc. html/?token=token&autoconnect=1&resize=browser','_blank','toolbar=no,location=no,status=no,menubar=no,resizable=yes,width=800,height=420'); Open a window
Works fine on EDGE

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -42,6 +42,10 @@ const UI = {
     reconnectCallback: null,
     reconnectPassword: null,
 
+    // Old browser resolution
+    bodyWidth_browser_resize:  0,
+    bodyHeight_browser_resize: 0,
+
     prime() {
         return WebUtil.initSettings().then(() => {
             if (document.readyState === "interactive" || document.readyState === "complete") {
@@ -988,6 +992,28 @@ const UI = {
             .classList.remove("noVNC_open");
     },
 
+    _updateBrowserWindows(width, height) {
+        let bodyWidth  = document.body.clientWidth;
+        let bodyHeight = document.body.clientHeight;
+
+        let OldResolutionEqual = false;
+        if (UI.bodyWidth_browser_resize === document.body.clientWidth &&
+            UI.bodyHeight_browser_resize === document.body.clientHeight) {
+                OldResolutionEqual = true;
+        }
+        if (UI.bodyHeight_browser_resize === 0 ||
+            OldResolutionEqual) {
+            if((width != 0) && (height != 0)) {
+                window.resizeBy(width - bodyWidth, height - bodyHeight);
+                UI.bodyWidth_browser_resize = width;
+                UI.bodyHeight_browser_resize = height;
+            }
+        } else {
+            // disabled
+            UI.bodyWidth_browser_resize = 0;
+        }
+    },
+
     connect(event, password) {
 
         // Ignore when rfb already exists
@@ -1058,6 +1084,12 @@ const UI = {
         UI.rfb.clipViewport = UI.getSetting('view_clip');
         UI.rfb.scaleViewport = UI.getSetting('resize') === 'scale';
         UI.rfb.resizeSession = UI.getSetting('resize') === 'remote';
+        if (UI.getSetting('resize') === 'off') {
+            UI.bodyHeight_browser_resize = 0;
+            UI.rfb.resizeBrowser = UI._updateBrowserWindows;
+        } else {
+            UI.rfb.resizeBrowser = false;
+        }
         UI.rfb.qualityLevel = parseInt(UI.getSetting('quality'));
         UI.rfb.compressionLevel = parseInt(UI.getSetting('compression'));
         UI.rfb.showDotCursor = UI.getSetting('show_dot');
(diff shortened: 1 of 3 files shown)


CendioOssman:

I'm thinking this might be a bit too niche, though, since it only works under very specific circumstances. Could you describe your setup a bit more, and where this feature fits in?

XiaoXianNv-boot:
Used to open the QEMU/KVM console in a new browser window, yes, QEMU/KVM will constantly change the resolution when it is launched

XiaoXianNv-boot:
I'm connected to the VNC port of KVM/QEMU via websockify

XiaoXianNv-boot:
[Browser APIs used: https://developer.mozilla.org/en-US/docs/Web/API/Window/resizeBy](https://developer.mozilla.org/en-US/docs/Web/API/Window/resizeBy)

CendioOssman (changes requested):
Okay. That sounds reasonble. I'm cautious about adding niche stuff to the general UI, though.

In TigerVNC, we have a similar behaviour but without a setting. Perhaps that could be used here as well?

The principle there is that it will resize the local window if the sizes match before the resize, but not if they are different. I.e.

1. VNC session is 1024x768, local window is 1024x768. VNC session resizes to 1920x1080, and the local window tries to resize to the same 1920x1080.
2. VNC session is 1024x768, but local window is 800x600. VNC session resizes to 1920x1080, but the local window stays at 800x600.

CendioOssman:
@XiaoXianNv-boot, are you interested in looking at the suggested changes?

XiaoXianNv-boot:
> @XiaoXianNv-boot, are you interested in looking at the suggested changes?

YES

XiaoXianNv-boot:
> Okay. That sounds reasonble. I'm cautious about adding niche stuff to the general UI, though.
> 
> In TigerVNC, we have a similar behaviour but without a setting. Perhaps that could be used here as well?
> 
> The principle there is that it will resize the local window if the sizes match before the resize, but not if they are different. I.e.
> 
> 1. VNC session is 1024x768, local window is 1024x768. VNC session resizes to 1920x1080, and the local window tries to resize to the same 1920x1080.
> 2. VNC session is 1024x768, but local window is 800x600. VNC session resizes to 1920x1080, but the local window stays at 800x600.

I have added adjustment rules now. Meet automatic adjustment in the following circumstances
1. The first adjustment size after setting this mode
2. The size of the current browser window is the size of the previous adjustment

If the above requirements are not met, the size of the browser will be disabled.
After adjusting to other modes, you can re -enable this mode (View Rules 1)

vnc.html
```diff
@@ -171,6 +171,7 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br>VNC</h1>
                             <option value="off">None</option>
                             <option value="scale">Local Scaling</option>
                             <option value="remote">Remote Resizing</option>
+                            <option value="browser" disabled="" id="noVNC_setting_browser_resize">Resizing browser</option>
```
CendioOssman, line 174: I don't think we need a new setting for this. This should be possible to have active as long as things aren't configured to `"scale"`.

core/rfb.js
```diff
@@ -2863,10 +2879,36 @@ export default class RFB extends EventTargetMixin {
                                              this._fbWidth, this._fbHeight);
     }
 
+    _updateBrowserWindows(width, height) {
```
CendioOssman, line 2882: The RFB object isn't really aware of things outside of its container. So I think these things are better placed in the UI object.

That means it needs some more information, though. We require `width` and `height` read-only properties on the RFB object. And we also need a `resize` event to be fired.

core/rfb.js
```diff
@@ -2863,10 +2879,36 @@ export default class RFB extends EventTargetMixin {
...
+        if (this._resizeBrowser &&
+            this._rfbConnectionState === 'connected') {
+            const fbWidth  = width;
+            const fbHeight = height;
+            let OldResolutionEqual = false;
+
+            if (bodyWidthBrowserResize === document.body.clientWidth &&
+                bodyHeightBrowserResize === document.body.clientHeight) {
+                OldResolutionEqual = true;
+            }
```
CendioOssman, line 2892: I'm not sure if this gives the desired behaviour. What we want to compare with is `_fbWidth` and `_fbHeight` just before the resize, and see if they match the browser window.

This code might still be doing that, but it's not obvious that it does.

CendioOssman (changes requested):
Looks like a good step in the right direction. But I think some more adjustments are needed to get a good fit.

Please have a look and see the comments.

XiaoXianNv-boot:
> 看起来这是意外关闭的，？

YES

XiaoXianNv-boot:
What is the situation

app/ui.js
```diff
@@ -1058,6 +1084,12 @@ const UI = {
         UI.rfb.clipViewport = UI.getSetting('view_clip');
         UI.rfb.scaleViewport = UI.getSetting('resize') === 'scale';
         UI.rfb.resizeSession = UI.getSetting('resize') === 'remote';
+        if (UI.getSetting('resize') === 'off') {
```
CendioOssman, line 1087: It should be relevant for `'remote'` as well, I think?

CendioOssman (changes requested):

PR #1850, opened 2024-04-19
Update zh_CN.json

SuperHuangXu:


CendioOssman:

Unfortunately, the primary files for the translations are the files in the `po/` directory. Do you think you could have a look at updating `po/zh_CN.po` instead?

SuperHuangXu:
> Thank you for your contribution!
> 
> Unfortunately, the primary files for the translations are the files in the `po/` directory. Do you think you could have a look at updating `po/zh_CN.po` instead?

Okay, I've resubmitted a PR #1851

PR #1852, opened 2024-04-19
Update greek translations

kosmasgiannis:


samhed:
Thank you!

Only the files in the po/ directory should be changed I'm afraid, the .json files are generated. Do you think you could update your commit to only include the .po file?

PR #1872, opened 2024-06-26
ping keepalive feature

sancelot:
Implements a keepalive feature

x11vnc server implements 1x1 frame update buffer ping every n seconds 

I am managing it on the client side to know if connection is broken .

to disable feature KEEP_ALIVE_INTERVAL must be set <= 0

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -91,7 +91,7 @@ const extendedClipboardActionProvide = 1 << 28;
 export default class RFB extends EventTargetMixin {
     constructor(target, urlOrChannel, options) {
         if (!target) {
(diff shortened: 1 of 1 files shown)


CendioOssman:

Unfortunately, this approach is not very robust and not something we can accept. A 1×1 rect is not guaranteed to be a keep-alive ping, so we cannot make any assumption based on seeing one. We also don't know if the connection is permanently dead, so we should leave the decision of disconnecting up to the user. We should simply inform the user that it seems like the connection is stalled.

Please see the discussion thread you responded to for some guidelines on how this could be solved more robustly.

PR #1876, opened 2024-07-08
WebCodecs H.264

any1:
This adds the Open H.264 encoding to noVNC, as defined here: https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#open-h-264-encoding

This was tested on the following browsers:
 * Firefox 127.0.2  with `dom.media.webcodecs.enabled=true`
 * Firefox Nightly 129.0a1 (2024-07-08)
 * Chromium 126.0.6478.126

I haven't run into any issues with Chromium, but Firefox keeps killing the decoder after a few frames because it thinks that the hardware decoder is too slow, so it wants to switch to software. However, after the switch, a new key frame is needed, so the decoder goes into a bad state and a reconfiguration is required, so this sequence of events is repeated indefinitely.

We can work around Firefox's issue by setting a frame duration for each frame, but that would be a hack. We don't know the frame duration for sure, and it is an optional parameter so Firefox **should** not rely on it.

I created an issue in bugzilla here: https://bugzilla.mozilla.org/show_bug.cgi?id=1906761

On the server side, I tested this against wayvnc.
```
WLR_BACKENDS=headless sway &
WAYLAND_DISPLAY=wayland-2 wayvnc -g
```

WayVNC can also run as a WebSocket server using the `-w` flag.

### DIFF
--- a/core/decoders/h264.js
+++ b/core/decoders/h264.js
@@ -1,0 +1,321 @@
+/*
+ * noVNC: HTML5 VNC client
+ * Copyright (C) 2024 The noVNC Authors
+ * Licensed under MPL 2.0 (see LICENSE.txt)
+ *
+ * See README.md for usage and integration instructions.
+ *
+ */
+
+import * as Log from '../util/logging.js';
+
+class H264Parser {
+    constructor(data) {
+        this._data = data;
+        this._index = 0;
+        this.profileIdc = null;
+        this.constraintSet = null;
+        this.levelIdc = null;
+    }
+
+    _getStartSequenceLen(index) {
+        let data = this._data;
+        if (data[index + 0] == 0 && data[index + 1] == 0 && data[index + 2] == 0 && data[index + 3] == 1) {
+            return 4;
+        }
+        if (data[index + 0] == 0 && data[index + 1] == 0 && data[index + 2] == 1) {
+            return 3;
+        }
+        return 0;
+    }
+
+    _indexOfNextNalUnit(index) {
+        let data = this._data;
+        for (let i = index; i < data.length; ++i) {
+            if (this._getStartSequenceLen(i) != 0) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    _parseSps(index) {
+        this.profileIdc = this._data[index];
+        this.constraintSet = this._data[index + 1];
+        this.levelIdc = this._data[index + 2];
+    }
+
+    _parseNalUnit(index) {
+        const firstByte = this._data[index];
+        if (firstByte & 0x80) {
+            throw new Error('H264 parsing sanity check failed, forbidden zero bit is set');
+        }
+        const unitType = firstByte & 0x1f;
+
+        switch (unitType) {
+            case 1: // coded slice, non-idr
+                return { slice: true };
+            case 5: // coded slice, idr
+                return { slice: true, key: true };
+            case 6: // sei
+                return {};
+            case 7: // sps
+                this._parseSps(index + 1);
+                return {};
+            case 8: // pps
+                return {};
+            default:
+                Log.Warn("Unhandled unit type: ", unitType);
+                break;
+        }
+        return {};
+    }
+
+    parse() {
+        const startIndex = this._index;
+        let isKey = false;
+
+        while (this._index < this._data.length) {
+            const startSequenceLen = this._getStartSequenceLen(this._index);
+            if (startSequenceLen == 0) {
+                throw new Error('Invalid start sequence in bit stream');
+            }
+
+            const { slice, key } = this._parseNalUnit(this._index + startSequenceLen);
+
+            let nextIndex = this._indexOfNextNalUnit(this._index + startSequenceLen);
+            if (nextIndex == -1) {
+                this._index = this._data.length;
+            } else {
+                this._index = nextIndex;
+            }
+
+            if (key) {
+                isKey = true;
+            }
+            if (slice) {
+                break;
+            }
+        }
+
+        if (startIndex === this._index) {
+            return null;
+        }
+
+        return {
+            frame: this._data.subarray(startIndex, this._index),
+            key: isKey,
+        };
+    }
+}
+
+class H264Context {
+    constructor(width, height) {
+        this.lastUsed = 0;
+        this._width = width;
+        this._height = height;
+        this._profileIdc = null;
+        this._constraintSet = null;
+        this._levelIdc = null;
+        this._decoder = null;
+        this._pendingFrames = [];
+    }
+
+    _handleFrame(frame) {
+        let pending = this._pendingFrames.shift();
+        if (pending === undefined) {
+            throw new Error("Pending frame queue empty when receiving frame from decoder");
+        }
+
+        if (pending.timestamp != frame.timestamp) {
+            throw new Error("Video frame timestamp mismatch. Expected " +
+                frame.timestamp + " but but got " + pending.timestamp);
+        }
+
+        pending.frame = frame;
+        pending.ready = true;
+        pending.resolve();
+
+        if (!pending.keep) {
+            frame.close();
+        }
+    }
+
+    _handleError(e) {
+        throw new Error("Failed to decode frame: " + e.message);
+    }
+
+    _configureDecoder(profileIdc, constraintSet, levelIdc) {
+        if (this._decoder === null || this._decoder.state === 'closed') {
+            this._decoder = new VideoDecoder({
+                output: frame => this._handleFrame(frame),
+                error: e => this._handleError(e),
+            });
+        }
+        const codec = 'avc1.' +
+            profileIdc.toString(16).padStart(2, '0') +
+            constraintSet.toString(16).padStart(2, '0') +
+            levelIdc.toString(16).padStart(2, '0');
+        this._decoder.configure({
+            codec: codec,
+            codedWidth: this._width,
+            codedHeight: this._height,
+            optimizeForLatency: true,
+        });
+    }
+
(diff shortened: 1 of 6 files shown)


core/display.js
```diff
@@ -380,6 +380,17 @@ export default class Display {
...
+    videoFrame(x, y, width, height, frame) {
+        this._renderQPush({
+            'type': 'frame',
+            'frame': frame,
+            'x': x,
+            'y': y,
+            'width': width,
+            'height': height
+        });
+    }
```
CendioOssman, line 392: I'm not too thrilled about adding stuff to the general Display class that is specific to one decoder.

But I suppose it is small enough that we can refactor it later if needed.

core/display.js
```diff
@@ -511,6 +522,36 @@ export default class Display {
...
+                        const sx = 0;
+                        const sy = 0;
+                        const sw = a.width;
+                        const sh = a.height;
+                        const dx = a.x;
+                        const dy = a.y;
+                        const dw = sw;
+                        const dh = sh;
+                        this._drawCtx.drawImage(frame, sx, sy, sw, sh, dx, dy, dw, dh);
+                        this._damage(a.x, a.y, a.width, a.height);
```
CendioOssman, line 545: I would prefer if we don't hide actual draw operations down here, though. Please extend `Desktop.drawImage()` instead.

core/rfb.js
```diff
@@ -2115,6 +2128,9 @@ export default class RFB extends EventTargetMixin {
         encs.push(encodings.encodingCopyRect);
         // Only supported with full depth support
         if (this._fbDepth == 24) {
+            if (H264Decoder.check()) {
+                encs.push(encodings.encodingH264);
+            }
```
CendioOssman, line 2133: We also need to check that the H.264 encoding is supported by the browser. That is far from universal.

core/decoders/h264.js
```diff
@@ -0,0 +1,346 @@
...
+                this._decoder.decode(chunk);
+            } catch (e) {
+                Log.Warn("Failed to decode:", e);
+            }
+
+            // If the decoder wants to use the timestamp for approximating a
+            // minimum frame duration, it's better to have a non-zero difference
+            // between frames, so we will assume 60 Hz refresh rate as it is the
+            // most common one.
+            timestamp += 16667;
```
CendioOssman, line 249: What if we put a huge value here to work around Firefox bug?

core/rfb.js
```diff
@@ -804,6 +808,15 @@ export default class RFB extends EventTargetMixin {
...
     }
 
+    _requestKeyFrame() {
+        if (this._enabledContinuousUpdates) {
+            RFB.messages.fbUpdateRequest(this._sock, false, 0, 0,
+                                         this._fbWidth, this._fbHeight);
+        } else {
+            this._keyFrameNeeded = true;
+        }
+    }
```
CendioOssman, line 818: This seems overly complex to work around a bug in Firefox as the feature isn't deemed ready for usage there yet.

And I don't understand why this is needed. You wrote you worked around it by setting a frame duration?

any1:
> We try to make sure all features in noVNC work on all current browsers, so it's annoying that Firefox isn't ready yet. :/ Is the new API the only method? Have you looked at using the older WebRTC API?
> 

WebRTC is not a good fit for this as it does not expose an API for decoding video at the required level of abstraction.

CendioOssman:
The WebCodecs definitely looks like a better fit. No arguments there. But if WebCodecs isn't fully supported yet, the question is if it's possible to build a more complex solution around WebRTC that works on more browsers.

any1:
> The WebCodecs definitely looks like a better fit. No arguments there. But if WebCodecs isn't fully supported yet, the question is if it's possible to build a more complex solution around WebRTC that works on more browsers.

I suppose that it might be possible to create two peers within the browser, connect them together and stream video from one to the other which would render into an offscreen video element that gets duplicated onto the canvas. I have not explored this option. It's just hypothetical. It would be very hacky.

In any case, I don't think that this would be worth the effort as WebCodec support is very close to being complete in Firefox.

core/decoders/h264.js
CendioOssman: Ah. And what if we just put the time we got the frame? That's technically when it's supposed to be shown given the current system.

core/rfb.js
CendioOssman: I see. In that case, I'm not sure if it's worth the complexity. Can we leave it out for now so we can focus on getting the core stuff merged?

core/decoders/h264.js
any1: > Ah. And what if we just put the time we got the frame?

That's the same as dropping this hunk entirely, which is probably best anyway. This just guards against hypothetical silliness in browsers in the future. It's probably better to handle that if/when it happens.

core/rfb.js
any1: We can leave it out, but how should we handle decoding failures then? Raise an exception?

CendioOssman: That seems reasonable. That's how we handle problems in the other decoders.

VNC is a reliable transport, so it should not expect dropped frames or such that other video transports might have to deal with.

any1: > VNC is a reliable transport, so it should not expect dropped frames or such that other video transports might have to deal with.

Yes, but it is apparent that the browser may choose to drop the decoding context based on other things, which leads to decoding failures when no key frame is received after such an event.

Currently, in Firefox, this happens as a result of a bug, but I don't see any guarantees in the API that this won't happen under other circumstances. Maybe they'll decide at some point that decoding when the window is off-screen is useless, so they start dropping packets to the decoder?

CendioOssman: I think that sounds unreasonable for the browser. It doesn't know what we intend to do with the data. If it isn't explicitly specified that it can drop data if it feels like it, then we should assume it will do what is requested.

any1:
The requested changes have been made.

CendioOssman (approved):
Great! Thanks!

Did you check if it was possible to generate some data that could be used for unit tests? Like we have for JPEG and PNG rects.

any1:
I added a couple of basic tests for H264Parser and H264Context, but I intend to add more. I hope it's OK to export those. I don't see any other way to test them.

I hardcoded the colour space because it seems that Firefox and Chrome have different defaults. Colour spaces were never specified for the Open H.264 extension as far as I can see. The protocol is missing a way to tell the client what colour space was used for encoding the video. Maybe we can decide a colour space and just say, in the protocol, that the video should always be encoded using that?

CendioOssman:
> I added a couple of basic tests for `H264Parser` and `H264Context`, but I intend to add more. I hope it's OK to export those. I don't see any other way to test them.

They should not require explicit testing, as they are not part of the interface. Can't we feed the data to `H264Decoder`? That's the interface we want a stable behaviour out of.

> I hardcoded the colour space because it seems that Firefox and Chrome have different defaults. Colour spaces were never specified for the Open H.264 extension as far as I can see. The protocol is missing a way to tell the client what colour space was used for encoding the video. Maybe we can decide a colour space and just say, in the protocol, that the video should always be encoded using that?

VNC doesn't explicitly define color space anywhere. But given its nature, sRGB would be the base assumption.

CendioOssman:
> VNC doesn't explicitly define color space anywhere. But given its nature, sRGB would be the base assumption.

That is, assuming the data stream doesn't specify an explicit colour space. In which case, I'd say anything goes. It's up to the client at that point to match things up.

any1:
> > VNC doesn't explicitly define color space anywhere. But given its nature, sRGB would be the base assumption.
> 
> That is, assuming the data stream doesn't specify an explicit colour space. In which case, I'd say anything goes. It's up to the client at that point to match things up.

Well, it's YCbCr encoded, so I don't think sRGB applies.

CendioOssman:
I don't think we have any plans of reusing those components elsewhere, so it's quite sufficient to test only that which we might see from an actual VNC server.

any1:
> I don't think we have any plans of reusing those components elsewhere, so it's quite sufficient to test only that which we might see from an actual VNC server.

I think that we have very different ideas about what unit tests is supposed to achieve, but it's up to you.

any1:
I've added some more tests and I've fixed some bugs that I found when writing them.

Generating good test data was actually a bit of a headache and I was unable to get ffmpeg to do colour conversions correctly so I ended up writing a C program to do it using libavcodec and libswscale. I looked deeper into colour spaces along the way and did my best to set them up correctly in other projects.

I wrote some lower level unit tests to help me figure out some things and I've included them, but I can throw them out if you don't want them. They did help me to find some bugs and to understand what was going wrong with the higher level tests.

The tests don't run on all the CIs, due to lack of WebCodecs support. Do you know if a test file can be skipped depending on whether a feature exists or not?

CendioOssman:
> The tests don't run on all the CIs, due to lack of WebCodecs support. Do you know if a test file can be skipped depending on whether a feature exists or not?

Yes, you can do runtime checks and skip tests. An example here:

https://github.com/novnc/noVNC/blob/c6c8e5e51329001f778a72607bc68432b9d359f8/tests/test.keyboard.js#L206

any1:
Thanks, the test skipping works.

Now it appears that I've exposed a bug in Safari. It doesn't scale from limited range up to full range like it should when rendering the frame to canvas. How should that be handled?

any1:
Well, I managed to make the tests pass on all browsers by carefully crafting the test data. It's now BT.709 full range.

Ideally, the browsers should get this right, but I can't be chasing down bugs in all the browsers.

any1:
Ping?

PR #1881, opened 2024-07-31
Interrupt AltGr sequence detection on focus lost, fixes #1880

### LINKED ISSUE #1880: Keyboard.allKeysUp is not working as expected
**Describe the bug**
`window.onblur` will be triggered before the key down event, if the user hits shortcuts such as Ctrl+T too fast. Thus, the control button is not automatically released.

**To Reproduce**
1. connect to a vnc server
2. hit Ctrl+T as fast as you can
3. switch from new tab back to noVNC tab
4. left control is still down (on both novnc's side and the server side), but the user is not holding it anymore

**Expected behavior**
All keys should be released properly

**Screenshots**
expected behavior
^ expected behavior

onkeyevent down arrived after window.onblur
^ onkeyevent down arrived after window.onblur

 - OS: Windows 11 23H2 build 22631.3880, AMD64
 - Browser:
   - Firefox 128.0.3
   - Google Chrome 126.0.6478.185

 - noVNC version:
   - 1.3.0, Debian packaged 1:1.3.0-2
   - 1.5.0, last release
   - 1230a4ce738ebccce80fc08ee6172bc3a19fadba, master when issuing this bug report
 - VNC server: TigerVNC, Debian packaged 1.13.1+dfsg-3
 - WebSocket proxy: websockify, Debian packaged 0.12.0+dfsg1-2

**Additional context**
I'm not sure if it is a problem of the browser implementation. One possible workaround could be setting a flag to ignore all incoming key events after `window.onblur` and unset it after `window.onfocus`

leedagee:


CendioOssman (approved):
Thanks for a diagnosis and fix!

PR #1888, opened 2024-08-15
Add Zlib encoding

markpeek:
The FreeBSD hypervisor (bhyve) has a built-in VNC server supporting both Raw and Zlib. Adding Zlib encoding to noVNC will reduce the data transferred between the server and client. And will add this encoding for other use cases as well.

### DIFF
--- a/core/decoders/zlib.js
+++ b/core/decoders/zlib.js
@@ -1,0 +1,51 @@
+/*
+ * noVNC: HTML5 VNC client
+ * Copyright (C) 2024 The noVNC Authors
+ * Licensed under MPL 2.0 (see LICENSE.txt)
+ *
+ * See README.md for usage and integration instructions.
+ *
+ */
+
+import Inflator from "../inflator.js";
+
+export default class ZlibDecoder {
+    constructor() {
+        this._zlib = new Inflator();
+        this._length = 0;
+    }
+
+    decodeRect(x, y, width, height, sock, display, depth) {
+        if ((width === 0) || (height === 0)) {
+            return true;
+        }
+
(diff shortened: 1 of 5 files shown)


core/decoders/zlib.js
```diff
@@ -0,0 +1,64 @@
...
+        this._zlib = new Inflator();
+        this._length = 0;
+    }
+
+    decodeRect(x, y, width, height, sock, display, depth) {
+        if ((width === 0) || (height === 0)) {
+            return true;
+        }
+
+        const pixelSize = depth == 8 ? 1 : 4;
```
CendioOssman, line 23: The alternative depth is just for the raw encoding, so you can simplify things here and assume 32-bit.

core/decoders/zlib.js
```diff
@@ -0,0 +1,64 @@
...
+
+        if (sock.rQwait("ZLIB", 4)) {
+            return false;
+        }
```
CendioOssman, line 27: This should strictly speaking be inside the next if block, as you might have already read the length header.

CendioOssman (changes requested):
Thanks! Just some minor cleanup that can be done, and then we can merge this!

core/decoders/zlib.js
markpeek: Fixed, removed from decoder and tests.

core/decoders/zlib.js
markpeek: Thank you for the review. Fixed.

CendioOssman (approved):
Great! Thanks for your contribution!

PR #1892, opened 2024-08-23
Fix sQpushBytes sending the beginning of the array multiple times

Bobini1:
It turns out that pasting large amounts of text didn't work.


CendioOssman (approved):
Thanks!

We actually have a test for this, but apparently it was not a very good test. :/

Issue #350, opened 2014-03-18
Dead keys do not work


samhed:
No dead keys work in Firefox on Linux. By dead keys I mean ~"^`' 
Tested on CentOS 6 and Fedora 18 with both Firefox 26 and 27.

It does work in other browsers such as Google Chrome and there is no problem on Windows (tested with Win8.1).

hex-m:
As a workaround, you can disable dead keys on your client. Doesn't help in a multi-user environment but works for me. (Firefox on Xenial and noVNC on Proxmox 4.1)

CendioOssman:
Unfortunately no, because:

 a) The browser might not be able to provide physical keys in all cases
 b) The server might not support this extension
 c) Sending physical keys is not a silver bullet, but rather a different trade-off. As such some users may prefer the symbol based approach.

danielhb:
I don't think (b) is a problem nowadays - recent VNC servers have been supporting this extension for 5+ years, giving the original QEMU VNC extension was developed in 2010 or so.

But I agree with (a) and (c). I think this bug will have to wait for the solution of the Firefox bug mentioned a few comments above.

CendioOssman:
I don't believe any of the RealVNC derived servers supports this, i.e. TigerVNC, TightVNC, TurboVNC, or the new proprietary RealVNC. So I'd say b) is very much an issue. Is there any server besides QEMU and libvncserver that supports it?

danielhb:
You're right. Disregard my comment about (b).

laggarcia:
> I don't believe any of the RealVNC derived servers supports this, i.e.
> TigerVNC, TightVNC, TurboVNC, or the new proprietary RealVNC. So I'd
> say b) is very much an issue. Is there any server besides QEMU and
> libvncserver that supports it?
>
>
Servers based on gtk-vnc library supports this extension.

CendioOssman:
This is broken in Chrome as well, and likely in all browsers:

https://bugs.chromium.org/p/chromium/issues/detail?id=1135952
https://bugs.chromium.org/p/chromium/issues/detail?id=586568

CendioOssman:
Also reported to the standard folks:

https://github.com/whatwg/dom/issues/900

sergiomb2:
> Doesn't #21 fix this bug too?

I think  it not , I can't write an é (&eacute;)  and all others  accented keys in a novnc session  novnc-1.3.0

sergiomb2:
[...]

PR #1907, opened 2024-10-17
WIP: US-International Keyboard Deadkeys

kwibus:
This proof of concept for Solution 1 for  issues #350 

# Solution 1, guess dead key from event.code etz.
This has the problem this depends on they keyboard layout. which is unknown to the browser.
But:

You only have to do this for Dead Keys, not all keys
there lots of overlap between keyboard layouts, [dead keys list](https://kbdlayout.info/features/deadkeys)
when the overlap, it works for both
If the conflict you can't resolve it
This is not a perfect solution. But you can get it to work for most used keyboard layouts, with deadkeys.
Which might be better the Nothing.

I have a working POC that works for us international keyboard layout with:

XK_dead_grave: `
XK_dead_diaeresis: "
XK_dead_acute: ´
XK_dead_circumflex: ^
XK_dead_tilde: ~

### DIFF
--- a/core/input/util.js
+++ b/core/input/util.js
@@ -175,6 +175,10 @@ export function getKeysym(evt) {
         return DOMKeyTable[key][location];
     }
 
+    if(key === "Dead" ){
+        return getDeadKeysym(evt);
+    }
+
     // Now we need to look at the Unicode symbol instead
 
     // Special key? (FIXME: Should have been caught earlier)
@@ -188,4 +192,32 @@ export function getKeysym(evt) {
     }
 
     return null;
+}
+
+// Try to guess Keysym for Dead key. For now only should work for US-int
+// TODO test
+// TODO try to find and implement Dead keys for more keyboard layouts
+export function getDeadKeysym(evt) {
+    switch(evt.code){
+        case "Quote":
+            if (evt.shiftKey){
+                return KeyTable.XK_dead_diaeresis //
+            } else {
+                return KeyTable.XK_dead_acute // ´
+            }
+        case "Backquote":
+            if (evt.shiftKey){
+                return KeyTable.XK_dead_grave
+            } else {
+                return KeyTable.XK_dead_tilde
+            }
+        case 'Digit6':
+            if (evt.shiftKey){
+                return KeyTable.XK_dead_circumflex
+            }
+
+        default:
+            console.log(evt)
+            return null
+    }
 }
--- a/core/input/keysym.js
+++ b/core/input/keysym.js
@@ -1,6 +1,42 @@
 /* eslint-disable key-spacing */
 
 export default {
+    // TODO remove unused
+    XK_dead_grave:                  0xfe50, // `
+    XK_dead_acute:                  0xfe51, // ´
+    XK_dead_circumflex:             0xfe52, // ^
+    XK_dead_tilde:                  0xfe53, // ~
+    XK_dead_macron:                 0xfe54,
+    XK_dead_breve:                  0xfe55,
+    XK_dead_abovedot:               0xfe56,
+    XK_dead_diaeresis:              0xfe57,
+    XK_dead_abovering:              0xfe58,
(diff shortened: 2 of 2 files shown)


kwibus:
@sergiomb2 Can you give me some more information:

* what did you try?
* which keyboard layout did you use?
* which OS?

So i can try to reproduce the problem

sergiomb2:
I use PT_pt keyboard layout , I have 5 dead keys  only use 4 (`, ´, ~, and  ^)  I changed the following code and work perfectly with Firefox  browser , with google chrome don't 

```
export function getDeadKeysym(evt) {
    switch(evt.code){
        case "BracketRight":
            if (evt.shiftKey){
                return KeyTable.XK_dead_grave
            } else {
                return KeyTable.XK_dead_acute // ´
            }
        case "Backslash":
            if (evt.shiftKey){
                return KeyTable.XK_dead_circumflex
            } else {
                return KeyTable.XK_dead_tilde
            }
```

CendioOssman:
I'm afraid I don't understand the principle here. The code looks to be simply assuming one specific layout as it is just hard-coding what symbol is expected from a specific physical key?

CendioOssman:
Never mind. It seems Safari has firmly said no to that API because of fingerprinting concerns. And Firefox seem to be of the same opinion:

https://bugzilla.mozilla.org/show_bug.cgi?id=1469017

kwibus:
> I'm afraid I don't understand the principle here. The code looks to be simply assuming one specific layout as it is just hard-coding what symbol is expected from a specific physical key?

Yeah that is true, Now it assumes keyboard layout is us-international.
you might be able to expand it a bit. Most keyboard layouts don't have that many dead keys.
But i think its impossible to convert this solution to a generic solution that will work for all keyboard layouts.

kwibus:
I have no longer time to work on this for at least a month. 
I hope can make time for this after that

core/input/util.js
```diff
@@ -175,6 +175,10 @@ export function getKeysym(evt) {
         return DOMKeyTable[key][location];
     }
 
+    if(key === "Dead" ){
+        return getDeadKeysym(evt);
+    }
+
```
sergiomb2, line 181: seems to me that `key==="Dead"` just works in Firefox and not in Chrome .

In firefox I got all accents working perfectly . 

I' like use electron app , because I can use ctrl + key , how we catch dead keys in chrome as it here ?

kwibus:
Strange i can not really reproduce that. for me it works on
* firefox: 131.0
* google chrome: 130.0

I do see a difference in output between linux, and windows. 
and if you check events for text input or not, its different 

but for a non text text input, for linux (wayland) . see the same result for google-chrome as firefox.
Do you get a different result, with my test html? 

I test it with test.html
[test.html.csv](https://github.com/user-attachments/files/17558369/test.html.csv)

I had to add  .csv to be allowed to upload it. so you have to remove the .csv extension.

I test with keyboard layout:

us-international with dead keys. 
I type dead key '  (quote)

chrome
firefox

sergiomb2:
yes , thank you, I got all accents working perfectly on chrome 130.0 , from my chromebook .
And it is awesome, all keya are working , accents , ctrl and alt, ctrl + w  for example

sergiomb2:
Curiously, with electron 31.7.5 based on Chromium v126.0.6478.234, it doesn't work, or rather there are things that seem to already work but in general it doesn't work, it seems to me that this change requires very updated versions ?

Neustradamus:
There is a progress about it?

sergiomb2:
Thank you for reminder, my keyboard_pt.patch works very well on Firefox and Chrome 
[keyboard_pt.patch](https://github.com/user-attachments/files/23676794/keyboard_pt.patch)

I’ve been thinking that maybe in the control bar you could add a keyboard-layout selector,, with the various languages that are configured. For example, there would be a PT option to use the layout I sent in patch, and later maybe automate the keyboard selection based on the layout the client’s browser is using.

PR #1908, opened 2024-10-24
Add native touch gestures support with UltraVNC

EvilAngel00:
This PR aims to start a discussion about alternative ways of managing gestures. 

It is an implementation of a gesture handler that sends touches to an UltraVNC server (Windows only) via the GII pseudo-encoding. It allows for multi-touch, zooming, panning and scrolling while using the native Windows touch events instead of sending mouse events. At the moment, it can be activated via a checkbox in the noVNC settings before establishing a connection and during the handshake, the GII pseudo-encoding is added to the list of encodings sent to the server. Touches are then managed internally with an array that keeps track of the touch status (touchstart, touchmove or touchend/touchcancel) while sending GII messages every time there is a change. Touch events contain an identifier to determine which pointer is being moved, this identifier is given by the browser and is handled differently on Chromium based browsers (Edge, Chrome) and Firefox. Chromium browsers assign the smallest possible ID while Firefox keeps assigning incrementally i.e. if I place three fingers down one after the other, lift the second one and place it back I would have an array of IDs:

[0, 1, 2] ->[0, 2] -> [0, 2, 1] on Chromium
[0, 1, 2] -> [0, 2] -> [0, 2, 3] on Firefox

This caused issues when sending higher ID numbers to UltraVNC and we have to manage those IDs manually instead.

It could be interesting to have a framework in place to allow people to have their own implementation of touch events and gestures depending on the support from the VNC server being used. I do not intend to expand this at the moment, this is more of a starting point for the potential for different types of gestures.

In any case, thank you for work on this library!


CendioOssman:
Thanks for your contribution! You haven't gone unnoticed. We just haven't had time to look at this PR yet.

CendioOssman (changes requested):
I have not had time to look at this in detail yet, but a casual look shows that everything is labeled "ultravnc" even though it implements the GII extension.

Could you go over things and try to follow a naming that matches whats documented in rfbproto.

There were also a bunch of magical constants. We should create constants for these so things are more readable, just like we have for other protocol stuff at the top of rfb.js.

EvilAngel00:
Hi Pierre, thanks for looking into the PR!

The reason why everything is labeled like that is because this GII implementation is specific to UltraVNC. I added the GII pseudo-encoding to the list of sent encodings when the option is set but the rest (setting up the GII device, sending the touches, ...) are based on what UltraVNC does with their own viewer.
What I would suggest then, instead of having a checkbox in the UI, is to have a dropdown with the choice to use the noVNC implementation of gestures or this UltraVNC specific one. This option would then be passed to the RFB object and the constructor would select the appropriate gesture handler. Like that, it can then be extended as people could just add their own implementation of gestures depending on their needs. What do you think?

I will also solve the merge conflicts and set the constants as you suggested.

EvilAngel00:
I made the changes but rebasing the main branch into mine didn't work very well with the PR 😖

I can make a new PR for this if you'd prefer.

EvilAngel00:
Sorry for the delay on this, I've finally fixed the mess of rebasing with a merge instead.

I've implemented the requested changes but a lot of these numbers come directly from the UltraVNC implementation of GII. It could be nice to get the conversation started again on the potential for custom gesture modes.

EvilAngel00:
Hello noVNC team! Any update/feedback on this PR?

CendioOssman:
Unfortunately, no one has had any spare time to look at this yet. The GII extension is underdocumented and poorly understood, which means it takes a bit more effort to check this.

It's still on the todo list, so you're not forgotten.

PR #1916, opened 2024-11-22
Allow very small twodrag and pinch gestures

samhed:
Before this commit, we strictly required the user to make a ~50 pixel move in order to get these two gestures. We can get a more precise experience if we allow very small gestures as well. This can be achieved by making better use of our "_twoTouchTimeout".

Our two touch timeout handling was written to handle this well anyway. We can distinguish between drag and pinch using the angle.

This is a prerequisite for browser-side gestures like in #1809 to work well.

### DIFF
--- a/core/input/gesturehandler.js
+++ b/core/input/gesturehandler.js
@@ -186,14 +186,14 @@ export default class GestureHandler {
         }
 
         if (!this._hasDetectedGesture()) {
-            // Ignore moves smaller than the minimum threshold
-            if (Math.hypot(deltaX, deltaY) < GH_MOVE_THRESHOLD) {
-                return;
-            }
 
-            // Can't be a tap or long press as we've seen movement
-            this._state &= ~(GH_ONETAP | GH_TWOTAP | GH_THREETAP | GH_LONGPRESS);
-            this._stopLongpressTimeout();
+            let deltaMove = Math.hypot(deltaX, deltaY);
+
+            // Can't be a tap or long press if we've seen movement
+            if (deltaMove >= GH_MOVE_THRESHOLD) {
+                this._state &= ~(GH_ONETAP | GH_TWOTAP | GH_THREETAP | GH_LONGPRESS);
+                this._stopLongpressTimeout();
+            }
 
             if (this._tracked.length !== 1) {
                 this._state &= ~(GH_DRAG);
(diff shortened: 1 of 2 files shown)


samhed:
This also makes things easier for things like #1835

core/input/gesturehandler.js
```diff
@@ -186,14 +186,14 @@ export default class GestureHandler {
...
+            let deltaMove = Math.hypot(deltaX, deltaY);
+
+            // Can't be a tap or long press if we've seen movement
+            if (deltaMove >= GH_MOVE_THRESHOLD) {
+                this._state &= ~(GH_ONETAP | GH_TWOTAP | GH_THREETAP | GH_LONGPRESS);
+                this._stopLongpressTimeout();
+            }
 
             if (this._tracked.length !== 1) {
                 this._state &= ~(GH_DRAG);
```
CendioOssman, line 199: This should have already been excluded in the touchstart handler. Perhaps best to remove this redundancy to make things easier to read and understand.

core/input/gesturehandler.js
```diff
@@ -186,14 +186,14 @@ export default class GestureHandler {
         }
 
         if (!this._hasDetectedGesture()) {
-            // Ignore moves smaller than the minimum threshold
-            if (Math.hypot(deltaX, deltaY) < GH_MOVE_THRESHOLD) {
-                return;
-            }
```
CendioOssman, line 189: This breaks two finger gestures if the first finger moves slightly before the second finger makes its initial touch. Which I'd suspect is very likely, given how sensitive touch devices are.

core/input/gesturehandler.js
```diff
@@ -234,7 +234,8 @@ export default class GestureHandler {
                     }
                 } else if (!this._isTwoTouchTimeoutRunning()) {
                     // We can't determine the gesture right now, let's
-                    // wait and see if more events are on their way
+                    // wait and see if more events are on their way.
+                    // If not, we'll have to decide which gesture it is.
                     this._startTwoTouchTimeout();
```
CendioOssman, line 239: This breaks other two finger gestures as the two touch timeout blindly assumes that only a pinch or drag is left to choose from. It might need to be fixed so it also just excludes things, leaving it still uncertain for any other gestures that have yet to be ruled out.

I'm not sure if this timeout was ever really used, or at least very little, given the existing large movement threshold. We might need to reexamine it in detail.

CendioOssman: The lack of distance threshold also means that the timeout handler has much more noisy data to deal with when trying to identify a pinch from a drag. So it might be much less robust there as well.

CendioOssman (changes requested):
I think some more design work is needed, as this seems to cause some gestures to become very unreliable. The threshold was pretty fundamental in the existing design to separate things.

Issue #1918, opened 2024-11-27
Support mouse back and forward buttons


CendioOssman:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to use my mouse to go back and forward in my browser inside my VNC session. Currently that makes my local browser go back and forward instead.

**Describe the solution you'd like**
See above.

**Describe alternatives you've considered**
None

**Additional context**
None

CendioOssman:
Issue reported to Firefox:

https://bugzilla.mozilla.org/show_bug.cgi?id=1933746

CendioOssman:
Server-side support for this is available in development versions of TigerVNC:

https://github.com/TigerVNC/tigervnc/pull/1826

samhed:
Fixed as of #1919.

Not much we can do about Firefox, opened #1930 to keep track of the state in that browser.

PR #1919, opened 2024-11-28
Add support for forward and back mouse buttons

CendioHalim:
This commit implements the extendedMouseButtons pseudo-encoding, which makes it possible to use the forward and back mouse buttons.

It looks like chromium is the only browser that has support for this currently.  See #1918 for more info.

Confirmed that it works with chromium on Mac, Linux and Windows against TigerVNC server.

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -152,6 +152,8 @@ export default class RFB extends EventTargetMixin {
 
         this._qemuExtKeyEventSupported = false;
 
+        this._extendedPointerEventSupported = false;
+
         this._clipboardText = null;
         this._clipboardServerCapabilitiesActions = {};
         this._clipboardServerCapabilitiesFormats = {};
@@ -1051,10 +1053,11 @@ export default class RFB extends EventTargetMixin {
             1: 1 << 2, // Right
             2: 1 << 1, // Middle
             3: 1 << 7, // Back
+            4: 1 << 8, // Forward
         };
 
         let bmask = 0;
-        for (let i = 0; i < 4; i++) {
+        for (let i = 0; i < 5; i++) {
             if (buttons & (1 << i)) {
                 bmask |= buttonMaskMap[i];
             }
@@ -1189,8 +1192,20 @@ export default class RFB extends EventTargetMixin {
         if (this._rfbConnectionState !== 'connected') { return; }
         if (this._viewOnly) { return; } // View only, skip mouse events
 
-        RFB.messages.pointerEvent(this._sock, this._display.absX(x),
-                                  this._display.absY(y), mask);
+        // Highest bit in mask is never sent to the server
+        if (mask & 0x8000) {
+            throw new Error("Illegal mouse button mask (mask: " + mask + ")");
+        }
+
+        let extendedMouseButtons = mask & 0x7f80;
+
+        if (this._extendedPointerEventSupported && extendedMouseButtons) {
+            RFB.messages.extendedPointerEvent(this._sock, this._display.absX(x),
+                                              this._display.absY(y), mask);
+        } else {
+            RFB.messages.pointerEvent(this._sock, this._display.absX(x),
+                                      this._display.absY(y), mask);
+        }
     }
 
     _handleWheel(ev) {
@@ -2229,6 +2244,7 @@ export default class RFB extends EventTargetMixin {
         encs.push(encodings.pseudoEncodingContinuousUpdates);
         encs.push(encodings.pseudoEncodingDesktopName);
         encs.push(encodings.pseudoEncodingExtendedClipboard);
+        encs.push(encodings.pseudoEncodingExtendedMouseButtons);
 
         if (this._fbDepth == 24) {
             encs.push(encodings.pseudoEncodingVMwareCursor);
@@ -2658,6 +2674,10 @@ export default class RFB extends EventTargetMixin {
             case encodings.pseudoEncodingExtendedDesktopSize:
                 return this._handleExtendedDesktopSize();
 
+            case encodings.pseudoEncodingExtendedMouseButtons:
+                this._extendedPointerEventSupported = true;
+                return true;
+
             case encodings.pseudoEncodingQEMULedEvent:
                 return this._handleLedEvent();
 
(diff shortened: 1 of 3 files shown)


CendioHalim:
This MR should partly fix #1920, but only on chromium.

core/rfb.js
```diff
@@ -1060,15 +1062,27 @@ export default class RFB extends EventTargetMixin {
         let pos = clientToElement(ev.clientX, ev.clientY,
                                   this._canvas);
 
+        /* Map mouse back and forward mouse buttons (3 and 4) to 7 and 8
+         *
+         * NOTE: This only works on chromium-based browsers. There is
+         *       no support for firefox/safari.
```
CendioOssman, line 1068: Strictly speaking, Safari has buggy support rather than no support. Right?

tests/test.rfb.js
```diff
@@ -4931,7 +4931,14 @@ describe('RFB messages', function () {
         it('should send correct data for pointer events', function () {
             RFB.messages.pointerEvent(sock, 12345, 54321, 0xab);
             let expected =
-                [ 5, 0xab, 0x30, 0x39, 0xd4, 0x31];
+                [ 5, 0x2b, 0x30, 0x39, 0xd4, 0x31];
+            expect(sock).to.have.sent(new Uint8Array(expected));
+        });
```
CendioOssman, line 4936: The masking should probably be tested as an explicitly different test.

CendioOssman (changes requested):

core/rfb.js
CendioHalim: Correct.

I just realized that safari send the incorrect `ev.button` value, but the button mask `ev.buttons` seems to work correctly. With back pressed, `ev.buttons` is 8, and with forward pressed, it is 16.

We could look at those instead of keeping mouse button state ourselves?

CendioHalim:
With #1921 merged, this PR has been rebased and should now work with both chromium-based browsers and safari.

tests/test.rfb.js
```diff
@@ -3265,6 +3265,11 @@ describe('Remote Frame Buffer protocol client', function () {
                     expect(spy).to.have.been.calledOnce;
                     expect(spy.args[0][0].detail.name).to.equal('som€ nam€');
                 });
+
+                it('should handle the extendedMouseButtons pseudo-encoding', function () {
+                    sendFbuMsg([{ x: 0, y: 0, width: 0, height: 0, encoding: -316 }], [[]], client);
+                    expect(client._extendedPointerEventSupported).to.equals(true);
```
CendioOssman, line 3271: We do not want to look at internal stuff. We want to see changes in externally observed behaviour. I.e., what pointer events it will send.

tests/test.rfb.js
```diff
@@ -3884,6 +3892,26 @@ describe('Remote Frame Buffer protocol client', function () {
                                                                  50, 70, 0x0);
             });
 
+            it('should send extended pointer event when server supports extended pointer events', function () {
+                client._extendedPointerEventSupported = true;
```
CendioOssman, line 3896: Same thing here. Don't fiddle with internal stuff. Do the necessary exchange.

tests/test.rfb.js
```diff
@@ -3884,6 +3892,26 @@ describe('Remote Frame Buffer protocol client', function () {
...
+                client._extendedPointerEventSupported = false;
+                sendMouseButtonEvent(50, 70, true, 0x10, client);
+
+                expect(pointerEvent).to.have.been.calledOnceWith(client._sock,
+                                                                 50, 70, 0x100);
+            });
+
+            it('should not send extended pointer event with illegal mask', function () {
+                expect(() => RFB.messages.extendedPointerevent(client._sock, 50, 70, 0xfe00)).to.throw(Error);
+            });
```
CendioOssman, line 3913: I thought you meant to remove this?

tests/test.rfb.js
```diff
@@ -3884,6 +3892,26 @@ describe('Remote Frame Buffer protocol client', function () {
...
 
+            it('should send extended pointer event when server supports extended pointer events', function () {
+                client._extendedPointerEventSupported = true;
+                sendMouseButtonEvent(50, 70, true, 0x10, client);
+
+                expect(extendedPointerEvent).to.have.been.calledOnceWith(client._sock,
+                                                                         50, 70, 0x100);
+            });
+
+            it('should send normal pointer event when server supports does not extended pointer events', function () {
```
CendioOssman, line 3903: Should also have a separate test that shows that buttons are filtered out for old servers.

CendioOssman (changes requested):

tests/test.rfb.js
CendioHalim: Whoops, this should have been removed

tests/test.rfb.js
```diff
@@ -3265,6 +3265,18 @@ describe('Remote Frame Buffer protocol client', function () {
                     expect(spy).to.have.been.calledOnce;
                     expect(spy.args[0][0].detail.name).to.equal('som€ nam€');
                 });
+
+                it('should handle the extendedMouseButtons pseudo-encoding', function () {
```
CendioOssman, line 3269: This test is redundant and a bit out of place now that the other tests properly send a pseudorect.

tests/test.rfb.js
```diff
@@ -3884,6 +3899,34 @@ describe('Remote Frame Buffer protocol client', function () {
...
+
+            it('should filter extended buttons when server does not support extended pointer events', function () {
+                sendMouseButtonEvent(50, 70, true, 0x11, client);
+
+                expect(pointerEvent).to.have.been.calledOnceWith(client._sock,
+                                                                 50, 70, 0x101);
+
+                let expected =
+                    [ 5, 0x1, 0x0, 0x32, 0x0, 0x46];
+                expect(client._sock).to.have.sent(new Uint8Array(expected));
```
CendioOssman, line 3927: This got a bit odd. But I guess it's because it's actually the `pointerEvent()` function that does the filtering? If so, then it's probably best to have this filtering test with the other tests for that function. Which I guess you already have?

CendioOssman (changes requested):

tests/test.rfb.js
CendioHalim: > This got a bit odd. But I guess it's because it's actually the pointerEvent() function that does the filtering? I

Correct, the filtering is done in `pointerEvent()`

> If so, then it's probably best to have this filtering test with the other tests for that function. Which I guess you already have?

Kind of, there is one `pointerEvent()` test that checks that the marker bit gets zeroed. I'll modify that test so that it checks that extended bits are filtered as well and remove this one.

CendioOssman (approved):

PR #1921, opened 2024-11-29
Use MouseEvent.buttons for button state tracking

CendioHalim:
Instead of keeping track of button states ourselves by looking at `MouseEvent.button`, we can use the `MouseEvent.buttons` which already contains the state of all buttons.

Note that in `MouseEvent.button`,  button numbers are mapped as follows:
```
0 -> Left
1 -> Middle
2 -> Right
3 -> Back
4 -> Forward
```
But when looking at the `MouseEvent.buttons` property, the bits middle and right are switched, which is confusing:
```
0 -> Left
1 -> Right
2 -> Middle
3 -> Back
4 -> Forward
```

With this change, #1919 could be modified to work with Safari as well.

Unfortunately, the internal `_mouseButtonStateMask` state is not removed in this PR as it would require a large rewrite of the general architecture.

Tested on:
* Linux
  * Firefox
  * Chrome
  * Epiphany
* Windows 11
  * Edge
  * Chrome
  * Firefox
* macOS
  * Safari
  * Edge
  * Chrome
  * Firefox
* iPad
  * Safari

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -1033,6 +1033,35 @@ export default class RFB extends EventTargetMixin {
         this.sendKey(keysym, code, down);
     }
 
+    static _convertButtonMask(buttons) {
+        /* The bits in MouseEvent.buttons property correspond
+         * to the following mouse buttons:
+         *     0: Left
+         *     1: Right
+         *     2: Middle
+         *     3: Back
+         *     4: Forward
+         *
+         * These bits needs to be converted to what they are defined as
+         * in the RFB protocol.
+         */
+
+        const buttonMaskMap = {
+            0: 1 << 0, // Left
+            1: 1 << 2, // Right
+            2: 1 << 1, // Middle
+            3: 1 << 7, // Back
+        };
+
+        let bmask = 0;
+        for (let i = 0; i < 4; i++) {
+            if (buttons & (1 << i)) {
+                bmask |= buttonMaskMap[i];
+            }
+        }
+        return bmask;
+    }
+
     _handleMouse(ev) {
         /*
          * We don't check connection status or viewOnly here as the
@@ -1062,76 +1091,73 @@ export default class RFB extends EventTargetMixin {
         let pos = clientToElement(ev.clientX, ev.clientY,
                                   this._canvas);
 
+        let bmask = RFB._convertButtonMask(ev.buttons);
+
+        let down = ev.type == 'mousedown';
         switch (ev.type) {
             case 'mousedown':
-                setCapture(this._canvas);
-                this._handleMouseButton(pos.x, pos.y,
-                                        true, 1 << ev.button);
-                break;
             case 'mouseup':
-                this._handleMouseButton(pos.x, pos.y,
-                                        false, 1 << ev.button);
+                if (this.dragViewport) {
+                    if (down && !this._viewportDragging) {
+                        this._viewportDragging = true;
+                        this._viewportDragPos = {'x': pos.x, 'y': pos.y};
+                        this._viewportHasMoved = false;
+
+                        // Skip sending mouse events, instead save the current
+                        // mouse mask so we can send it later.
+                        this._mouseButtonMask = bmask;
+                        break;
+                    } else {
+                        this._viewportDragging = false;
+
+                        // If we actually performed a drag then we are done
+                        // here and should not send any mouse events
+                        if (this._viewportHasMoved) {
+                            this._mouseButtonMask = bmask;
+                            break;
+                        }
+                        // Otherwise we treat this as a mouse click event.
+                        // Send the previously saved button mask, followed
+                        // by the current button mask at the end of this
+                        // function.
+                        this._sendMouse(pos.x, pos.y,  this._mouseButtonMask);
+                    }
+                }
+                if (down) {
+                    setCapture(this._canvas);
+                }
+                this._handleMouseButton(pos.x, pos.y, bmask);
                 break;
             case 'mousemove':
+                if (this._viewportDragging) {
+                    const deltaX = this._viewportDragPos.x - pos.x;
+                    const deltaY = this._viewportDragPos.y - pos.y;
+
+                    if (this._viewportHasMoved || (Math.abs(deltaX) > dragThreshold ||
+                                                   Math.abs(deltaY) > dragThreshold)) {
+                        this._viewportHasMoved = true;
+
+                        this._viewportDragPos = {'x': pos.x, 'y': pos.y};
+                        this._display.viewportChangePos(deltaX, deltaY);
+                    }
+
+                    // Skip sending mouse events
+                    break;
+                }
                 this._handleMouseMove(pos.x, pos.y);
                 break;
         }
     }
 
-    _handleMouseButton(x, y, down, bmask) {
-        if (this.dragViewport) {
-            if (down && !this._viewportDragging) {
-                this._viewportDragging = true;
-                this._viewportDragPos = {'x': x, 'y': y};
-                this._viewportHasMoved = false;
-
-                // Skip sending mouse events
-                return;
-            } else {
-                this._viewportDragging = false;
-
-                // If we actually performed a drag then we are done
-                // here and should not send any mouse events
-                if (this._viewportHasMoved) {
-                    return;
-                }
-
-                // Otherwise we treat this as a mouse click event.
-                // Send the button down event here, as the button up
-                // event is sent at the end of this function.
-                this._sendMouse(x, y, bmask);
-            }
-        }
-
+    _handleMouseButton(x, y, bmask) {
         // Flush waiting move event first
         this._flushMouseMoveTimer(x, y);
 
-        if (down) {
-            this._mouseButtonMask |= bmask;
-        } else {
-            this._mouseButtonMask &= ~bmask;
-        }
-
+        this._mouseButtonMask = bmask;
         this._sendMouse(x, y, this._mouseButtonMask);
     }
 
     _handleMouseMove(x, y) {
-        if (this._viewportDragging) {
-            const deltaX = this._viewportDragPos.x - x;
-            const deltaY = this._viewportDragPos.y - y;
-
-            if (this._viewportHasMoved || (Math.abs(deltaX) > dragThreshold ||
-                                           Math.abs(deltaY) > dragThreshold)) {
-                this._viewportHasMoved = true;
-
-                this._viewportDragPos = {'x': x, 'y': y};
-                this._display.viewportChangePos(deltaX, deltaY);
-            }
-
-            // Skip sending mouse events
-            return;
-        }
-
         this._mousePos = { 'x': x, 'y': y };
 
         // Limit many mouse move events to one every MOUSE_MOVE_DELAY ms
@@ -1175,6 +1201,7 @@ export default class RFB extends EventTargetMixin {
         let pos = clientToElement(ev.clientX, ev.clientY,
                                   this._canvas);
 
+        let bmask = RFB._convertButtonMask(ev.buttons);
         let dX = ev.deltaX;
         let dY = ev.deltaY;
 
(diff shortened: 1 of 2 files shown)


core/rfb.js
```diff
@@ -1108,13 +1108,8 @@ export default class RFB extends EventTargetMixin {
...
 
-        if (down) {
-            this._mouseButtonMask |= bmask;
-        } else {
-            this._mouseButtonMask &= ~bmask;
-        }
-
-        this._sendMouse(x, y, this._mouseButtonMask);
+        this._sendMouse(x, y, bmask);
+        this._mouseButtonMask = bmask;
```
CendioOssman, line 1112: If you change the order of these two, then you can retain that `_sendMouse()` is always called with `this._mouseButtonMask`.

core/rfb.js
```diff
@@ -1200,22 +1195,22 @@ export default class RFB extends EventTargetMixin {
         // for one of the axes is large enough.
         if (Math.abs(this._accumulatedWheelDeltaX) >= WHEEL_STEP) {
             if (this._accumulatedWheelDeltaX < 0) {
-                this._handleMouseButton(pos.x, pos.y, true, 1 << 5);
-                this._handleMouseButton(pos.x, pos.y, false, 1 << 5);
+                this._handleMouseButton(pos.x, pos.y, true, this._mouseButtonMask | 1 << 5);
+                this._handleMouseButton(pos.x, pos.y, false, this._mouseButtonMask &  ~(1 << 5));
```
CendioOssman, line 1199: Do we really need the unset of the bit? Shouldn't this method be in full control of the wheel button bits?

I feel it might imply that the wheel bits might be set in other places as well.

core/rfb.js
```diff
@@ -1060,15 +1060,15 @@ export default class RFB extends EventTargetMixin {
...
                                   this._canvas);
 
+        let bmask = RFB.convertButtonMask(ev.buttons);
+
         switch (ev.type) {
             case 'mousedown':
                 setCapture(this._canvas);
-                this._handleMouseButton(pos.x, pos.y,
-                                        true, 1 << ev.button);
+                this._handleMouseButton(pos.x, pos.y, true, bmask);
```
CendioOssman, line 1068: It's a bit weird keeping the `down` argument now that we have a full mask. But that's a cleanup I suppose we can do later.

core/rfb.js
```diff
@@ -2941,6 +2936,35 @@ export default class RFB extends EventTargetMixin {
             "raw", passwordChars, { name: "DES-ECB" }, false, ["encrypt"]);
         return legacyCrypto.encrypt({ name: "DES-ECB" }, key, challenge);
     }
+
+    static convertButtonMask(buttons) {
```
CendioOssman, line 2940: This is an internal method, so it should have a `_` prefix.

CendioOssman, line 2940: Given that it's an internal function, as is `_handleMouse()`, can't we put them next to each other to make it easier to read?

tests/test.rfb.js
```diff
@@ -3508,6 +3508,9 @@ describe('Remote Frame Buffer protocol client', function () {
             // client coordinate calculations
             client.focusOnClick = false;
 
+            // We need to keep track of MouseEvent.buttons state
+            client._buttonsState = 0;
```
CendioOssman, line 3512: This feels like it obfuscating a lot of what the tests are supposed to do. Can't we make sure the tests provide this?

CendioOssman (changes requested):

CendioHalim:
I've updated my branch and addressed all comments

core/rfb.js
```diff
@@ -1097,7 +1126,7 @@ export default class RFB extends EventTargetMixin {
                 // Otherwise we treat this as a mouse click event.
                 // Send the button down event here, as the button up
                 // event is sent at the end of this function.
-                this._sendMouse(x, y, bmask);
+                this._sendMouse(x, y, this._mouseButtonMask);
```
CendioOssman, line 1129: I don't think the comment here has been considered. Can you still click in place with dragging active?

I guess that means we are lacking at least one unit test.

CendioOssman: This drag handling might be on the wrong layer. Perhaps it should be moved up one level in the stack.

core/rfb.js
```diff
@@ -1196,26 +1221,27 @@ export default class RFB extends EventTargetMixin {
...
 
+
         // Generate a mouse wheel step event when the accumulated delta
         // for one of the axes is large enough.
         if (Math.abs(this._accumulatedWheelDeltaX) >= WHEEL_STEP) {
             if (this._accumulatedWheelDeltaX < 0) {
-                this._handleMouseButton(pos.x, pos.y, true, 1 << 5);
-                this._handleMouseButton(pos.x, pos.y, false, 1 << 5);
+                this._handleMouseButton(pos.x, pos.y, true, bmask | 1 << 5);
+                this._handleMouseButton(pos.x, pos.y, false, bmask);
```
CendioOssman, line 1230: Looks good. Do we have unit tests that make sure that buttons + wheel work at the same time?

core/rfb.js
```diff
@@ -1307,33 +1333,36 @@ export default class RFB extends EventTargetMixin {
                     case 'longpress':
                         this._fakeMouseMove(ev, pos.x, pos.y);
                         break;
-                    case 'twodrag':
+                    case 'twodrag': {
+                        let bmask = RFB._convertButtonMask(ev.detail.buttons);
```
CendioOssman, line 1337: There is no `buttons` attribute for gesture events. I'm concerned that this passes the unit tests. Something must be missing/wrong there.

core/rfb.js
```diff
@@ -1307,33 +1333,36 @@ export default class RFB extends EventTargetMixin {
...
                         // Always scroll in the same position.
                         // We don't know if the mouse was moved so we need to move it
                         // every update.
                         this._fakeMouseMove(ev, pos.x, pos.y);
                         while ((ev.detail.magnitudeY - this._gestureLastMagnitudeY) > GESTURE_SCRLSENS) {
-                            this._handleMouseButton(pos.x, pos.y, true, 0x8);
-                            this._handleMouseButton(pos.x, pos.y, false, 0x8);
+                            this._handleMouseButton(pos.x, pos.y, true, bmask | 0x8);
```
CendioOssman, line 1343: You should probably rather keep using the known mouse state. I.e. `this._mouseButtonMask`.

core/rfb.js
```diff
@@ -1307,33 +1333,36 @@ export default class RFB extends EventTargetMixin {
...
+                            this._handleMouseButton(pos.x, pos.y, false, 0x0);
```
CendioOssman, line 1344: Same thing here.

CendioOssman (changes requested):

core/rfb.js
CendioHalim: Looks like most dragging unit tests don't check which button masks are sent. They only check how many times `RFB.messages.pointerEvent` has been called.

Adding this breaks the test (0x0 is sent twice):

```diff --git a/tests/test.rfb.js b/tests/test.rfb.js
index 65c84c9..c2856b5 100644
--- a/tests/test.rfb.js
+++ b/tests/test.rfb.js
@@ -639,9 +639,19 @@ describe('Remote Frame Buffer protocol client', function () {
 
             it('should send button messages when release without movement', function () {
                 // Just up and down
-                client._handleMouseButton(13, 9, 0x001);
-                client._handleMouseButton(13, 9, 0x000);
+                client._handleMouseButton(13, 9, 0x001, 0x1);
+
+                expect(RFB.messages.pointerEvent).to.not.have.been.called;
+
+                client._handleMouseButton(13, 9, 0x000, 0x0);
+
                 expect(RFB.messages.pointerEvent).to.have.been.calledTwice;
+                expect(RFB.messages.pointerEvent.firstCall).to.have.been
+                                                 .calledWith(client._sock,
+                                                             13, 9, 0x1);
+                expect(RFB.messages.pointerEvent.secondCall).to.have.been
+                                                 .calledWith(client._sock,
+                                                             13, 9, 0x0);
```

core/rfb.js
CendioHalim: Seems like `ev.detail.<whatever>` returns 0 for some reason?

core/rfb.js
CendioHalim: Doesn't look like it. I'll add some tests for that.

core/rfb.js
```diff
@@ -1313,23 +1339,23 @@ export default class RFB extends EventTargetMixin {
                         // every update.
                         this._fakeMouseMove(ev, pos.x, pos.y);
                         while ((ev.detail.magnitudeY - this._gestureLastMagnitudeY) > GESTURE_SCRLSENS) {
-                            this._handleMouseButton(pos.x, pos.y, true, 0x8);
-                            this._handleMouseButton(pos.x, pos.y, false, 0x8);
+                            this._handleMouseButton(pos.x, pos.y, true, this._mouseButtonMask | 0x8);
+                            this._handleMouseButton(pos.x, pos.y, false, this._mouseButtonMask & ~0x8);
```
CendioOssman: Not just `this._mouseButtonMask` without the `& ~`?

It's a bit of a corner case where desired behaviour isn't clear, but it's more consistent with the rest of the code.

CendioOssman: Or just drop `this._mouseButtonMask` altogether? Do we need to support mixing mouse and touch?

tests/test.rfb.js
```diff
@@ -3687,60 +3687,69 @@ describe('Remote Frame Buffer protocol client', function () {
...
-                sendMouseMoveEvent(10, 10, client);
+
+                sendMouseMoveEvent(10, 10, 0x0, client);
                 clock.tick(50);
                 expect(pointerEvent).to.not.have.been.called;
             });
 
             it('should handle left mouse button', function () {
-                sendMouseButtonEvent(10, 10, true, 0, client);
+
```
CendioOssman, line 3748: Some stray newlines here.

tests/test.rfb.js
```diff
@@ -4030,6 +4031,21 @@ describe('Remote Frame Buffer protocol client', function () {
                 expect(pointerEvent.secondCall).to.have.been.calledWith(client._sock,
                                                                         10, 10, 0);
             });
+
+            it('should handle wheel event with buttons pressed', function () {
```
CendioOssman, line 4078: Can this be moved earlier? So we can see that your changes just kept existing behaviour.

tests/test.rfb.js
```diff
@@ -768,6 +768,26 @@ describe('Remote Frame Buffer protocol client', function () {
                                                                                     13, 9, 0x0);;
             });
 
+            it('should send button messages when longpress with small movement', function () {
```
CendioOssman, line 771: Is this a changed behaviour? Otherwise, I think they should also be earlier to indicate you are keeping existing behaviour.

core/rfb.js
```diff
@@ -1091,46 +1091,61 @@ export default class RFB extends EventTargetMixin {
...
 
+        let down = ev.type == 'mousedown';
         switch (ev.type) {
             case 'mousedown':
-                setCapture(this._canvas);
-                this._handleMouseButton(pos.x, pos.y, true, bmask);
-                break;
             case 'mouseup':
-                this._handleMouseButton(pos.x, pos.y, false, bmask);
+                if (this.dragViewport) {
```
CendioOssman, line 1098: Did dragging still work before this commit? Or did it fix breakage introduced by switching to a button mask?

If it's the latter, then ideally this change should be moved earlier so that every commit works.

CendioOssman (changes requested):
Really nice improvements to the tests! But I have some questions on the new changes.

core/rfb.js
CendioHalim: `_handleMouseButton()` mutates `this._mouseButtonMask`, so we have to do this.

tests/test.rfb.js
```diff
@@ -740,6 +754,20 @@ describe('Remote Frame Buffer protocol client', function () {
                                                                                      15, 14, 0x0);
             });
 
+            it('should send button messages when tapping with small movement', function () {
+                // Small movement
+                gestureStart('onetap', 13, 9, client);
+                gestureEnd('onetap', 15, 14, client);
```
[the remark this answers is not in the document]
CendioHalim: I was able to get slightly different coordinates between a 'gesturestart' and 'gestureend' when testing. Nonetheless, might as well change the test to send the same coordinates to minimize confusion.

core/rfb.js
CendioHalim: Dragging worked before this commit. All behaviours should be identical, except for the changes introduced in 8453c311cd2125ccaa9f6d6e5dc9faac00c411dd

tests/test.rfb.js
```diff
@@ -787,6 +815,50 @@ describe('Remote Frame Buffer protocol client', function () {
...
             });
 
+            it('should initiate viewport dragging on sufficient drag gesture movement', function () {
+                sinon.spy(client._display, "viewportChangePos");
+
+                // Sufficient movement
+                gestureStart('drag', 13, 9, client);
+                gestureMove('drag', 43, 9, client);
+
+                expect(RFB.messages.pointerEvent).to.have.been.calledOnce;
```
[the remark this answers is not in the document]
CendioHalim: I've moved the test so that it gets added when this is fixed so that we don't have tests that test "undesired" behaviour.

PR #1923, opened 2024-12-12
remove sound, control bar

sravan-abacus:


samhed:
This is a feature which some people rely on. Since you didn't give an explanation or reasoning for removing it, I will just close this.

PR #1924, opened 2024-12-14
Update Japanese translation

nnn1590:
Add translations for `Running without HTTPS is not recommended, crashes or other issues are likely.` and `Failed to connect to server: `


CendioOssman:
Thanks! Merged.

PR #1926, opened 2025-01-13
Slight UI refresh

samhed:
This PR mainly includes a refreshed input.css:

 * No more gradients
 * Larger input elements
 * Clear visual difference between buttons and text inputs
 * New styles for indicating pressed state (:active)

But also a few changes to base.css & vnc.html:

 * More spacing between elements
 * Toggle switches instead of checkboxes in settings

### DIFF
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -456,6 +456,7 @@ html {
 .noVNC_panel hr {
     border: none;
     border-top: 1px solid var(--novnc-lightgrey);
+    width: 100%; /* <hr> inside a flexbox will otherwise be 0px wide */
 }
 
 .noVNC_panel label {
@@ -493,8 +494,20 @@ html {
     vertical-align: bottom;
 }
 
-.noVNC_submit {
-    float: right;
+.noVNC_panel form {
+    display: flex;
+    flex-direction: column;
+    gap: 12px
+}
+
+.noVNC_panel .button_row {
+    margin-top: 10px;
+    display: flex;
+    gap: 10px;
+    justify-content: space-between;
+}
+.noVNC_panel .button_row *:only-child {
+    margin-left: auto; /* Align single buttons to the right */
 }
 
 /* Expanders */
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -729,13 +729,10 @@ html {
 }
 #noVNC_connect_button {
     width: 100%;
-    padding: 5px 30px;
-
+    padding: 6px 30px;
     cursor: pointer;
-
-    border-color: var(--novnc-darkblue);
+    border-color: transparent;
     border-radius: 6px;
-
     background-color: var(--novnc-blue);
     color: white;
 }
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -737,9 +737,9 @@ html {
     }
 }
 #noVNC_connect_dlg div {
-    padding: 12px;
+    padding: 18px;
 
-    background-color: var(--novnc-blue);
+    background-color: var(--novnc-darkgrey);
     border-radius: 12px;
     text-align: center;
     font-size: 20px;
@@ -751,7 +751,7 @@ html {
     padding: 6px 30px;
     cursor: pointer;
     border-color: transparent;
-    border-radius: 6px;
+    border-radius: 12px;
     background-color: var(--novnc-blue);
     color: white;
 
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -31,6 +31,7 @@
 
 :root {
     font-family: sans-serif;
+    line-height: 1.6;
 }
 
 body {
@@ -477,9 +478,12 @@ html {
 .noVNC_panel .noVNC_heading {
     background-color: var(--novnc-blue);
     border-radius: 6px;
-    padding: 5px;
+    padding: 5px 8px;
     /* Compensate for padding in image */
-    padding-right: 8px;
+    padding-right: 11px;
+    display: flex;
+    align-items: center;
+    gap: 6px;
     color: white;
     font-size: 20px;
     white-space: nowrap;
@@ -736,6 +740,11 @@ html {
     border-radius: 6px;
     background-color: var(--novnc-blue);
     color: white;
+
+    display: flex;
+    justify-content: center;
+    place-items: center;
+    gap: 4px;
 }
 
 #noVNC_connect_button img {
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -53,7 +53,7 @@ html {
 }
 
 .noVNC_disabled {
-    color: rgb(128, 128, 128);
+    color: var(--novnc-grey);
 }
 
 /* ----------------------------------------
@@ -241,7 +241,7 @@ html {
 
     transition: 0.5s ease-in-out;
 
-    background-color: rgb(110, 132, 163);
+    background-color: var(--novnc-blue);
     border-radius: 0 10px 10px 0;
 
     user-select: none;
@@ -286,7 +286,7 @@ html {
     z-index: -1;
     cursor: pointer;
     border-radius: 5px;
-    background-color: rgb(83, 99, 122);
+    background-color: var(--novnc-darkblue);
     background-image: url("../images/handle_bg.svg");
     background-repeat: no-repeat;
     background-position: right;
@@ -371,7 +371,7 @@ html {
     opacity: 0;
     transition: 0.2s ease-in-out;
     background: transparent;
-    box-shadow: 0 0 10px black, inset 0 0 10px 10px rgba(110, 132, 163, 0.8);
+    box-shadow: 0 0 10px black, inset 0 0 10px 10px var(--novnc-darkblue);
     border-radius: 10px;
     transition-delay: 0s;
 }
@@ -478,7 +478,7 @@ html {
 
 .noVNC_panel hr {
     border: none;
-    border-top: 1px solid rgb(192, 192, 192);
+    border-top: 1px solid var(--novnc-lightgrey);
 }
 
 .noVNC_panel label {
@@ -499,7 +499,7 @@ html {
 }
 
 .noVNC_panel .noVNC_heading {
-    background-color: rgb(110, 132, 163);
+    background-color: var(--novnc-blue);
     border-radius: 5px;
     padding: 5px;
     /* Compensate for padding in image */
@@ -577,7 +577,7 @@ html {
 }
 
 #noVNC_modifiers {
-    background-color: rgb(92, 92, 92);
+    background-color: var(--novnc-darkgrey);
     border: none;
     padding: 10px;
 }
@@ -745,7 +745,7 @@ html {
 #noVNC_connect_dlg div {
     padding: 12px;
 
-    background-color: rgb(110, 132, 163);
+    background-color: var(--novnc-blue);
     border-radius: 12px;
     text-align: center;
     font-size: 20px;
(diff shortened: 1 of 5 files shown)


samhed:
Note that the failed check from `utils/validate` looks like a false positive. The property `scrollbar-gutter: stable` looks like it should be fully supported:

https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter

Unsure if we want to add an exception to `utils/validate` for that..

app/styles/base.css
```diff
@@ -30,30 +30,30 @@
  */
 
 :root {
-  font-family: sans-serif;
+    font-family: sans-serif;
```
CendioOssman, line 33: I guess this unfortunately breaks `git blame` for basically every line in these files? :/

CendioOssman, line 33: Is there a CSS lint we can turn on to avoid issues in the future?

app/styles/input.css
```diff
@@ -1,13 +1,17 @@
...
+ * Copyright (C) 2025 The noVNC authors
  * noVNC is licensed under the MPL 2.0 (see LICENSE.txt)
  * This file is licensed under the 2-Clause BSD license (see LICENSE.txt).
  */
 
 /*
  * Common for all inputs
  */
+:root {
+    --input-xpadding: 1em;
```
CendioOssman, line 11: I would vote against a variable for this, as it is so few places that need it.

app/styles/base.css
```diff
@@ -491,6 +491,13 @@ html {
     margin: 5px;
 }
 
+.noVNC_panel label > button,
+.noVNC_panel label > select,
+.noVNC_panel label > textarea,
+.noVNC_panel label > input:not([type=checkbox]):not([type=radio]) {
```
CendioOssman, line 474: Did you consider doing a cleanup to use `:is()`?

app/styles/constants.css
```diff
@@ -19,3 +19,9 @@
     --novnc-green: rgb(0, 128, 0);
     --novnc-yellow: rgb(255, 255, 0);
 }
+
+/* ---------- SIZES ----------- */
+
+:root {
+    --border-width: 1px;
```
CendioOssman, line 29: This also seems superfluous to have a variable for. It's not like we're going to fiddle around with this?

app/styles/constants.css
```diff
@@ -24,4 +24,6 @@
 
 :root {
     --border-width: 1px;
+    --small-border-radius: 6px; /* smaller elements such as buttons */
+    --large-border-radius: 12px; /* larger elements such as panels */
```
CendioOssman, line 31: Dito

CendioOssman:
Could we get some before/after pictures? :)

app/styles/input.css
```diff
@@ -332,6 +339,15 @@ input[type=range] {
     padding: 0;
     background: transparent;
 }
+input[type=range]:hover {
+    cursor: grab;
```
CendioOssman, line 417: This isn't mentioned in the commit message?

app/styles/constants.css
```diff
@@ -30,3 +30,9 @@
     --small-border-radius: 6px; /* smaller elements such as buttons */
     --large-border-radius: 12px; /* larger elements such as panels */
 }
+
+/* ------ MISC PROPERTIES ------ */
+
+:root {
+    --line-height: 1.6;
```
CendioOssman, line 37: Similar issue here. This variable is barely used?

app/styles/input.css
```diff
@@ -85,7 +86,7 @@ textarea {
     /* Make textareas show at minimum one line. This does not work when
        using box-sizing border-box, in which case, vertical padding and
        border width needs to be taken into account. */
-    min-height: 1em;
+    min-height: calc(1em * var(--line-height));
```
CendioOssman, line 89: Have you checked if there are any magical CSS units that might do what we want without `calc()`?

app/styles/base.css
manics: You can tell GitHub to ignore specified commits in git blame:
https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

I use prettier to format html/css/js/markdown but you could configure it for CSS only: https://prettier.io/

CendioOssman: Oooh! Nice! I was not aware of that feature. Might be a bunch of old commits we want to add to that as well.

app/styles/input.css
samhed: It is used in 6 places, in both input.css and base.css. I'll keep it for now.

app/styles/base.css
samhed: No, not right now. Would be nice though

app/styles/constants.css
samhed: Removed these two as well

app/styles/input.css
samhed: My bad, updated the commit message

app/styles/constants.css
samhed: Removed it

app/styles/input.css
samhed: Sweet, you are right! There is a unit that does what we want!! `1lh` :fire: :confetti_ball: :rocket:

Updated to use that one

app/styles/constants.css
```diff
@@ -0,0 +1,30 @@
...
+
+/* ---------- COLORS ----------- */
+
+:root {
+    --novnc-grey: rgb(128, 128, 128);
+    --novnc-lightgrey: rgb(192, 192, 192);
+    --novnc-darkgrey: rgb(92, 92, 92);
+
+    /* Transparent to make button colors adapt to the background */
+    --novnc-buttongrey: rgba(192, 192, 192, 0.5);
```
ThinLinc-Zeijlon, line 16: The choice of color here, I think, results in unchecked checkboxes being a bit too close in color to the gray areas that, e.g., can be expanded in the settings menu. Hence, possibly making them hard to see for people with visual impairments.

app/styles/base.css
```diff
@@ -729,13 +729,9 @@ html {
 }
 #noVNC_connect_button {
     width: 100%;
-    padding: 5px 30px;
-
+    padding: 6px 30px;
     cursor: pointer;
-
-    border-color: var(--novnc-darkblue);
```
ThinLinc-Zeijlon: Shouldn't it be `border-color: transparent;` if the intent is to remove visible borders?
For me, in chrome, there is a visible border after the `border-color` was removed.

app/styles/constants.css
samhed: I tried to account for that, isn't it good enough? Making the buttons even darker makes things look very contrasty on a white background.

Do you have an idea for how to do it better?

app/styles/input.css
```diff
@@ -409,6 +409,44 @@ input[type=range] {
...
+         M 4.5 3 \
+         L 4.5 0 \
+         L 120.5 0 \
+         L 120.5 3 \
+         A 1 1 0 0 1 120.5 12 \
+         L 120.5 15 \
+         L 4.5 15 \
+         L 4.5 12 \
+         A 1 1 0 0 1 4.5 3 \
+        ');
```
ThinLinc-Zeijlon, line 444: Is it possible for the box-section of the clip path to depend on the height of the thumb. (Note that I'm not sure if it is made to match the exact height/size of the tumb, or if you already have added a margin there.)

samhed: Not that I could find, sadly. MDN says:

> There are some limitations to using the path() function. The path has to be defined as a single string, so a custom path can't be created using variables (var() functions). Also, all the lengths in the path are implicitly defined in pixel (px) units; other units can't be used. The shape() function offers more flexibility than the path() function.

https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/path

The shape() function isn't available on chrome yet.

Another method I tried was to use url() instead of path(). But I couldn't get an svg data URL to work for clip-path.

app/styles/input.css
```diff
@@ -409,6 +409,44 @@ input[type=range] {
...
+           shadow of -webkit-slider-thumb since it doesn't match the smaller
+           border-radius of the progress track. The below clip-path has two
+           circular sides to make the ends of the track have correctly rounded
+           corners. The clip path shape looks something like this:
+
+                  +-------------------------------+
+              /---|                               |---\
+             |                                         |
+              \---|                               |---/
+                  +-------------------------------+
```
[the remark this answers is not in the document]
samhed: Hehe thank you

app/styles/base.css
samhed: Whoops, I'll fix that

app/styles/base.css
```diff
@@ -474,6 +480,8 @@ html {
 .noVNC_panel label > textarea,
 .noVNC_panel label > input:not([type=checkbox]):not([type=radio]) {
     margin-left: 6px;
+    /* Prevent inputs in panels from being too wide */
+    max-width: calc(100% - var(--input-xpadding) * 2);
```
ThinLinc-Zeijlon, line 484: It looks like this does not prevent  too wide inputs in the settings panel, where labels are not wrapping the input, but are instead using the `<label for="...">`-syntax.
too_wide_inputs

The screenshot was taken in Chrome's inspector mode with **iPhone 12 Pro** screen size setting. The same is **not** seen when checking via a real iPhone (13 Pro in this case).

samhed:
> Could we get some before/after pictures? :)

Connect button before:
connect before
Connect button after:
connect after

Auth dialog before:
dialog before
Auth dialog after:
dialog after

Settings before:
settings before
Settings after:
settings after

app/styles/base.css
samhed: Fixed now

app/styles/base.css
samhed: Fixed now

samhed:
> Note that the failed check from `utils/validate` looks like a false positive. The property `scrollbar-gutter: stable` looks like it should be fully supported:
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter
> 
> Unsure if we want to add an exception to `utils/validate` for that..

Added exceptions in b747069

app/styles/constants.css
ThinLinc-Zeijlon: I think it would work if the checkboxes had the same color as e.g. the select boxes, take a look at the one for setting *Scaling mode*. It's just a little bit darker.

I don't know it it's intended, but selects are a bit darker than input buttons which are of the same color as the checkboxes.

samhed: The color is the same, the large elements just look darker because they're big. To remedy the contrast issue slightly, the background of the expanded sections was made a bit lighter in 2c48769.

app/styles/base.css
```diff
@@ -759,7 +759,7 @@ html {
     cursor: pointer;
 
     border-color: var(--novnc-darkblue);
-    border-radius: 5px;
+    border-radius: 12px;
```
ThinLinc-Zeijlon: Did you intend to set the border radius here to `6px`? I Just found it as a little inconsistent with the other changes in the same commit, where you converted all `5px` to `6px`.

I think it looks fine either way.

samhed: Good catch! It is intended to be 12px, but I see it ended up in the wrong commit. A mistake from a recent rebase. Fixed now in 6db9dbc

samhed:
I noticed that the color was wrong for the range slider track, that was also fixed in 8800923. The track now matches the other input's better.

PR #1929, opened 2025-01-17
Use semantic tags instead of style tags

ThinLinc-Zeijlon:
It is good practice to provide styling that help indicate the importance of the presented text.

### DIFF
--- a/vnc.html
+++ b/vnc.html
@@ -330,7 +330,7 @@
             The server has provided the following identifying information:
         </div>
         <div id="noVNC_fingerprint_block">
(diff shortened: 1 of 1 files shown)


CendioOssman:
Looks good. Thanks!

Any CSS that needs to be updated to match?

CendioOssman (changes requested):
Actually, "strong" is probably not semantically correct here as it is not really part of a sentence that you want to emphasize.

CendioOssman (approved):

PR #1942, opened 2025-03-10
Let browser handle parsing of urls before relaying

ThinLinc-Zeijlon:
Apparently there is some magic happening under the surface when the `path`-string containing a search query is assigned to `url.pathname`.

This is what caused `?` to be translated to `%3F`.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1058,7 +1058,12 @@ const UI = {
             if (port) {
                 url.port = port;
             }
-            url.pathname = '/' + path;
+
+            // "./" is needed to force URL() to interpret the path-variable as
+            // a path and not as an URL. This is relevant if for example path
+            // starts with more than one "/", in which case it would be
+            // interpreted as a host name instead.
+            url = new URL("./" + path, url);
         } else {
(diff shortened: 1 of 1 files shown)


app/ui.js
```diff
@@ -1052,13 +1052,17 @@ const UI = {
...
+            let _url = "https://" + host;
 
-            url.protocol = UI.getSetting('encrypt') ? 'wss:' : 'ws:';
             if (port) {
-                url.port = port;
+                _url += ":" + port;
             }
-            url.pathname = '/' + path;
+
+            _url += path.startsWith("/") ? path : "/" + path;
```
CendioOssman, line 1061: Do we need the ternary? Can't we just always add the `/`? I assume it will be stripped out if duplicated, just like before.

ThinLinc-Zeijlon: Then again, urls like [https://github.com//novnc////noVNC////////pull//1942](https://github.com//novnc////noVNC////////pull//1942), are apparently valid, so I guess either works..

ThinLinc-Zeijlon: Removed the ternary in favor of keeping the change in behavior minimal.

PR #1951, opened 2025-04-03
Qemu audio support

maurerdietmar:
Add support for the Qemu audio extension.

PR #1457, opened 2020-08-27
QEMU Extended Audio Support

tinyzimmer:
I opened a fresh draft because I had way too many mistakes on my first branch :stuck_out_tongue: 

I have not yet tested these changes and obviously documentation will need to be updated. This is mainly intended to continue the discussion from #302.

Forgive me if I made some mistakes, since this is my first time really digging into the underlying protocol, but this is my first draft attempt at implementing handling for the [QEMU Audio Server Message](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#qemu-audio-server-message) and public methods wrapping the [QEMU Audio Client Message](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#qemu-audio-client-message).

I'm trying to figure out the best way I could go about testing these changes and fine tuning them. In my current project where I'm using noVNC I am simply proxying the data from an Xvnc server, so I would need to find a good one to use that supports the extended protocol.

PR #1456, opened 2020-08-27
[WIP] QEMU Extended Audio Support

tinyzimmer:
I have not yet tested these changes and obviously documentation will need to be updated. This is mainly intended to continue the discussion from #302. 

Forgive me if I made some mistakes, since this is my first time really digging into the underlying protocol, but this is my first draft attempt at implementing handling for the [QEMU Audio Server Message](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#qemu-audio-server-message) and public methods wrapping the [QEMU Audio Client Message](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#qemu-audio-client-message).

I'm trying to figure out the best way I could go about testing these changes and fine tuning them. In my current project where I'm using noVNC I am simply proxying the data from an Xvnc server, so I would need to find a good one to use that supports the extended protocol.

Issue #302, opened 2013-09-18
Audio support


samhed:
Would require implementing, for example, a PulseAudio server in javascript.

monkey-jsun:
Samhed, can you elaborate your thoughts on this?

I see two general approaches:

1.  streaming PCM or encoded audio over internet (perhaps extending VNC protocol, or simply open another websocket) and then use web audio API to playback on client machine.

2. extend the webserving capability on the remote machine to provide some audio URL source, e.g., <remote>:5901/system-audio.mp3.  And on client side we simply present a <audio> tag

What do you think?

samhed:
My thoughts were option 1, I actually have a prototype that works for ESound which uses Web Audio API, but that's quite an outdated protocol..

kanaka:
@samhed did you mean "Audio Data API". Web Audio API is the modern API. The old deprecated one that Mozilla prototype was the Audio Data API (https://wiki.mozilla.org/Audio_Data_API)

samhed:
Nope, i meant Web Audio API. I made it last year. I used ESound since it is a very simple protocol and I was mostly after a PoC.

monkey-jsun:
BTW, just curious about the choice of ESound.

I thought we might  be able to use a regular compression codec and
piggyback on VNC (using a custom type).

Jun

ikreymer:
Just wanted to check if anyone is currently working on this (I'm guessing not, but good to check)

@samhed Is your EsounD solution open source?

Otherwise, I was thinking perhaps https://github.com/pdeschen/pcm.js may be a start for pulseaudio

I would like to add sound to my project: oldweb.today

samhed:
@ikreymer I'm afraid it's not open source, but I don't think the EsounD solution is worth much. I'm currently not working on this however.

One issue I see with implementing audio support into the noVNC project is that such a solution would require server-side modifications to function, which I don't think fit in noVNC. First, pulseaudio on the server has to be configured to transmit sound from applications to a TCP port. Furthermore a WebSocket has to be opened to send the audio data from that port to the noVNC client. This requires some sort of server-side service.

But after all noVNC could simply be able to interpret and play audio data from a WebSocket which probably is sufficient.

samhed:
@ikreymer I had a discussion with my boss about the EsounD solution. We decided to make the client-side open. You can find the patch attached to the following bug:

https://www.cendio.com/bugzilla/show_bug.cgi?id=4822

macdonaldj:
[...]

PR #1952, opened 2025-04-06
Qemu audio support v2

maurerdietmar:
Add support for the Qemu audio extension.

changes in v2:

- ability to re-sync by resampling audio
- move audio code to separate file

### DIFF
--- a/core/audio.js
+++ b/core/audio.js
@@ -1,0 +1,168 @@
+// The RFB protocol (VNC) is designed for real-time user interactions
+// and allows transferring audio messages together with screen content.
+// It is not possible to use any kind of buffering, because that would
+// introduce large delays between user interaction and content display.
+//
+// This is not really a problem with screen content, because the human
+// brain is quite tolerate about slight speed changes in video content,
+// and we mostly transfer non-video data anyways.
+//
+// With audio, the situation is quite different, as it must be played
+// at a constant speed. Any delay leads to audio distortion, which is
+// unpleasant for humans.
+//
+// Without buffering, it is always possible for audio frames to arrive
+// too late or too early due to changing network speeds.
+//
+// We use the following algorithm:
+//
+// - small Jitter buffer to tolerate small speed changes (20ms)
+// - simply discard late audio frame
+// - Queue early frames with slight speedup (pitch scale) to re-sync audio
+// - if we get to many early frames, skip frames for fast re-sync
+
+import * as Log from './util/logging.js';
+
+// We use/expect Qemu U16 audio data
+
+export default class Audio {
+    constructor(sample_rate, channels) {
+        this._next_start = 0;
+        this._context = null;
+        this._jitter = 0.02;
+        this._resample_trigger = 5*this._jitter;
+        this._stable_time = 1.0;
+
+        // ===== PROPERTIES =====
+        this._sample_rate = sample_rate;
+        this._channels = channels;
+    }
+
+    // ===== PROPERTIES =====
+    get sample_rate() { return this._sample_rate; }
+    get channels() { return this._channels; }
+
+    // ===== PUBLIC METHODS =====
+
+    // Stop audio playback
+    //
+    // Further audio frames are simply dropped.
+    stop() {
+        this._context = null;
+        this._next_start = 0;
+    }
+
+    start() {
+        this._context = new AudioContext({
+            latencyHint: "interactive",
+            sampleRate: this._sample_rate,
+        });
+        this._next_start = 0;
+    }
+
+    play(payload) {
+        if (this._context === null) {
+            return true;
+        }
+
+        let ctime = this._context.currentTime;
+
+        let time_offset = this._next_start - ctime;
+
+        let sample_bytes = 2*this._channels;
+
+        if ((time_offset < this._jitter) && (this._resample_trigger !== 5*this._jitter)) {
+            Log.Warn("Stop resampling because audio is in sync (delay = " + time_offset + " sec)");
+            this._resample_trigger = 5*this._jitter;
+        }
+
+        let buffer = null;
+        if (time_offset > this._resample_trigger && (payload.length > (100*sample_bytes))) {
+            if (this._resample_trigger !== this._jitter) {
+                Log.Warn("Start resampling to re-sync audio (delay = " + time_offset + " sec)");
+                this._resample_trigger = this._jitter;
+            }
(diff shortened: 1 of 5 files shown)


core/audio.js
```diff
@@ -0,0 +1,168 @@
...
+// - small Jitter buffer to tolerate small speed changes (20ms)
+// - simply discard late audio frame
+// - Queue early frames with slight speedup (pitch scale) to re-sync audio
+// - if we get to many early frames, skip frames for fast re-sync
+
+import * as Log from './util/logging.js';
+
+// We use/expect Qemu U16 audio data
+
+export default class Audio {
```
ThinLinc-Zeijlon, line 28: `QEMUAudio` (based on your comment above)?

core/audio.js
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        let ctime = this._context.currentTime;
+
+        let time_offset = this._next_start - ctime;
+
+        let sample_bytes = 2*this._channels;
+
+        if ((time_offset < this._jitter) && (this._resample_trigger !== 5*this._jitter)) {
+            Log.Warn("Stop resampling because audio is in sync (delay = " + time_offset + " sec)");
```
ThinLinc-Zeijlon, line 75: This should be debug logging, a normal user shouldn't see this.

core/audio.js
```diff
@@ -0,0 +1,168 @@
...
+            this._resample_trigger = 5*this._jitter;
+        }
+
+        let buffer = null;
+        if (time_offset > this._resample_trigger && (payload.length > (100*sample_bytes))) {
+            if (this._resample_trigger !== this._jitter) {
+                Log.Warn("Start resampling to re-sync audio (delay = " + time_offset + " sec)");
```
ThinLinc-Zeijlon, line 82: Same here as the previous comment. When I tested [video](http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4) playback in full screen on a Fedora 41 vm with 8 cores (no gpu), running on my local machine, there was quite a bit of logging.

ThinLinc-Zeijlon (commented):
Overall a nice implementation!

I found that I could make the audio stutter, e.g. when running a video in full screen. But I'm not sure if this is due to the CPU being overloaded or not, the Image would also stutter when this happened.

I would also like to see unittests for your code.

maurerdietmar:
I am not really sure if we should rename class Audio to QemuAudio, because we can easily add support for other formats in the future. If you really want to rename the class, should we also rename the file to "qemu-audio.js"?

core/rfb.js
```diff
@@ -154,6 +154,13 @@ export default class RFB extends EventTargetMixin {
 
         this._qemuExtKeyEventSupported = false;
 
+        this._qemuAudioSupported = false;
+        this._page_had_user_interaction = false;
+        this._audio_next_start = 0;
+        this._audio_sample_rate = 44100;
+        this._audio_channels = 2;
```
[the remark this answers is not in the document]
ThinLinc-Zeijlon: I've given this a bit more thought, and I'll have to go back to what I said here. We don't want to add new things to the RFB class/API if we don't have to.

Ideally, we would like to ask the system/browser running the client for it's preferred audio settings by some means. But the settings you had there before are good enough as a starting point.

core/audio.js
ThinLinc-Zeijlon: > I am not really sure if we should rename class Audio to QemuAudio, because we can easily add support for other formats in the future. If you really want to rename the class, should we also rename the file to "qemu-audio.js"?

No, I agree that we want a generalized Audio class.

ThinLinc-Zeijlon: The comment is probably a bit improper in this case? Are there things that could be documented in more general terms that aren't specific to just the Qemu use case?

ThinLinc-Zeijlon:
Please resolve the comments you think are fixed, and edit your commits to what you would want the end result to look like and do a push with `--force`.

core/audio.js
maurerdietmar: done

core/audio.js
maurerdietmar: done

liyimeng:
Great!

CendioZeijlon:
Apologies for the lack of activity in this PR on my part, I'll try to pick up the slack :) 

I left this in a review comment, maybe it was missed:
>I've given this a bit more thought, and I'll have to go back to what I said here. We don't want to add new things to the RFB class/API if we don't have to.
> 
> Ideally, we would like to ask the system/browser running the client for it's preferred audio settings by some means. But the settings you had there before are good enough as a starting point.

Let's also assume that users want audio enabled, such that we don't have to add settings for toggling audio on and off etc. This should mean that you don't have to change the RFB-class API by adding e.g. an enable_audio()-function.

maurerdietmar:
> Let's also assume that users want audio enabled, such that we don't have to add settings for toggling audio on and off etc. This should mean that you don't have to change the RFB-class API by adding e.g. an enable_audio()-function.

Considering audio is not compressed (currently), this can lead to high bandwidth usage. So IMHO it a good choice to make this configurable...

CendioZeijlon:
> Considering audio is not compressed (currently), this can lead to high bandwidth usage. So IMHO it a good choice to make this configurable...

@CendioOssman, do you have any input here? I know we prefer to avoid *visible* changes to the RFB object, but @maurerdietmar raises a good point.

CendioOssman:
VNC by its very nature already assumes a decent availability of bandwidth, but I suppose in some cases audio could be the tipping point.

Make sure it is a property, though, like the other settings.

liyimeng:
> > Let's also assume that users want audio enabled, such that we don't have to add settings for toggling audio on and off etc. This should mean that you don't have to change the RFB-class API by adding e.g. an enable_audio()-function.
> 
> Considering audio is not compressed (currently), this can lead to high bandwidth usage. So IMHO it a good choice to make this configurable...

Is h.264 handle audio compression together with video? when not take care of audio and video via h.264?

kroese:
Can this be merged? Or what is holding it back?

Neustradamus:
This PR: https://github.com/novnc/noVNC/pull/1952 replaces:
- https://github.com/novnc/noVNC/pull/1456
- https://github.com/novnc/noVNC/pull/1457
- https://github.com/novnc/noVNC/pull/1951

Linked to:
- https://github.com/novnc/noVNC/issues/302

Issue #1943, opened 2025-03-13
noVNC cannot be bundled due to top-level await in browser.js


mudge:
**Describe the bug**
3677afe30514fb3c678e35bb9b6bde924820be8d introduced a [top-level await in `lib/util/browser.js`](https://github.com/novnc/noVNC/blob/a8dfd6a3ea3c74244f5ebdaa5a7f1023007a7820/core/util/browser.js#L152) which prevents the library from being bundled with [esbuild](https://esbuild.github.io) when previously it was fine.

**To reproduce**
1. In a new directory, run `npm install -S @novnc/novnc esbuild`
2. Create an `index.js` with `import RFB from "@novnc/novnc";`
3. Attempt to bundle the file with `npx esbuild index.js --bundle`
4. See errors like the following:
    ```
    ✘ [ERROR] This require call is not allowed because the imported file "node_modules/@novnc/novnc/lib/util/browser.js" contains a top-level await

    node_modules/@novnc/novnc/lib/util/cursor.js:7:23:
      7 │ var _browser = require("./browser.js");
        ╵                        ~~~~~~~~~~~~~~

  The top-level await in "node_modules/@novnc/novnc/lib/util/browser.js" is here:

    node_modules/@novnc/novnc/lib/util/browser.js:179:68:
      179 │ exports.supportsWebCodecsH264Decode = supportsWebCodecsH264Decode = await _checkWebCodecsH264DecodeSupport();
          ╵   
    ```

**Expected behavior**
The file bundles successfully.

**Server (please complete the following information):**
 - noVNC version: 1.6.0
 - Node version: v20.9.0
 - OS: macOS 15.3.1

**Additional context**
This happens even when using esbuild's default [target](https://esbuild.github.io/api/#target) of `esnext`. Is it possible to compile a version of `browser.js` that does not use a top-level await?

SteveW94:
Sadly had to revert back to 1.5.0 also due to this issue :/

CendioOssman:
I don't think modules were designed with bundling in mind, so I suspect this will always be a bit shaky.

Bundling is also not something we've set out to explicitly support, so off the bat this would be an esbuild issue rather than a noVNC one. They seem to already be tracking it as evanw/esbuild#253.

As for not using top level awaits; that is not something we can commit to. Most web APIs are asynchronous. Using this construct allows us to avoid a lot of complexity and bugs.

[...]

PR #1958, opened 2025-05-06
Fix broken NPM bundle by publishing in ES-Module format

elikoga:
Closes https://github.com/novnc/noVNC/issues/1943
Repeat of #1944 with different head branch

Testable by placing `"@novnc/novnc": "npm:@elikoga/novnc@1.6.1-alpha.1",` in `package.json`

Uses `exports` in `package.json` to make sure that old imports still work

This one incorporates the review requests in the previous PR by simply dropping the conversion step using babel.

CI Passes in my repo


CendioOssman:
Please don't open a new PR as that means it is more difficult to see all the context. You can just adjust the existing one with your changes.

package.json
```diff
@@ -1,25 +1,32 @@
...
+    "./lib/*": "./core/*.js",
+    "./lib/*.js": "./core/*.js",
+    "./core/*": "./core/*.js",
+    "./core/*.js": "./core/*.js",
+    "./lib/vendor/*": "./vendor/*.js",
+    "./lib/vendor/*.js": "./vendor/*.js",
+    "./vendor/*": "./vendor/*.js",
+    "./vendor/*.js": "./vendor/*.js",
+    "./package.json": "./package.json"
+  },
```
CendioOssman, line 26: RFB is our only interface. Do we need anything more than the first line?

package.json
```diff
@@ -1,25 +1,32 @@
 {
   "name": "@novnc/novnc",
-  "version": "1.6.0",
+  "version": "1.6.1-alpha.1",
```
CendioOssman, line 3: Oops?

package.json
```diff
@@ -1,25 +1,32 @@
...
   "description": "An HTML5 VNC client",
-  "browser": "lib/rfb",
```
CendioOssman, line 5: Shouldn't this be kept in some form to indicate that this is code meant to be run in a browser?

package.json
```diff
@@ -1,25 +1,32 @@
...
-  "directories": {
-    "lib": "lib",
-    "doc": "docs",
-    "test": "tests"
-  },
```
CendioOssman, line 10: Are these not used anymore?

utils/genkeysymdef.cjs
CendioOssman, line 1: Couldn't this be converted as well?

CendioOssman (changes requested):

micheldpcarlos:
Hey! Just wondering if there’s any ETA on when this might get merged. :) 
Happy to assist if anything’s needed to move it along. Thanks!

LeeTeng2001:
damn nobody use this huh

CendioOssman:
@elikoga, is this still on your todo list?

package.json
elikoga: There are uses I could find online using the github search:

with lib: https://github.com/SDU-eScience/UCloud/blob/e78be592106cd915bb9004fc095daa5140e23ed1/frontend-web/webclient/app/Applications/Jobs/Vnc.tsx#L13

https://github.com/kubevirt-ui/kubevirt-plugin/blob/447047ed328719698db1f9e1de5f2db81dc40b89/src/utils/components/Consoles/components/vnc-console/VncConsole.tsx#L6

with core: https://github.com/harvester/dashboard/blob/6da3d8db703521eef18cb30ad60c4148bba4edff/pkg/harvester/components/novnc/NovncConsoleWrapper.vue#L5

And since I do not want these inputs to break I have written it as it is there.

I could probably remove the part about vendor, since it does not seem to be used.

package.json
elikoga: Yes, Oops

package.json
elikoga: I don't know which software reads this in the presence of "exports" I could not find any examples

package.json
elikoga: See npm js docs: [directories](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#directories)

> The CommonJS [Packages](http://wiki.commonjs.org/wiki/Packages/1.0) spec details a few ways that you can indicate the structure of your package using a directories object. If you look at [npm's package.json](https://registry.npmjs.org/npm/latest), you'll see that it has directories for doc, lib, and man.
> 
> In the future, this information may be used in other creative ways.

Since this is no longer exported using ComonJS Modules, I am not sure who would use this.

package.json
elikoga: Fixed

utils/genkeysymdef.cjs
elikoga: Fixed

package.json
CendioOssman: Reading the npm documentation, this isn't entirely clear. It states that "exports" replaces "main". And since "browser" is an alternative to "main", I suppose we can assume that it is also replaced.

So let's go with your suggestion.

It would be nice if there was something in "exports" that still indicates that this is a browser package, but I can't find anything along those lines.

package.json
CendioOssman: Those are all unsupported usages of noVNC. I'm not keen on giving any illusion of that being official and stable.

Ideally, we just have:

```js
  "exports": "./core/rfb.js",
```

Which is the only supported interface at the moment.

The question is what that transition means for people. Given that we are switching from CommonJS to modules, I assume there is some transition work needed anyway? So might be a good idea to do this at the same time?

package.json
CendioOssman: Sounds fair. But let's have a separate commit that explicitly states that these are being removed because they are old and unused.

Can probably put the removal of "browser" in the same commit.

seandilda:
> damn nobody use this huh

This is still an issue for me.  I'm currently having to lock at version 1.5.0 so that I can properly bundle the code.

package.json
elikoga: `util/logging.js` is clearly also exposed api, otherwise, the provided example vnc.html would not work:

https://github.com/novnc/noVNC/blob/4cb5aa45ae559f8fa85fe2b424abbc6ef6d4c6f9/vnc.html#L53

elikoga: As are the files under `input/key*`:
https://github.com/novnc/noVNC/blob/4cb5aa45ae559f8fa85fe2b424abbc6ef6d4c6f9/app/ui.js#L15-L17

jsuchenia:
> > damn nobody use this huh
> 
> This is still an issue for me. I'm currently having to lock at version 1.5.0 so that I can properly bundle the code.

I think that @elikoga is doing his best to implement a change, you can test his changes from his repo (https://github.com/elikoga/noVNC/tree/switch-to-es-module-publish-and-use-exports) and let here know if that's suitable for you

From a number of projects referencing this broken version (1.6.0 via code builders) - you're not the only one waiting for a fix

@CendioOssman Do you need any help in a validation of a changes here?

package.json
CendioOssman: Exposed is not the same as supported or stable. I.e. you use them at your own risk. We will not consider it a bug if we change them in a way that breaks callers.

For our internal use, that is fine as we can change it at the same time. But external users cannot really do the same safely.

CendioOssman:
> @CendioOssman Do you need any help in a validation of a changes here?

Please do. We don't use the npm version ourselves, so feedback from actual users is very welcome. Including on the proposed adjustments in my reviews.

package.json
CendioOssman: I'm aware this will cause some pain for some users. But that might be the best thing in the long run, as that will force a discussion of what is missing from the official API and needs to be added.

jsuchenia:
> > @CendioOssman Do you need any help in a validation of a changes here?
> 
> Please do. We don't use the npm version ourselves, so feedback from actual users is very welcome. Including on the proposed adjustments in my reviews.

Ok - so I've created a simple project that uses ViteJS and ReactJS ([samples in this gist](https://gist.github.com/jsuchenia/bcaed8505881390610b009c2715c6e80)) - and confirmed an issue with 1.6.0 version
Then - I've switched to a branch from this PR - and it's working as expected - no errors and code works fine

So - for web-based apps this change will give a possibility to bundle a code using modern package manages. 

What other tests shall I perform?

GeTOUO:
这个事项还在推进中吗？是否有一个可供期望的发布时间？目前我也是因为这个问题导致不得不停留在v1.5.0

elikoga:
> 这个事项还在推进中吗？是否有一个可供期望的发布时间？目前我也是因为这个问题导致不得不停留在v1.5.0

I am still available to look at it if needed. I do not know when it will be released. Upstream is not directly concerned with keeping npm up to date due to their releasing mechanisms not going through there. 1.5.0 is still completely fine, unless you need one of the newer features.

GeTOUO:
> > 这个事项还在推进中吗？是否有一个可供期望的发布时间？目前我也是因为这个问题导致不得不停留在v1.5.0
> 
> 如果需要，我仍然可以查看它。我不知道什么时候会发布。上游并不直接关注保持 npm 的最新状态，因为它们的发布机制没有通过那里。1.5.0 仍然完全没问题，除非您需要更新的功能之一。

好的，我可以继续使用1.5.0,但还是非常期待这个合并能帮助我升级到1.6.0以上。感谢。

jsuchenia:
@CendioOssman What shall we do to merge that mr? It's been few months to merge so simple (and organisational) change?

package.json
micheldpcarlos: Hi 👋
Just checking in to see if there are any updates on this thread

CendioOssman:
I made the last discussed adjustments and merged this. It will momentarily be available from NPM under the "dev" tag. Please have a look and see how it works.

torchsec-roni:
Is there another semver release planned soon?

CendioOssman:
That's the idea. Hopefully a beta soon.

ianb-mp:
I've tested [1.6.0-g44b7489](https://www.npmjs.com/package/@novnc/novnc/v/1.6.0-g44b7489) - lgtm 👍

PR #1959, opened 2025-05-07
Disable setting showDotCursor in RFB constructor

ThinLinc-Zeijlon:
This has been deprecated for around six years now. Let's remove the deprecation warning and disable setting `showDotCursor` via the options-parameter.


CendioOssman (approved):
Looks good!

PR #1966, opened 2025-06-02
Enhance readQueryVariable to support boolean, numeric and string values

mauigna06:


vnc_lite.html
```diff
@@ -117,7 +117,22 @@
...
             if (match) {
                 // We have to decode the URL since want the cleartext value
-                return decodeURIComponent(match[1]);
+                const value = decodeURIComponent(match[1]);
+                if (value.toLowerCase() === 'false' || value === '0') {
+                    return false;
+                }
+                if (value.toLowerCase() === 'true' || value === '1') {
+                    return true;
+                }
```
CendioOssman, line 126: This is not going to work since you don't know if the caller wants a Number or Boolean.

I'd suggest adding a typed wrapper so the caller can be clear what it wants. I.e. `readBoolQueryVariable()`.

CendioOssman (changes requested):

PR #1973, opened 2025-08-01
Feature: Enable self-signed certificates

blitztide:
This will use OpenSSL and generate a new certificate, then start the server in SSL-Only mode

### DIFF
--- a/utils/novnc_proxy
+++ b/utils/novnc_proxy
@@ -17,6 +17,8 @@ usage() {
     echo "                          Default: 6080 (on all interfaces)"
     echo "    --vnc VNC_HOST:PORT   VNC server host:port proxy target"
     echo "                          Default: localhost:5900"
+    echo "    --self-sign hostname  Generate self-signed certificates for hostname"
+    echo "                          Requires OpenSSL to be installed"
     echo "    --cert CERT           Path to combined cert/key file, or just"
     echo "                          the cert file if used with --key"
     echo "                          Default: self.pem"
@@ -51,6 +53,7 @@ HOST=""
 PORT="6080"
 LISTEN="$PORT"
 VNC_DEST="localhost:5900"
+SELF_SIGN=""
 CERT=""
 KEY=""
 WEB=""
@@ -90,6 +93,7 @@ while [ "$*" ]; do
     case $param in
     --listen)  LISTEN="${OPTARG}"; shift            ;;
     --vnc)     VNC_DEST="${OPTARG}"; shift        ;;
+    --self-sign) SELF_SIGN="${OPTARG}"; shift     ;;
     --cert)    CERT="${OPTARG}"; shift            ;;
     --key)     KEY="${OPTARG}"; shift             ;;
     --web)     WEB="${OPTARG}"; shift            ;;
@@ -145,6 +149,18 @@ elif [ -e "${HERE}/../share/novnc/vnc.html" ]; then
     WEB=${HERE}/../share/novnc/
 else
     die "Could not find vnc.html"
+fi
+
+# Create self-signed certificates
+if [ -n "${SELF_SIGN}" ]; then
+	if [ ! -f $(pwd)/self.pem ]; then
+		echo "Generating Certificate for: ${SELF_SIGN}"
+		openssl req -x509 -newkey rsa:4096 -keyout key.pem -out self.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=NoVNC/L=NoVNC/O=NoVNC/OU=NoVNC/CN=${SELF_SIGN}"
+	fi
+	CERT=$(pwd)/self.pem
+	KEY=$(pwd)/key.pem
+	echo "Forcing SSL"
+	SSLONLY="--ssl-only"
 fi
 
 # Find self.pem


CendioZeijlon:
Nice idea, I'll take a look!

utils/novnc_proxy
```diff
@@ -17,6 +17,8 @@ usage() {
     echo "                          Default: 6080 (on all interfaces)"
     echo "    --vnc VNC_HOST:PORT   VNC server host:port proxy target"
     echo "                          Default: localhost:5900"
+    echo "    --self-sign hostname  Generate self-signed certificates for hostname"
```
CendioZeijlon, line 20: Wouldn't it be better to not force users to provide the `hostname` argument here? If they care enough, they will probably generate a more detailed certificate.

utils/novnc_proxy
```diff
@@ -147,6 +151,18 @@ else
...
+# Create self-signed certificates
+if [ -n "${SELF_SIGN}" ]; then
+	if [ ! -f $(pwd)/self.pem ]; then
+		echo "Generating Certificate for: ${SELF_SIGN}"
+		openssl req -x509 -newkey rsa:4096 -keyout key.pem -out self.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=NoVNC/L=NoVNC/O=NoVNC/OU=NoVNC/CN=${SELF_SIGN}"
+	fi
+	CERT=$(pwd)/self.pem
+	KEY=$(pwd)/key.pem
+	echo "Forcing SSL"
+	SSLONLY="--ssl-only"
```
CendioZeijlon: Wouldn't it be better to let them choose if they want to force SSL or not?

utils/novnc_proxy
```diff
@@ -147,6 +151,18 @@ else
...
 fi
 
+# Create self-signed certificates
+if [ -n "${SELF_SIGN}" ]; then
+	if [ ! -f $(pwd)/self.pem ]; then
+		echo "Generating Certificate for: ${SELF_SIGN}"
+		openssl req -x509 -newkey rsa:4096 -keyout key.pem -out self.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=NoVNC/L=NoVNC/O=NoVNC/OU=NoVNC/CN=${SELF_SIGN}"
+	fi
+	CERT=$(pwd)/self.pem
+	KEY=$(pwd)/key.pem
```
CendioZeijlon, line 182: If I were to run
```sh
novnc_proxy --cert mycert.pem --key mykey.pem --ssl-sign myhostname ...
```
it wouldn't be clear to me that the cert and key I provided will be ignored. There should probably be a check that makes sure you aren't using these options at the same time.

utils/novnc_proxy
```diff
@@ -17,6 +17,8 @@ usage() {
     echo "                          Default: 6080 (on all interfaces)"
     echo "    --vnc VNC_HOST:PORT   VNC server host:port proxy target"
     echo "                          Default: localhost:5900"
+    echo "    --self-sign hostname  Generate self-signed certificates for hostname"
+    echo "                          Requires OpenSSL to be installed"
```
CendioZeijlon, line 21: What happens if it's not installed?

utils/novnc_proxy
```diff
@@ -147,6 +151,18 @@ else
     die "Could not find vnc.html"
 fi
 
+# Create self-signed certificates
+if [ -n "${SELF_SIGN}" ]; then
+	if [ ! -f $(pwd)/self.pem ]; then
```
CendioZeijlon: If I run the script with a different hostname a second time, the CN in the file won't change.

CendioZeijlon: This could also be an argument for not allowing users to set CN in the generated cert.

utils/novnc_proxy
blitztide: What would you suggest for the default value?

blitztide: I was using it to point to something behind a reverse proxy, for testing, so hostname is pretty critical for this.
Once fully set up, then a proper SSL would be generated.

utils/novnc_proxy
blitztide: It would be pretty silly to request a self-signed certificate and not want SSL enabled?

utils/novnc_proxy
blitztide: Fixed in latest commit

utils/novnc_proxy
blitztide: Checked and fixed in latest commit

utils/novnc_proxy
blitztide: Fixed in latest commit

utils/novnc_proxy
blitztide: Removed in latest commit

blitztide:
Tested changes, ensuring argument compatibility

utils/novnc_proxy
CendioZeijlon: I would suggest something equally generic to the other properties like `CN=noVNC-host` or perhaps `CN=${HOSTNAME}` if that environment variable is set.

> I was using it to point to something behind a reverse proxy, for testing, so hostname is pretty critical for this.
Once fully set up, then a proper SSL would be generated.

Why is CN in the certificate important in this case?

utils/novnc_proxy
```diff
@@ -153,14 +161,25 @@ fi
 
 # Create self-signed certificates
 if [ -n "${SELF_SIGN}" ]; then
-	if [ ! -f $(pwd)/self.pem ]; then
-		echo "Generating Certificate for: ${SELF_SIGN}"
-		openssl req -x509 -newkey rsa:4096 -keyout key.pem -out self.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=NoVNC/L=NoVNC/O=NoVNC/OU=NoVNC/CN=${SELF_SIGN}"
-	fi
+    # Check if OpenSSL is installed
+    which openssl > /dev/null
```
CendioZeijlon, line 165: You can do `which openssl &> /dev/null` if you also want to silence the output from stderr.

utils/novnc_proxy
```diff
@@ -153,14 +161,25 @@ fi
...
+	if [ -f $(pwd)/self.pem ]; then
+        read -p "$(pwd)/self.pem aleady exists, overwrite? (Y/N) " overwrite
+        if [ "$overwrite" != "Y" ]; then
+            echo "Not overwriting $(pwd)/self.pem"
+            exit 1
+        fi
+    fi
+	echo "Generating Certificate for: ${SELF_SIGN}"
+	openssl req -x509 -newkey rsa:4096 -keyout key.pem -out self.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=NoVNC/L=NoVNC/O=NoVNC/OU=NoVNC/CN=${SELF_SIGN}"
+
```
CendioZeijlon, line 180: Wouldn't it be more convenient if `--self-sign` only created a certificate if one was not found and echoed some info otherwise, `Skip generating self-signed certificate. $(pwd)/self.pem and $(pwd)/key.pem already exists` ? This way you could just manually delete the old files if you want to regenerate a cert.

Right now, if I e.g. run `novnc_proxy --self-sign` as a part of a larger script, I would have to manage the input to `read` on each subsequent run.

I know this maybe reverts parts of what you added here; sorry about that.

utils/novnc_proxy
```diff
@@ -132,6 +132,14 @@ if [ -z "${HOST}" ]; then
     fi
 fi
 
+# Check if (cert | key) & self-sign are set, as they are incompatible
+if [ -n "$CERT" ] || [ -n "$KEY" ] && [ -n "$SELF_SIGN" ]; then
+    echo "Arguments --cert and --key and incompatible with --self-sign"
+    echo ""
+    usage
+    exit 1
+fi
```
CendioZeijlon, line 141: Nice!

utils/novnc_proxy
CendioZeijlon, line 165: We should only require OpenSSL when we know it will be needed.

CendioZeijlon (changes requested):

utils/novnc_proxy
blitztide: --self-signed is more likely to be used in a test scenario.

If you are running scripts to spawn something, then certificate management will also be used in the higher level script.

utils/novnc_proxy
blitztide: I feel like it would be beneficial for someone to get a specific error if it arises.

'Which' normally returns errors in the retcode.

blitztide: The only pathway after this check where openssl is not required is where they don't want to overwrite their current certificate.

I think there is the intent to use OpenSSL at this point in the code.

utils/novnc_proxy
CendioZeijlon: I hear you, I still think forcing users to make a choice every time they run with `--self-sign` adds unnecessary inconvenience. It's more likely that they will just want to continue where they left off on subsequent runs.

utils/novnc_proxy
```diff
@@ -153,14 +161,25 @@ fi
...
+    # Check that the file doesn't already exist
+	if [ -f $(pwd)/self.pem ]; then
+        read -p "$(pwd)/self.pem aleady exists, overwrite? (Y/N) " overwrite
+        if [ "$overwrite" != "Y" ]; then
+            echo "Not overwriting $(pwd)/self.pem"
+            exit 1
+        fi
+    fi
+	echo "Generating Certificate for: ${SELF_SIGN}"
+	openssl req -x509 -newkey rsa:4096 -keyout key.pem -out self.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=NoVNC/L=NoVNC/O=NoVNC/OU=NoVNC/CN=${SELF_SIGN}"
```
CendioZeijlon, line 179: `key.pem` should be called `self.key` since `key.pem` might as well be the key for a valid certificate.

utils/novnc_proxy
```diff
@@ -153,14 +161,25 @@ fi
...
-		openssl req -x509 -newkey rsa:4096 -keyout key.pem -out self.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=NoVNC/L=NoVNC/O=NoVNC/OU=NoVNC/CN=${SELF_SIGN}"
-	fi
+    # Check if OpenSSL is installed
+    which openssl > /dev/null
+    if [ $? != 0 ]; then
+        echo "Unable to find OpenSSL, please ensure you have OpenSSL installed and available in \$PATH"
+        exit 1
+    fi
+    # Check that the file doesn't already exist
+	if [ -f $(pwd)/self.pem ]; then
```
CendioZeijlon, line 171: You should probably check for the existence of all files that are generated by your OpenSSL command.

utils/novnc_proxy
CendioZeijlon: Fair enough!

utils/novnc_proxy
CendioZeijlon: It's also allowed to set `-keyout self.pem` so that both cert and pubkey end up in the same file. Would probably simplify existence checks a bit.

PR #1974, opened 2025-08-02
Appending the copied Item from terminal, instead of Overwritting in the text area.

Ankit8969:
Now, when copying text from the noVNC terminal, the copied content is appended rather than overwriting the existing clipboard data. This enhancement is useful when copying multiple text snippets sequentially.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -45,7 +45,7 @@ const UI = {
     inhibitReconnect: true,
     reconnectCallback: null,
     reconnectPassword: null,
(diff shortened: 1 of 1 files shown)


CendioZeijlon:
We want the text area to always represent the clipboard state in the connected session. Adding your proposed changes would break this behavior.

There is also a security concern here -- potentially sensitive data would be kept in the text area, possibly without the user realizing, until cleared manually from outside the connected session.

PR #1983, opened 2025-08-26
Fix issue where tests fail to find chai library.

tsukasa-au:
A breaking change to how the chai library is bundled was made between 5.x.x and 6.x.x (it was originally made in 5.3.1, but that change was reverted in 5.3.2, then made again after bumping the major version number). This results in all locally executed karma tests failing due to a missing file.

This is a temporary work around until someone correctly fixes our usage or this libary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@babel/preset-env": "latest",
     "babel-plugin-import-redirect": "latest",
     "browserify": "latest",
-    "chai": "latest",
+    "chai": "<6.0.0",
     "commander": "latest",
     "eslint": "latest",
     "fs-extra": "latest",


CendioOssman:
We'd prefer to fix it properly. Do you have a reference to what has changed?

tsukasa-au:
I _think_ the original change was done in chaijs/chai#1694 (remove bundled chai), it was rolled back in chaijs/chai#1699, then rolled forward (with a version bump) in chaijs/chai#1700. 

My knowledge of modern javascript is a little rusty... But chaijs/chai#1698 seems to suggest we should be using `require` and a bundler.

CendioZeijlon:
This was fixed in commit 8ebd9ddef9b5b8ab5636a40854fb59fcb327fb39

PR #1984, opened 2025-08-26
Add wakelock support

tsukasa-au:
Add a new configuration option `use_wakelock` to allow noVNC to stop the local display from going to sleep. This is especially useful with view-only sessions.

We only hold the view lock while connected to a server. We will also attempt to reacquire the wakelock if we lost it due to a visibility change (the tab becoming inactive, or during the transition into/from fullscreen).

All existing unittests have been run (with #1983 manually patched into my working directory), and the change has been manually tested in Firefox 142.

### DIFF
--- a/docs/EMBEDDING.md
+++ b/docs/EMBEDDING.md
@@ -95,6 +95,10 @@ Currently, the following options are available:
 * `logging` - The console log level. Can be one of `error`, `warn`, `info` or
   `debug`.
 
+* `use_wakelock` - Should we prevent the (local) display from going into sleep
+  mode while a connection is active? Useful for view-only sessions where there
+  unlikely to be any keyboard/mouse activity to keep the device active.
+
 ## HTTP serving considerations
 ### Browser cache issue
 
--- a/app/ui.js
+++ b/app/ui.js
@@ -28,6 +28,7 @@ const UI = {
 
     connected: false,
     desktopName: "",
+    wakelock: null,
 
     statusTimeout: null,
     hideKeyboardTimeout: null,
@@ -189,6 +190,7 @@ const UI = {
         UI.initSetting('repeaterID', '');
         UI.initSetting('reconnect', false);
         UI.initSetting('reconnect_delay', 5000);
+        UI.initSetting('use_wakelock', false);
     },
     // Adds a link to the label elements on the corresponding input elements
     setupSettingLabels() {
@@ -1154,6 +1156,10 @@ const UI = {
         UI.showStatus(msg);
         UI.updateVisualState('connected');
 
+        if (UI.getSetting('use_wakelock')) {
+            UI.wakelockAcquire();
+        }
+
         // Do this last because it can only be used on rendered elements
         UI.rfb.focus();
     },
@@ -1166,6 +1172,11 @@ const UI = {
         // the server, we need to do it here as well since
         // UI.disconnect() won't be used in those cases.
         UI.connected = false;
+
+        if (UI.wakelock !== null) {
+            UI.wakelock.release();
+            UI.wakelock = null;
+        }
 
         UI.rfb = undefined;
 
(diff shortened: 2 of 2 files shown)


CendioZeijlon:
There could absolutely be cases where this can be useful, but we can't assume that all users want to prevent their screen from locking when they are connected to a session.

It should probably be an opt-in feature.

tsukasa-au:
> </snip>
> It should probably be an opt-in feature.

100% agree.

This feature is controlled by the newly added `use_wakelock` option. The default value of this option is `false`.

If you want, I can try to plumb this configuration option into the (advanced) settings in the UI. This would also require me to add support for toggling the configuration at runtime.

CendioZeijlon:
> If you want, I can try to plumb this configuration option into the (advanced) settings in the UI. This would also require me to add support for toggling the configuration at runtime.

That sounds great! :smiley:

CendioZeijlon:
How are you progressing on this? :slightly_smiling_face:

tsukasa-au:
Thanks for checking up on this.

I have rewritten this, but am currently working on tests. I am at a conference this/next week, but will try to get the tests finished after that.

I have pushed my new changes to this PR, but I want to clean them up a little:
  - add documentation for the new WakeLockManager class.
  - add a couple more tests
  - moving the test-related code into the test commit (the event related code, that allows the tests to track the state machine transitions).
  - move the list of valid state transitions into the test.

tsukasa-au:
Actually... I ended up finding some time to hack on the last few things I wanted with this. I have updated this PR.

Not sure if you would like me to squash these down commits.

tsukasa-au:
This change should be in a good state to review.

samhed:
Works great! Tested on Android 15

I would suggest we put this option as ON by default.

CendioOssman:
> I would suggest we put this option as ON by default.

What's your thinking here? I'd consider it very misbehaved for any other application to not respect my screen timeout. Applications that are passively used are normally the exception, not the rule.

core/rfb.js
```diff
@@ -413,6 +417,24 @@ export default class RFB extends EventTargetMixin {
         }
     }
 
+    get requestLocalWakelock() {
```
CendioOssman: This seems like an entirely frontend thing. Any reason to put stuff in the `RFB` object?

docs/EMBEDDING.md
```diff
@@ -95,6 +95,10 @@ Currently, the following options are available:
 * `logging` - The console log level. Can be one of `error`, `warn`, `info` or
   `debug`.
 
+* `request_wakelock` - Should we prevent the (local) display from going into
```
CendioOssman: Most users probably don't know what a "wakelock" is. Can we have something more behavioural in the name?

CendioOssman (changes requested):

samhed:
> > I would suggest we put this option as ON by default.
> 
> What's your thinking here? I'd consider it very misbehaved for any other application to not respect my screen timeout. Applications that are passively used are normally the exception, not the rule.

I would personally expect a remote desktop client to keep my screen awake. But perhaps that isnt the case for everyone.

If we dont want to enable this by default, I would at least move it out of "advanced".

CendioOssman:
In what way do you think using remote applications differs from using local ones? Perhaps there is some room for automation here.

samhed:
> In what way do you think using remote applications differs from using local ones? Perhaps there is some room for automation here.

I use desktop applications and mobile applications differently. In my mind the norm is to have a significantly longer screen timeout, if any, on desktop. While a short timeout is probably quite common on mobile.

When connecting to a remote desktop via noVNC I'd like things to act more like a desktop. I'd be very annoyed if my desktop screen automatically locked after 30 seconds of inactivity.

core/rfb.js
tsukasa-au: Before I had created the WakeLockManager object, I needed to track more state that felt like it make more sense being in a class instance rather than a singleton.

Now that I have the other class, I have updated the code to use it directly from `ui.js`.

docs/EMBEDDING.md
tsukasa-au: Done.

I have renamed this `keep_device_awake`.

tsukasa-au:
> I would personally expect a remote desktop client to keep my screen awake. But perhaps that isnt the case for everyone.
> 
> If we dont want to enable this by default, I would at least move it out of "advanced".

For the moment, I will keep the default set to `false` as it is quite easy to override when deploying noVNC, _and_ the setting is saves on the client.

As for moving it out of "advanced", I can if you want... I am just very hesitant to move it, given that most other settings are in "advanced" (like `show_dot`, or the `path` settings).

samhed:
> For the moment, I will keep the default set to `false` as it is quite easy to override when deploying noVNC, _and_ the setting is saves on the client.
> 
> As for moving it out of "advanced", I can if you want... I am just very hesitant to move it, given that most other settings are in "advanced" (like `show_dot`, or the `path` settings).

I'm fine with that. I understand your reasoning for keeping it set to false by default. Having it under advanced or not isn't a big deal. Given that @CendioOssman is satisfied with the review comments, I think we are ready to merge!

app/wakelock.js
```diff
@@ -0,0 +1,199 @@
+/*
```
CendioOssman, line 1: This file is only used from the app, so it should probably be moved to the app directory instead.

tests/test.wakelock.js
```diff
@@ -0,0 +1,198 @@
...
+        this.released = true;
+        this.dispatchEvent(new Event("release"));
+    }
+}
+
+function waitForStateTransition(wakelockManager, newState) {
+    const {promise, resolve} = Promise.withResolvers();
+
+    const eventListener = (event) => {
+        console.warn(`Got state change: ${event.oldState} -> ${event.newState}`);
```
CendioOssman: This should probably be removed as it will generate noise when running the tests.

tsukasa-au: Removed.

app/wakelock.js
tsukasa-au: Done.

tsukasa-au:
> There are some minor details left.

Done.

> And the linter isn't happy.

I believe I have fixed these now.

tsukasa-au:
What are the next steps here? I can see that this change has been approved, but I don't think I am able to merge this into the novnc repo myself. Is there something that I am missing?

samhed:
Apologies for keeping you waiting. I re-tested your PR in Firefox on Android; it still works as advertised!

Well done on sorting the review comments.

Issue #1986, opened 2025-08-30
Delay incremental FramebufferUpdateRequests to reduce network traffic


redneck-f25:
**Is your feature request related to a problem? Please describe.**
I need to observe the screen of a remote device. The screen ist recorded with an USB camera and `ffplay` from another headless device and shared with ThightVNC and noVNC. The interval of relevant changes on the screen is very slow.

As mentioned in *6.4.3 FramebufferUpdateRequest (rfbproto-3.8.pdf)* I need to *regulate the rate at which it sends
incremental FramebufferUpdateRequests to avoid hogging the network* to reduce network traffic over a metered connection.

**Describe the solution you'd like**
A new property `delayIncrementalUpdateRequests` on RFB instances delays incremental `FramebufferUpdateRequest` (client to server) messages after `FramebufferUpdate` (server to client) messages.

If you like the idea I would integrate my current implementation (see below) into `rfb.js` and `vnc.html` and create a PR.

**Describe alternatives you've considered**
None.

**Additional context**

Below is a redacted screenshot from a page with two `<iframe>`s to the same (no)VNC server.
- https://example.org/properties={"compressionLevel":9,"qualityLevel":4}
- https://example.org/properties={"compressionLevel":9,"qualityLevel":4,"delayIncrementalUpdateRequests":500}

Over about 11 minutes I could reduce the transferred bytes to 16% from 93.56 MiB to 14.72 MiB and the number of messages (TCP frames/IP packets) to 14% from 90523 to 12052 with a delay of 500 ms.

<img width="266" height="308" alt="Image" src="https://github.com/user-attachments/assets/ffcdd23b-aa07-47a8-a1b3-35eb43815efe" />

Here is my current implementation to fix this in this use case:

```javascript
const $$screen = document.getElementById('screen');
const rfbUrl = location.origin.replace(/^http/, 'ws') + location.pathname.replace(/[^/]*$/, 'websockify');
// FIXME: DO NOT DO THIS IS REAL LIFE!!!
const rfbOptions = JSON.parse(decodeURIComponent(location.search.match(/[?&]options=([^&]*)/)[1]));
const rfbProperties = JSON.parse(decodeURIComponent(location.search.match(/[?&]properties=([^&]*)/)[1]));
const rfb = Object.assign(new RFB($$screen, rfbUrl, rfbOptions), rfbProperties);
// 6.4.3 FramebufferUpdateRequest (rfbproto-3.8.pdf)
//
[...]

PR #1990, opened 2025-09-12
Crop transferred region of framebuffer

redneck-f25:
Add option `crop_rect` to select the region of the remote framebuffer to be transferred.

Preview: https://download.bitctrl.de/novnc-u6Nm/noVNC-feature-crop-fb/vnc.html?crop_rect=120x48-0-0

See discussion at #1986 

image

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -88,11 +88,14 @@ const extendedClipboardActionNotify  = 1 << 27;
 const extendedClipboardActionProvide = 1 << 28;
 
 // [GEOMETRY SPECIFICATIONS](https://www.x.org/releases/X11R7.7/doc/man/man7/X.7.xhtml#heading7)
+// [X0VNCSERVER −Geometry](https://tigervnc.org/doc/x0vncserver.html#:~:text=is%2060.-,%E2%88%92Geometry%20geometry)
 const geometryRegExp = new RegExp([
     '^',
     '(?<width>0|[1-9][0-9]*)', 'x', '(?<height>0|[1-9][0-9]*)',
+    '(?:',
     '(?:', '[+](?<left>0|[1-9][0-9]*)', '|', '[-](?<right>0|[1-9][0-9]*)', ')',
     '(?:', '[+](?<top>0|[1-9][0-9]*)',  '|', '[-](?<bottom>0|[1-9][0-9]*)', ')',
+    ')?',
     '$',
 ].join(''));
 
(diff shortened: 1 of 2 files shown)


CendioOssman:
Seems reasonable for some niche cases. But I'm worried the UI for this is a bit technical.

I don't think it's appropriate to have a complex syntax in the RFB object. Better if it just takes a DOMRect.

redneck-f25:
What do you mean with _complex syntax_. It is the [well known x11 geometry syntax](https://www.x.org/releases/X11R7.7/doc/man/man7/X.7.xhtml#heading7):
`<width> "x" <height> [ "+" <left> | "-" <right> ] [ "+" <top> | "-" <bottom> ].`

How do you want it? `new DOMRect(x, y, width, height)`
- `crop_rect=3720,2232,120,48`
- `crop_rect=x=3720,y=2232,w=120,h=48`
- `crop_x=3720&crop_y=2232&crop_w=120&crop_h=48`

**EDIT**: Or do you mean how it looks in the settings?
image


Also with CSS it's possible to set width, height, right and bottom attributes.
In ThighVNC users can define listeners for selected areas of the screen with (almost) the same syntax.
Also TigerVNC uses this syntax (https://tigervnc.org/doc/x0vncserver.html).

```
−Geometry geometry

This option specifies the screen area that will be shown to VNC clients. The format is widthxheight+xoffset+yoffset ,
where ‘+’ signs can be replaced with ‘−’ signs to specify offsets from the right and/or from the bottom of the screen.
Offsets are optional, +0+0 is assumed by default (top left corner). If the argument is empty, full screen is shown to
VNC clients (this is the default).
```


I have three screens:
- [3] upper left: w=1920, h=1080, x=-1920, y=0
- [2] upper right: w=1920, h=1080, x=0, y=0 (primary)
- [1] lower right: w=1920, h=1200, x=0, y=1080

To have listeners for the clocks on [1] and [3] I configured TightVNC as follows
- [3] 120x48+-120+1032 at 5902/tcp
- [1] 120x48+1800+2232 at 5901/tcp
image

Now I can observe my clocks
- [3]
  - `vnc.html?path=addr:127.0.0.1:5900&crop_rect=120x48-1920-1200` (which from the lower right corner up the height of my lower display and left the width of my right displays)
  - `vnc.html?path=addr:127.0.0.1:5902`
  - `vnc.html?path=addr:127.0.0.1:5900&crop_rect=120x48+1800-1032` (which is from the upper left corner down the height of my upper displays minus the height of the area and right the with of my left display minus the width of the are, if I need to change the size of the area I need to change the anchor point)
- [1]
  - `vnc.html?path=addr:127.0.0.1:5900&crop_rect=120x48-0-0` (which is just the lower right corner)
  - `vnc.html?path=addr:127.0.0.1:5901`
  - `vnc.html?path=addr:127.0.0.1:5900&crop_rect=120x48+3720+2232` (which is from the upper left corner right the sum of the width of my upper displays minus the width of the displays and down the sum of the height of my right displays minus the height of the area)

BTW: there was a bug `RFB::get cropRect()` if right or bottom is 0 (zero).

image

PR #1991, opened 2025-09-12
Show local cursor and cursor while dragging the clip area

redneck-f25:
Sometimes it's hard to see the small dot if there is no remote cursor, i. e. if in view-only mode.

Add option `show_local_cursor` to show the local cursor while the mouse is over the canvas.
Add option `show_drag_cursor` to show `grab` or `grabbing` cursor while in drag mode ore while dragging.

Preview: https://download.bitctrl.de/novnc-u6Nm/noVNC-feature-local-cursors/vnc.html?show_local_cursor=1&show_drag_cursor=1&show_dot=1&view_clip=1&view_only=1

See discussion at #1986 

EDIT: To see the first seconds of the video below, you may need to open it in another tab (at least in Chrome on Windows).

https://github.com/user-attachments/assets/93d781f4-3e5c-42cb-8f89-858bcc841bfa

### DIFF
--- a/vnc_lite.html
+++ b/vnc_lite.html
@@ -165,6 +165,7 @@
         // Set parameters that can be changed on an active connection
         rfb.viewOnly = readQueryVariable('view_only', false);
         rfb.scaleViewport = readQueryVariable('scale', false);
+        rfb.showLocalCursor = readQueryVariable('show_local_cursor', 0);
     </script>
 </head>
 


CendioOssman:
Disabling the local cursor is discussed in #974 and #1684, and my stance is still that I don't think it's appropriate in noVNC.

Allowing a different fallback cursor is perfectly okay, though. TigerVNC recently added the option of choosing between the dot cursor and the standard system cursor.

Switching to a drag cursor whilst dragging seems natural enough that I think we can do that implicitly, without a setting.

redneck-f25:
There is a cleaned up separate PR for the drag-cursor part #1992.

The local cursor in view-only or with buggy servers is another thing and it's not done the mentioned disuccions. I'll think about it.

PR #1992, opened 2025-09-15
Show local cursor while dragging the viewport

redneck-f25:
See #1991 for details.

Add `view_drag` parameter to enable dragging when connected.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1103,6 +1103,11 @@ const UI = {
         UI.rfb.qualityLevel = parseInt(UI.getSetting('quality'));
         UI.rfb.compressionLevel = parseInt(UI.getSetting('compression'));
         UI.rfb.showDotCursor = UI.getSetting('show_dot');
+        UI.rfb.showLocalCursor = {
+            drag: 'grab',
+            dragging: 'grabbing',
+            empty: 'default',
+        };
 
         UI.updateViewOnly(); // requires UI.rfb
     },


core/util/cursor.js
```diff
@@ -115,6 +115,10 @@ export default class Cursor {
         this._hotSpot.y = 0;
     }
 
+    setLocalCursor(cursor) {
```
CendioOssman: I don't think this fits well with how this class normally operates. I think you'll need to extend the `change()` API. And probably provide a fallback image for whe we cannot rely on a browser rendered cursor.

core/rfb.js
```diff
@@ -3067,6 +3069,11 @@ export default class RFB extends EventTargetMixin {
                             image.hotx, image.hoty,
                             image.w, image.h
         );
+        if (this._dragViewport) {
+            this._cursor.setLocalCursor(this._viewportDragging ? 'grabbing' : 'grab');
+        } else {
+            this._cursor.setLocalCursor('none');
+        }
```
CendioOssman: I was thinking that we only show the drag cursor whilst actively dragging. Otherwise, we want to keep showing the actual cursor to give proper feedback.

core/rfb.js
```diff
@@ -1117,9 +1117,11 @@ export default class RFB extends EventTargetMixin {
 
         let bmask = RFB._convertButtonMask(ev.buttons);
 
-        let down = ev.type == 'mousedown';
+        let down = false;
```
CendioOssman, line 1155: What does this improve?

app/ui.js
```diff
@@ -178,6 +178,7 @@ const UI = {
         UI.initSetting('password');
         UI.initSetting('autoconnect', false);
         UI.initSetting('view_clip', false);
+        UI.initSetting('view_drag', false);
```
CendioOssman, line 181: We already have the toolbar button for this. What's the purpose of this change?

CendioOssman (changes requested):

core/util/cursor.js
redneck-f25: Thanks for the hint.

Reverted with 59e40b443552a882c9ead0518e89fd5baaf502c6
Proposal b5b7ebe7345d8ac085f1fa6e195ca1b605c85ed6

core/rfb.js
redneck-f25: It doesn't make that many sense to only show a local cursor whilst dragging the viewport. There might be other reasons for a local cursor in different situations. See below.

Reverted with 254fff94acc708f3829a52f3cf20b3df43a91be9

core/rfb.js
redneck-f25: This is a minor non-functional optimization.

First, the event type is compared against the word `mousedown` and the result is stored in the local variable `down`.
Second, in case of the text value of the event type we switch to the responsible code block for`'mousedown`, `mouseup` and `mousemove`.

As the commit message says: [Do not compare to 'mousedown' on every 'mousemove'](https://github.com/novnc/noVNC/pull/1992/commits/57829159839e7a2358f5c22a53414565f5704087).

I guess `mousemove` is the most frequently invoked block, so it should be the first case block.

```diff
--- 0f48f31e08a70045e4fe5deebcac7d7b04fd816d/core/rfb.js
+++ 57829159839e7a2358f5c22a53414565f5704087/core/rfb.js
@@ -1117,11 +1117,9 @@ export default class RFB extends EventTargetMixin {

         let bmask = RFB._convertButtonMask(ev.buttons);

+        let down = false;
-        let down = ev.type == 'mousedown';
         switch (ev.type) {
             case 'mousedown':
+                down = true;
+            // eslint-disable-next-line no-fallthrough
             case 'mouseup':
                 if (this._dragViewport) {
                     this._cursor.setLocalCursor(down ? 'grabbing' : 'grab');
@@ -1157 +1159 @@
                 break;
             case 'mousemove':
```

app/ui.js
redneck-f25: There is a button, but no URL parameter to save this setting as bookmark or enable it automatically when embedding.

`https://example.org/vnc.html?autoconnect=1&view_clip=1&view_drag=1`

Nevertheless, it wasn't working as intended. Fixed with abc7d6310215e520a29730690250b9cef06da1df

redneck-f25:
I dived deeper into the code and came closer to #1991 again.

There are some reasons to show a local cursor:

- We are connected to a buggy server
- We are connected to a device without a cursor (kiosk, POS)
- We are **visually handicapped** or the content of the screen makes it impossible to realize the dot

There are some situations to show a local cursor:

- Whilst dragging the viewport to have a visual feedback
- Whilst dragging is enabled but the mouse is up to distinguish the condition if the control bar is not visible or accessible
- Whilst in view-only mode and the dot is too small
- Whilst the server hasn't send a cursor, from core/rfb.js:
  ```javascript
        // XXX: TightVNC 2.8.11 sends no cursor at all until Windows changes
        // it. Result: no cursor at all until a window border or an edit field
        // is hit blindly. But there are also VNC servers that draw the cursor
        // in the framebuffer and don't send the empty local cursor. There is
        // no way to satisfy both sides.
  ```
- As the 'defalut' cursor

With this PR the RFB class gets a new property `showLocalCursor` which is either `false` or falsy or an `object` with the following properties:
- `default`:
- `viewOnly`:
- `drag`:
- `dragging`:
- `empty`:

The values are set as the cursor for the RFB view as [documented at MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor). So have the options to set an 'embedded', an URL to an image or a data URI.

For the following screen capture I used another cursor set and the following local cursor:
image

```
UI.rfb.showLocalCursor = {
    drag: 'grab',
    dragging: 'grabbing',
    viewOnly: 'help',
    empty: 'url("https://upload.wikimedia.org/wikipedia/commons/f/f5/Cursor_%3F.svg") 10 10, default',
};
```
(As before, there are problems with the first few seconds, try to download or open an another tab.)

https://github.com/user-attachments/assets/3bbae19f-01e9-4960-af37-ec898d82335c

redneck-f25:
There is an export of this branch at https://download.bitctrl.de/novnc-u6Nm/noVNC-feature-drag-viewport-cursor/vnc.html#path=https://localhost:6443/addr:192.0.2.2&view_clip=1&view_drag=1&autoconnect=1

PR #1993, opened 2025-09-16
Add automatic clipboard sync when permissions allow

tobfah:
**Enable automatic clipboard sync when permissions allow -- otherwise
fall back to the clipboard panel.**

This PR adds seamless local-remote clipboard sync in browsers that
support relying on user-granted clipboard permissions. No widgets and
no native paste prompts required.

When proper support is detected, the clipboard panel is removed from the
control bar. In all other cases, the panel remains available as before.

Currently this works in Chromium-based browsers (e.g. Google Chrome,
Microsoft Edge, Chromium). Other families, such as Gecko (Firefox and
friends) and WebKit (Safari and friends), are more restrictive -- in
particular the latter -- regarding what they consider valid user
interactions, which makes it difficult to set up a consistent clipboard
framework.

In addition, they insist on producing the paste prompt on every
clipboard read. When testing this, I felt that this paste prompt provides
a poorer user experience compared to today's clipboard panel.

While not exhaustive, this work provides a foundation for extending
automatic clipboard sync to other browser families in the future.

Tested on
* Fedora 42
  - Chromium 140 ✓
  - Chrome 140 ✓
  - Firefox 142 ✘
  - Gnome Web 48.5 ✘
* Windows 11
  - Chrome 140 ✓
  - Edge 140 ✓
  - Firefox 142 ✘
* macOS 15.6
  - Chrome 140 ✓
  - Edge 140 ✓
  - Firefox 142 ✘
  - Safari 18.6 ✘
* Android 13
  - Chrome 139 ✓
  - Edge 139 ✓
  - Samsung internet 28 ✓
  - Firefox 142 ✘
* iOS 18.6
  - Chrome 140 ✘
  - Edge 140 ✘
  - Firefox 142 ✘
  - Safari 18.6 ✘

Unfortunately all browsers tested on iOS appears to be WebKit skins,
resulting in this feature being disabled across the board.

This work was inspired by PR #1347 from @juanjoDiaz.

### DIFF
--- a/docs/API-internal.md
+++ b/docs/API-internal.md
@@ -18,6 +18,8 @@ keysym values.
 * __Display__ (core/display.js): Efficient 2D rendering abstraction
 layered on the HTML5 canvas element.
 
+* __Clipboard__ (core/clipboard.js): Clipboard event handler.
+
 * __Websock__ (core/websock.js): Websock client from websockify
 with transparent binary data support.
 [Websock API](https://github.com/novnc/websockify-js/wiki/websock.js) wiki page.
@@ -25,10 +27,10 @@ with transparent binary data support.
 
 ## 1.2 Callbacks
 
-For the Mouse, Keyboard and Display objects the callback functions are
-assigned to configuration attributes, just as for the RFB object. The
-WebSock module has a method named 'on' that takes two parameters: the
-callback event name, and the callback function.
+For the Mouse, Keyboard, Display, and Clipboard objects, the callback
+functions are assigned to configuration attributes, just as for the RFB
+object. The WebSock module has a method named 'on' that takes two
+parameters: the callback event name, and the callback function.
 
 ## 2. Modules
 
@@ -81,3 +83,23 @@ None
 | blitImage          | (x, y, width, height, arr, offset, from_queue)          | Blit pixels (of R,G,B,A) to the display
 | drawImage          | (img, x, y)                                             | Draw image and track damage
 | autoscale          | (containerWidth, containerHeight)                       | Scale the display
+
+## 2.3 Clipboard module
+
+### 2.3.1 Configuration attributes
+
+None
+
+### 2.3.2 Methods
+
+| name               | parameters        | description
+| ------------------ | ----------------- | ------------
+| writeClipboard     | (text)            | An async write text to clipboard
+| grab               | ()                | Begin capturing clipboard events
+| ungrab             | ()                | Stop capturing clipboard events
+
+### 2.3.3 Callbacks
+
+| name    | parameters | description
+| ------- | ---------- | ------------
+| onpaste | (text)     | Called following a target focus event and an async clipboard read


core/clipboard.js
```diff
@@ -0,0 +1,44 @@
...
+    async writeClipboard(text) {
+        try {
+            await navigator.clipboard.writeText(text);
+        } catch (error) {
+            Log.Error("Clipboard write failed: ", error);
+        }
+    }
+
+    grab() {
+        if (this._isGrabbed) return;
```
CendioZeijlon: How likely do you think it is that `grab()` will be called after `_isGrabbed` has already been set to `true`? Is it very common, or should we call `Log.Warning()` to notify that this is something that shouldn't occurr?

core/clipboard.js
```diff
@@ -0,0 +1,44 @@
...
+        this._target.addEventListener('focus', this._handleFocus.bind(this));
+        this._isGrabbed = true;
+    }
+
+    ungrab() {
+        if (!this._isGrabbed) return;
```
CendioZeijlon: Same comment here as for `grab()`.

core/util/browser.js
```diff
@@ -11,6 +11,35 @@
...
+        return false;
+    }
+    return false;
+}
+
+export async function isAsyncClipboardAvailable() {
+    return (_isAsyncClipboardSupported() &&
+            await _hasAsyncClipboardPermissions());
+}
+
```
CendioZeijlon, line 46: What are your thoughts on moving this to `clipboard.js`?

core/rfb.js
```diff
@@ -315,8 +321,18 @@ export default class RFB extends EventTargetMixin {
             this._rfbConnectionState === "connected") {
             if (viewOnly) {
                 this._keyboard.ungrab();
+                (async () => {
+                    if (await isAsyncClipboardAvailable()) {
```
CendioZeijlon: Since `isAsyncClipboardAvailable()` needs to always be called before any clipboard actions, couldn't these calls be done as a part of the clipboard functions?

app/ui.js
```diff
@@ -1754,6 +1755,22 @@ const UI = {
...
+            document.getElementById('noVNC_clipboard_button')
+                .removeEventListener('click', UI.toggleClipboardPanel);
+            document.getElementById('noVNC_clipboard_text')
+                .removeEventListener('change', UI.clipboardSend);
+            if (UI.rfb) {
+                UI.rfb.removeEventListener('clipboard', UI.clipboardReceive);
+            }
+        })();
+    },
+
```
CendioZeijlon, line 1782: I wonder if it would be better to keep the button visible but change the behavior so that it becomes clear to users that the _native_ clipboard is enabled.

Less confusion for seasoned users, and the GUI remains consistent if one swaps between browsers with differing clipboard support.. at least until this can be enabled in all popular browsers.

core/rfb.js
CendioZeijlon: Also, what would the consequence be of not doing this check before attempting to, e.g., write to the clipboard?

core/clipboard.js
tobfah: Ideally it should be unlikely. However grabbing clipboard makes sense
doing whenever keyboard is grabbed which currently is done during
connection and when updating the ui based on the state of view-only. A
simple guard avoids adding more listeners than we need and we don't have
to worry about misuse.

core/util/browser.js
tobfah: Well to some extent these are browser gauging utilities. Maybe a frontend doesn't want to import the clipboard module for ui decisions.

core/rfb.js
tobfah: > Since `isAsyncClipboardAvailable()` needs to always be called before any clipboard actions, couldn't these calls be done as a part of the clipboard functions?

Yes that would technically be very similar. The reasoning was shifting responsibility away from the "lib" to the caller under what conditions it's proper to apply the API. It does add a lot of noise on the side of the caller though. We could split the API and permissions checks, but unfortunately it's the latter that ultimately dictates the course and is of asynchronous nature.

> Also, what would the consequence be of not doing this check before attempting to, e.g., write to the clipboard?

If we don't tie writing to a user interaction, depending on the browser it could succeed or silently fail. Same with read, but here proper user interaction is of even greater concern. Best case scenario you'd be presented with the paste prompt on every read attempt.

app/ui.js
tobfah: Yea not a bad idea -- could explore this and see how it turns out.

CendioOssman:
Are there bug reports to Firefox and WebKit detailing what they need to do? Having to maintain two methods indefinitely isn't too appealing.

core/clipboard.js
```diff
@@ -0,0 +1,44 @@
...
+ * Licensed under MPL 2.0 or any later version (see LICENSE.txt)
+ */
+
+import * as Log from './util/logging.js';
+
+export default class Clipboard {
+    constructor(target) {
+        this._target = target;
+        this._isGrabbed = false;
+        this.onRead = () => {};
```
CendioOssman: This is old style event handling. Please use the helper class we have to get the modern style.

core/clipboard.js
```diff
@@ -0,0 +1,44 @@
+/*
+ * noVNC: HTML5 VNC client
+ * Copyright (c) 2025 Tobias Fahleson
+ * Licensed under MPL 2.0 or any later version (see LICENSE.txt)
+ */
+
+import * as Log from './util/logging.js';
+
+export default class Clipboard {
```
CendioOssman: This seems like a lot of code for very little abstraction. Why not just handle the focus event directly in the `RFB` class?

core/clipboard.js
```diff
@@ -0,0 +1,44 @@
...
+
+    grab() {
+        if (this._isGrabbed) return;
+        this._target.addEventListener('focus', this._handleFocus.bind(this));
+        this._isGrabbed = true;
+    }
+
+    ungrab() {
+        if (!this._isGrabbed) return;
+        this._target.removeEventListener('focus', this._handleFocus.bind(this));
```
CendioOssman: I don't think this will work. You'll get a different object on each `bind()`.

tobfah:
> Are there bug reports to Firefox and WebKit detailing what they need to do? Having to maintain two methods indefinitely isn't too appealing.

There aren't really any bugs to report. Regarding reaching out with what behavior we would like to see for our use case -- that hasn't been done yet.

The greatest obstacle is that they insist on the paste prompt on clipboard read, which delivers a poor user experience. They seem to be clear on that point that this isn't going away any time soon.

core/util/browser.js
```diff
@@ -11,6 +11,35 @@
...
+    } catch {
+        return false;
+    }
+    return false;
+}
+
+export async function isAsyncClipboardAvailable() {
+    return (_isAsyncClipboardSupported() &&
+            await _hasAsyncClipboardPermissions());
+}
```
CendioOssman: Can this change during runtime? It would make things a lot cleaner if you just check this during startup and then store the value.

core/rfb.js
```diff
@@ -2208,7 +2224,14 @@ export default class RFB extends EventTargetMixin {
...
         this._resize(width, height);
 
-        if (!this._viewOnly) { this._keyboard.grab(); }
+        if (!this._viewOnly) {
+            this._keyboard.grab();
+            (async () => {
+                if (await isAsyncClipboardAvailable()) {
+                    this._clipboard.grab();
+                }
+            })();
```
CendioOssman: I think this is a bit messy. I would prefer a `.then()` approach like other places where we join sync and async code.

core/rfb.js
```diff
@@ -2323,6 +2346,20 @@ export default class RFB extends EventTargetMixin {
...
+            if (await isAsyncClipboardAvailable()) {
+                await this._clipboard.writeClipboard(text);
+            } else {
+                // Dispatch event for the alternative clipboard
+                this.dispatchEvent(
+                    new CustomEvent("clipboard", {detail: {text: text}})
+                );
+            }
+        })();
+    }
```
CendioOssman, line 2343: Here it is probably cleaner just to make the entire method async.

core/rfb.js
```diff
@@ -315,8 +321,18 @@ export default class RFB extends EventTargetMixin {
             this._rfbConnectionState === "connected") {
             if (viewOnly) {
                 this._keyboard.ungrab();
+                (async () => {
+                    if (await isAsyncClipboardAvailable()) {
+                        this._clipboard.ungrab();
```
CendioOssman: Can't we avoid this fiddling with event handlers by simply checking our state in the focus handler? Would be a lot more neatly contained in a single place.

app/ui.js
CendioOssman: -1 from me. Existing users will eventually learn. But every new user would be confused by a button that does nothing.

app/ui.js
```diff
@@ -9,7 +9,7 @@
 import * as Log from '../core/util/logging.js';
 import _, { l10n } from './localization.js';
 import { isTouchDevice, isMac, isIOS, isAndroid, isChromeOS, isSafari,
-         hasScrollbarGutter, dragThreshold }
+         hasScrollbarGutter, dragThreshold, isAsyncClipboardAvailable }
```
CendioOssman: I wonder if we might need to expose this from the `RFB` object. Our app is not the only user, and this here is an internal function that others should not use.

core/rfb.js
```diff
@@ -2323,6 +2346,20 @@ export default class RFB extends EventTargetMixin {
...
 
+    _writeClipboard(text) {
+        if (this._viewOnly) return;
+        (async () => {
+            if (await isAsyncClipboardAvailable()) {
+                await this._clipboard.writeClipboard(text);
+            } else {
+                // Dispatch event for the alternative clipboard
+                this.dispatchEvent(
+                    new CustomEvent("clipboard", {detail: {text: text}})
```
CendioOssman: The API documentation should be updated to reflect that this is now only used sometimes.

core/rfb.js
```diff
@@ -10,8 +10,8 @@
 import { toUnsigned32bit, toSigned32bit } from './util/int.js';
 import * as Log from './util/logging.js';
 import { encodeUTF8, decodeUTF8 } from './util/strings.js';
-import { dragThreshold, supportsWebCodecsH264Decode, isAsyncClipboardAvailable
-       } from './util/browser.js';
+import { dragThreshold, supportsWebCodecsH264Decode,
+         isAsyncClipboardAvailable } from './util/browser.js';
```
CendioOssman: Merge formatting fixes with the commit it fixes.

core/rfb.js
```diff
@@ -271,6 +271,7 @@ export default class RFB extends EventTargetMixin {
 
         this._clipboard = new Clipboard(this._canvas);
         this._clipboard.onRead = this.clipboardPasteFrom.bind(this);
+        this._isAsyncClipboardAvailable = isAsyncClipboardAvailable;
```
CendioOssman: No hacks like this, please. The tests should mock any external things that need to be adjusted.

tests/test.browser.js
```diff
@@ -1,6 +1,55 @@
 import { isMac, isWindows, isIOS, isAndroid, isChromeOS,
          isSafari, isFirefox, isChrome, isChromium, isOpera, isEdge,
-         isGecko, isWebKit, isBlink } from '../core/util/browser.js';
+         isGecko, isWebKit, isBlink,
+         isAsyncClipboardAvailable } from '../core/util/browser.js';
+
+describe('Async clipboard', function () {
```
CendioOssman, line 6: Tests should be included with the code changes. So merge this with the relevant commit.

tests/test.clipboard.js
```diff
@@ -0,0 +1,40 @@
...
+        sinon.stub(navigator.clipboard, "readText").resolves();
+        await clipboard._handleFocus(new Event("focus"));
+        sinon.assert.calledOnce(navigator.clipboard.readText);
+    });
+
+    it("_handleFocus triggers onRead with read text", async function () {
+        const text = "random text 123";
+        sinon.stub(navigator.clipboard, "readText").resolves(text);
+        clipboard.onRead = sinon.spy();
+        await clipboard._handleFocus(new Event("focus"));
```
CendioOssman: Please don't use internal stuff in tests.

tests/test.rfb.js
```diff
@@ -3466,18 +3466,62 @@ describe('Remote Frame Buffer protocol client', function () {
             });
         });
 
+        describe('_writeClipboard()', function () {
+
+            it('calls clipboard.writeClipboard if async clipboard is available', async function () {
+                client._isAsyncClipboardAvailable = async () => true;
+                client._clipboard = { writeClipboard: sinon.stub().resolves() };
+                const text = 'text to system clipboard';
+                await client._writeClipboard(text);
```
CendioOssman: Same thing here.

CendioOssman (changes requested):

CendioOssman:
> There aren't really any bugs to report. Regarding reaching out with what behavior we would like to see for our use case -- that hasn't been done yet.

If the behaviour differs between browsers, then that is considered a bug. They usually even have specific categories for that. I think they might even have a browser neutral repository to coordinate things.

> The greatest obstacle is that they insist on the paste prompt on clipboard read, which delivers a poor user experience. They seem to be clear on that point that this isn't going away any time soon.

That does not seem entirely true? I can paste things in Google Docs without any prompt.

I hope you don't mean that Firefox and Safari have actively rejected this behaviour? If so, then I'm sceptical about merging this. We should write noVNC for web standards. Not for Chrome specific extensions.

core/util/browser.js
CendioZeijlon: I see what you mean. It's not strictly connected to your clipboard class so I guess it could exist where it is right now. I also saw that there are other checker functions for other features in `browser.js`.

app/ui.js
CendioZeijlon: This is a bit of a side track, but don't you think it would be confusing for new users that mainly use a Chromium-based browser, hence being used to seamless copy/paste, if they at some point use a browser that doesn't support this new feature?

They would probably feel like copy/paste has stopped working when it, in fact, is hiding in the side menu.

Shouldn't we have some mechanism or visual tell for showing them whether it's enabled or not?

core/clipboard.js
CendioZeijlon: Alright! Also I don't know if my question is connected to @CendioOssman's comment
> Can this change during runtime? It would make things a lot cleaner if you just check this during startup and then store the value.

tobfah:
> > There aren't really any bugs to report. Regarding reaching out with what behavior we would like to see for our use case -- that hasn't been done yet.
> 
> If the behaviour differs between browsers, then that is considered a bug. They usually even have specific categories for that. I think they might even have a browser neutral repository to coordinate things.
> 
Sure, inconsistencies like this may be interpreted/categorized as bugs.
> > The greatest obstacle is that they insist on the paste prompt on clipboard read, which delivers a poor user experience. They seem to be clear on that point that this isn't going away any time soon.
> 
> That does not seem entirely true? I can paste things in Google Docs without any prompt.
> 
As far as I can tell, docs (and likely similar apps) don't use the async clipboard API. You're never poked for permissions, and resetting or denying the clipboard permission appears to have no effect. The document also appears to be listening for 'paste' events, so my guess would be they're handling trusted copy/paste events triggered for specific elements such as input, textarea, divs with editable content, etc.
> If so, then I'm sceptical about merging this. We should write noVNC for web standards. Not for Chrome specific extensions.

Sure it's not optimal to have branched mechanisms.

> I hope you don't mean that Firefox and Safari have actively rejected this behaviour? 

Well I base that conclusion in part on this text https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations. The key takeaway (supported by testing) is that Firefox and Safari require user interaction, and don't support passive permissions as alternatives. Even if you'd manage that aspect, you're still faced with the paste prompt on every read, something that Chrome browsers doesn't display as far as I've been able to gauge.

core/util/browser.js
tobfah: Agree that would be cleaner. I don't think it should be able to change during runtime, as you're prompted to reload the page to apply the updated setting.

core/clipboard.js
tobfah: Yea I guess the class turned out quite slim.

core/clipboard.js
tobfah: Aside from the examples I gave I'm not sure, but I don't think so. However that other comment by @CendioOssman,

> I don't think this will work. You'll get a different object on each bind().

shines light on the fact that we should probably just bind once and refer to a stored reference.

core/util/browser.js
tobfah: > as you're prompted to reload the page to apply the updated setting.

Not when answering the permissions dialog, that is. Reload is prompted when manually altering/resetting the provided Block or Allow.

core/clipboard.js
CendioZeijlon: > shines light on the fact that we should probably just bind once and refer to a stored reference.

Agreed!

app/ui.js
samhed: Switching browsers feels almost like switching operating systems nowadays, quite a big decision for most users. I don't think it is common.

Don't show the clipboard button at all when automatic sync is available.

samhed:
Since we have a situation where Firefox+Safari and Chrome have behaved differently regarding the clipboard for years. In my mind, the goal should be to have two different mechanisms in noVNC that both try to make things as seamless as possible.

I view this PR as the first step: making clipboard easy on Chrome (and derivatives). But I would perhaps not label this as the _long-term solution_ that we hope will be the future in all browsers. A next step would be needed that attempts to work around this _paste_ prompt in Safari and Firefox.

There seem to have been two camps for a while here; we might have to accept that we will need two different approaches and user experiences. We have to start somewhere, and this new mechanism has the old clipboard as a fallback.

webysther:
Applied this PR as patch and works as expected: https://github.com/jlesage/docker-baseimage-gui/issues/178

image

webysther:
> Since we have a situation where Firefox+Safari and Chrome have behaved differently regarding the clipboard for years. In my mind, the goal should be to have two different mechanisms in noVNC that both try to make things as seamless as possible.
> 
> I view this PR as the first step: making clipboard easy on Chrome (and derivatives). But I would perhaps not label this as the _long-term solution_ that we hope will be the future in all browsers. A next step would be needed that attempts to work around this _paste_ prompt in Safari and Firefox.
> 
> There seem to have been two camps for a while here; we might have to accept that we will need two different approaches and user experiences. We have to start somewhere, and this new mechanism has the old clipboard as a fallback.

Yeap, things are incremental... For me wrost than that is put in hold a feature just because its not complete, almost no software come to live with this thinking.

dyno-nexsoft:
How to sync clipboard/shortcut Window to MacOS
image

core/clipboard.js
tobfah: I'm not really sure what the issue is. This follows the pattern we see for example with `keyboard` instance in `rfb`,
```
this._keyboard = new Keyboard(this._canvas);
this._keyboard.onkeyevent = this._handleKeyEvent.bind(this);
```
i.e. an injected callback to `rfb`, where `onRead` is equivalent to `onkeyevent`.

Note: in the latest PR iteration (see [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb)), I changed the label `onRead` to `onpaste`, in an attempt to more closely follow the naming conventions.

tests/test.clipboard.js
tobfah: This has been addressed in the tests, see [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb)

tests/test.browser.js
tobfah: This has been addressed, see [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb)

core/clipboard.js
tobfah: This has been addressed, see [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb)

core/util/browser.js
tobfah: > Can this change during runtime? It would make things a lot cleaner if you just check this during startup and then store the value.

In [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb), I moved the gatekeeping from rfb to the relevant clipboard methods. Instead of unconditionally querying permissions, it's only done if we don't have an earlier result to lean on. 

Due to potential race conditions between the clipboard object instantiation and adding of event listeners and focus handling, we have to provide the permissions query option in grab() and _handleFocus() in case we haven't been able to determine async clipboard availability yet.

core/rfb.js
tobfah: This has been addressed in [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb) and [03466a9](https://github.com/novnc/noVNC/pull/1993/commits/03466a96b1a190f98daa31cc74f0ec572ab1fe48)

tests/test.rfb.js
tobfah: This has been addressed in the tests, see [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb)

core/rfb.js
tobfah: This is redundant now as rfb no longer imports the utility function in question, see [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb)

core/rfb.js
tobfah: The tests have been updated with this comment in mind, see [fa3557b](https://github.com/novnc/noVNC/pull/1993/commits/fa3557b59a2d40efb9d8e6e3378b25a3ec2c07cb)

core/clipboard.js
tobfah: I think the separation adds to readability of rfb.js as it is quite large as it is, and improves the distinction between the async clipboard operations and the vnc clipboard mechanisms. Perhaps refactoring all clipboard logic into clipboard.js wouldn't be such a bad idea.

app/ui.js
```diff
@@ -1754,6 +1755,31 @@ const UI = {
...
     },
 
+    updateClipboard() {
+        getAsyncClipboardStatus()
+            .then((status) => {
+                if (status === 'unsupported') {
+                    // Use fallback clipboard panel
+                    return;
+                }
+                if (status === 'denied' || status === 'available') {
```
jlesage: Should we also fallback to the clipboard panel when status is `denied` ? User might not want to allow automatic read of its clipboard, but would still use it manually.

core/util/browser.js
```diff
@@ -11,6 +11,32 @@
...
+    }
+    try {
+        const writePerm = await navigator.permissions.query(
+            {name: "clipboard-write", allowWithoutGesture: false});
+        const readPerm = await navigator.permissions.query(
+            {name: "clipboard-read",  allowWithoutGesture: false});
+        if (writePerm.state === "denied" || readPerm.state  === "denied") {
+            return 'denied';
+        }
+        if ((writePerm.state === "granted" || writePerm.state === "prompt") &&
```
jlesage, line 37: Why do we consider `prompt` as a valid state? While in this state, the permission is effectively denied.

app/ui.js
tobfah: Well the idea is that when automatic clipboard sync is available in the browser of choice, that is the only clipboard mechanism offered, which may be allowed or blocked. It is probably confusing for users that there's still an active clipboard to some extent even though they seemingly explicitly blocked clipboard, since the distinction between the two mechanisms might not be clear to everyone.

core/util/browser.js
tobfah: Yes you're right that while the permissions state is `prompt`, clipboard operations are blocked. These permissions function as a feature detection for browsers that support relying on permissions and as a result are less strict about the level of required user interactions. The most noticeable effect is the paste prompt which there doesn't seem to be a reliable way of detecting preemptively, except for the aforementioned clipboard permissions support.

app/ui.js
samhed: @tobfah agreed. If the user denies clipboard permissions, we shouldn't display the clipboard button.

core/util/browser.js
samhed: This isn't about detecting if the clipboard currently works; it's about detecting if such a state is available in the browser.

app/ui.js
jlesage: > Well the idea is that when automatic clipboard sync is available in the browser of choice, that is the only clipboard mechanism offered

Does this mean that we won't support users denying read access to their clipboard ? Looks like a legitimate use case for people wanting privacy.

tobfah:
> Looks nice! Tested, it works like a charm.
> 
> I look forward to a follow-up, where we attempt to make the experience better for Safari and Firefox users.

@samhed Thanks for the review. I had to make a final correction to the queried state of the `clipboard-write` permission (also added an associated unit test for that) so that it correctly reflects our intent and actual usage: `allowWithoutGesture: true`

Doesn't change the behavior of the clipboard at all, but the gatekeeping check is more in line with how we use the API now.

Did some cursory testing on chrome and firefox and appears to still be working as intended.

app/ui.js
CendioZeijlon: > Does this mean that we won't support users denying read access to their clipboard ? Looks like a legitimate use case for people wanting privacy.

@tobfah, @samhed. Does the current/old clipboard mechanism allow denying read access to the (local?) clipboard?

core/util/browser.js
CendioZeijlon: Still, maybe this should be made more clear. Perhaps a docstring or a slightly reworded function name?

samhed: Yeah, `browserClipboardSupport` or something

app/ui.js
samhed: Well, the question isn't relevant, since it doesn't do it automatically in the old mechanism.

samhed:
> Great work!
> 
> @CendioOssman and @samhed. Do you have further comments that you want to have addressed? Otherwise, I think we should merge!

In general, I think this approach is the correct way forward. Once we have the remaining comments sorted, I think we're ready for a merge.

core/util/browser.js
tobfah: Sure we can likely improve the function semantics a bit. And an explanatory comment is always good.

app/ui.js
CendioZeijlon: Ah of course :smile:

core/util/browser.js
CendioZeijlon: Looks good now I think!

CendioOssman:
I think merging this is premature if we don't have a long term plan in place. noVNC should be for the web, not for ChromeNet<TM>. We've already experienced what that looks like with Internet Explorer.

This should not be released until we have an idea on how to bring Safari and Firefox in to the fold eventually. IMO. This might not be a path they are willing to go, like they've expressed with other Chrome only technologies.

At the very least, I'd like to see bug reports where we can get clear feedback from Safari and Firefox on their stance.

app/ui.js
```diff
@@ -1754,6 +1755,31 @@ const UI = {
         }
     },
 
+    updateClipboard() {
+        browserAsyncClipboardSupport()
```
CendioOssman, line 1759: This is an internal method, so it will not be available for other users of noVNC. We probabably need some other interface to indicate if noVNC will handle the clipboard on its own, or if it needs help.

Neustradamus:
@tobfah: Excellent, good job, thanks!

A lot of people from @juanjoDiaz PR must be happy:

PR #2004, opened 2025-10-20
Allow noVNC to enter fullscreen in Safari on macOS

CendioZeijlon:
The fullscreen limitation in Safari seems to only concern Safari on iOS. On macOS it works just fine.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -146,8 +146,8 @@ const UI = {
 
     initFullscreen() {
         // Only show the button if fullscreen is properly supported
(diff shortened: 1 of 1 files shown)


samhed:
iOS on iPhones and iPadOS on iPads have diverged in the past years, but I guess the situation regarding fullscreen is the same on both platforms?

CendioZeijlon:
I didn't go into too much depth when testing. Only IPad was tested. The plan is to do a more complete mapping before removing the _Draft_ tag.

PR #2006, opened 2025-10-28
Display warning prompt before closing tab

tobfah:
Some key combinations appear to be intercepted by browsers, such as Ctrl+W, which closes a tab. Incidentally this in particular is widely used, e.g. for deleting the last word in a terminal or to cut text in emacs. Mobile browsers don't seem to have this problem at all, as I can perform this combination perfectly fine without closing the current tab.

This PR adds a guard to prevent accidentally disconnecting an active session by triggering a "Leave site?" warning prompt. It's implemented to only trigger for an active connection, since that's when it's particularly irritating having to reconnect. Note that closing the tab manually also triggers the warning prompt, though not in mobile browser. The "beforeunload" event in question is allegedly much less reliably fired in mobile browsers, but as mentioned above, they don't exhibit the same fundamental problem.

Tested in
- macOS 15.7
  - Safari 26
- Fedora 42
  - Firefox 144
  - Chromium 141
- iOS 26
  - Safari 26
- Android 13
  - Firefox 144
  - Chrome 141

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1155,6 +1155,8 @@ const UI = {
         UI.showStatus(msg);
         UI.updateVisualState('connected');
 
+        UI.updateBeforeUnload();
+
         // Do this last because it can only be used on rendered elements
         UI.rfb.focus();
     },
@@ -1191,6 +1193,8 @@ const UI = {
             UI.showStatus(_("Disconnected"), 'normal');
         }
 
+        UI.updateBeforeUnload();
+
         document.title = PAGE_TITLE;
 
         UI.openControlbar();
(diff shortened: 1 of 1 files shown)


app/ui.js
```diff
@@ -1211,6 +1215,24 @@ const UI = {
...
+        // considered legacy. The custom string is ignored by modern browsers,
+        // which display a native message, but older browsers will show it.
+        e.preventDefault();
+        e.returnValue = _("Are you sure you want to disconnect the session?");
+    },
+
+    updateBeforeUnload() {
+        // Remove first to avoid adding duplicates
+        window.removeEventListener("beforeunload", UI.handleBeforeUnload);
+        if (UI.connected) {
```
CendioZeijlon: A nitpick, but wouldn't it be better if this function checked the truthiness of an input argument instead of  a class variable?

In both places where you are calling the function, there is now an implicit dependency on that `UI.connected` has to be set before `updateBeforeUnload()` does what you want it to do.

tobfah: Sure, that enhances readability a bit. It also facilitates refactoring the function as a more general utility in the future, if we'd like to apply the same treatment in other places, such as the login form.

redneck-f25:
Hi,

this might be a helpful feature for work-sessions with remote input. For observing remote devices in view-only-mode this will be useless and annoying. The viewOnly flag should also be checked: `if (shouldWarn && !(UI.rfb.viewOnly))`

CendioZeijlon:
> this might be a helpful feature for work-sessions with remote input. For observing remote devices in view-only-mode this will be useless and annoying. The viewOnly flag should also be checked: if (shouldWarn && !(UI.rfb.viewOnly))

That's a great idea! @tobfah could you take a look at this?

tobfah:
> That's a great idea! @tobfah could you take a look at this?

I agree it's a good idea. Pushed an update accordingly.

app/ui.js
CendioZeijlon: I think I'll need to abandon my initial opinion here. Now that you have more than one UI property to check, I think it's better in this case to just make the check directly in the function.

There are already other places where this is done, so that would follow the style thats already being used in `ui.js`, e.g., `updatePowerButton()`.

tobfah: > I think I'll need to abandon my initial opinion here. Now that you have more than one UI property to check, I think it's better in this case to just make the check directly in the function.
> 
> There are already other places where this is done, so that would follow the style thats already being used in `ui.js`, e.g., `updatePowerButton()`.

Yea I had also missed that last bit -- let's follow that precedence for now. Pushed an update accordingly.

PR #2013, opened 2025-11-05
fix mac screenshare vnc

alisade:


CendioOssman:
Thank you for your suggestion. Unfortunately, the browsers have stated that HTTPS is a hard requirement for the web moving forward. It is the new baseline, rather than something extra.

As such, we don't think it is worthwhile to try to find workarounds for this when we start using new web features, like WebCrypto.

PR #2015, opened 2025-11-07
Showing the notification when the clipboard is updated.

Ankit8969:
Whenever the clipboard is updated, I am showing the notification.
which will enhance the user experience.


redneck-f25:
@Ankit8969 I like the idea to get notified if the clipboard has been updated. I send you [PR Ankit8969#1](https://github.com/Ankit8969/noVNC/pull/1) with some modifications
- merge lates commits from upstream
- use existing UI and API for the notification

- add a setting to enable or disable this feature

- add support for Clipboard module introduced with fb7e891841e2d1f8dd2d07e43a38de97b49ade94 (PR #1993)

[this is a link to compare the result with upstream](https://github.com/novnc/noVNC/compare/master...bitctrl:noVNC:feature/clipboard_updated_notification)

Ankit8969:
Thanks @redneck-f25

Ankit8969:
@redneck-f25 what is the next step for this PR ?

Neustradamus:
@Ankit8969: Nice idea!

samhed:
I tested this, and it works, but I can't really see the use case. Why do you guys want this?

Ankit8969:
@samhed  
This will enhance the user experience. For example, when a user selects something in the terminal while the clipboard panel is closed, there is no clear indication of whether the text has been copied or not. Adding this provides clear feedback to the user, making the behavior more intuitive.

samhed:
I tested it again, and unfortunately I think it would be too annoying. I don't want this type of notification.

PR #2016, opened 2025-11-08
feat(tests): move from karma to vite

### LINKED ISSUE #1883: Karma is deprecated (and broken)
**Describe the bug**
Our test runner, Karma, has been deprecated for over a year. It is now starting to break, with Chai moving to modules and none of the bindings between Karma and Chai seem to be maintained to handle this transition.

**To Reproduce**
1. Run `npm test`
2. Tests fail to run, complaining about module import/export

**Expected behavior**
Tests run just fine.

N/A

N/A

**Additional context**
https://github.com/karma-runner/karma#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes

NefixEstrada:
This closes #1883

There are 3 tests that now fail. Other than the performance one, I'm not sure the others are ok, maybe they were already failing? They're marked with a `TODO`.
I wasn't able to run the tests using karma

Please, let me know if there's something that I'm missing!


CendioOssman:
This is a bit more than just replacing Karma, if I understand things correctly. This also replaces Mocha, Chai and Sinon with vitest?

That requires a bit more consideration if we want to replace all of those familiar components and become fully dependent on a single one.

NefixEstrada:
Yes, this would change all the tecnologies to vitest. However:
- Mocha: the syntax is 100% supported, meaning no change for the developer
- Chai: the majority the syntax (assertions) is supported (and the unsupported ones change very little)

There's a comparison in the vitest docs: https://vitest.dev/guide/comparisons.html#mocha

Let me know if there's anything I can do!

PR #2019, opened 2025-11-11
Add noVNC_setting_warn_before_close

### LINKED ISSUE #2014: Disable close confirmation dialog
In the latest beta (1.7.0) a new dialog was added to confirm closing the tab, but there is no setting to disable it and restore the old behaviour. 

Due to my workflow this dialogue box is very annoying, and I don't understand why it is enabled by default without any option to disable it.

redneck-f25:
Ability to to toggle the warning prompt before closing the tab or window introduced by 63c2c14a50d60ac60686c28fd938a1d6e36cbf71 (PR #2006).

Closes #2014

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -184,6 +184,7 @@ const UI = {
         UI.initSetting('shared', true);
         UI.initSetting('bell', 'on');
         UI.initSetting('view_only', false);
+        UI.initSetting('warn_before_close', false);
         UI.initSetting('show_dot', false);
         UI.initSetting('path', 'websockify');
         UI.initSetting('repeaterID', '');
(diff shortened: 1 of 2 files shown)


CendioZeijlon:
Alright, now I've had time to test this after some delay...

Codewise, I think it looks good! But there is one thing that bothers me. When trying different combinations with the two options (**View only** and **Warn before close**), things look a bit ambiguous.

If **view only** is on, then **Warn before close** stays toggled in whatever position it was before view only was turned on. This can make it seem like the latter is on, and that you just can't turn it off without first disabling **view only**. So it's not obvious to users that there is a dependent behavior between the two options, especially that **Warn before close** gets implicitly disabled when **View only** is on.

I therefore wonder if it wouldn't be better to make these two options entirely independent of each other.

@redneck-f25 What do you think?

CendioZeijlon:
@CendioOssman You had some opinions about this when we talked outside GitHub. Care to share?

kroese:
Why is this still not merged? :(

CendioOssman:
> @CendioOssman You had some opinions about this when we talked outside GitHub. Care to share?

Mostly that this seems like a band-aid rather than a fix. The problem is that #2006 ended up being too broad, not that it fundamentally needs a setting.

The goal was to prevent *accidental* closing. E.g. pressing Ctrl+W and expecting it to be sent to the session, not close the tab. The goal was not to prevent *explicit* closing. E.g. clicking the X on the tab.

The technical difficulty is how to tell the difference between the two. A setting could be acceptable until we figure that out.

PR #2026, opened 2025-11-19
added the reconnect timeout

Ankit8969:
This update introduces a maximum reconnect timeout for the VNC client. Currently, when the VNC client attempts to reconnect, it may continue retrying indefinitely if the connection keeps failing, which is not ideal.

With this change, we add a new configuration key, reconnect_max_time, in default.json. This setting allows us to define a maximum duration for reconnection attempts. Once the specified time limit is reached, the client will stop trying to reconnect, preventing unnecessary retries and improving overall stability.


Ankit8969:
@samhed, please check this PR

app/ui.js
```diff
@@ -1121,6 +1125,22 @@ const UI = {
...
+        if (UI.firstReconnectTime === null) {
+            UI.firstReconnectTime = Date.now();
+        }
+        const elapsedTime = Date.now() - UI.firstReconnectTime;
+
+                // Check if we've exceeded the max reconnect time
+        if ((Date.now() - UI.firstReconnectTime) >= maxTime) {
+            // hiding the previous status message
+            UI.hideStatus();
+            UI.showStatus(_("Maximum reconnect attempts reached. Failed to connect to the server."), 'error', 1000*60*240); // Show for 4 hours
```
samhed: Why would this be limited in how long it is shown?

app/ui.js
```diff
@@ -1121,6 +1125,22 @@ const UI = {
...
         UI.reconnectCallback = null;
+        const maxTime = UI.getSetting('reconnect_max_time') ?? 600000; // 20s - 20 * 1000 ms
+
+        // Initialize first reconnect time if it's the first attempt
+        if (UI.firstReconnectTime === null) {
+            UI.firstReconnectTime = Date.now();
+        }
+        const elapsedTime = Date.now() - UI.firstReconnectTime;
+
+                // Check if we've exceeded the max reconnect time
```
samhed: Indentation looks off here

samhed (changes requested):
Thanks for your contribution. I didn't test it yet, but a few comments from briefly checking the code.

app/ui.js
Ankit8969: Fixed the indentation

app/ui.js
Ankit8969: I have changed it to a long time.

samhed: Did you test that code? Such a timeout executes instantly for me. See https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#maximum_delay_value

I'd suggest modifying showStatus to do what you want instead.

Ankit8969: Thanks @samhed 
I have updated the code and tested it.
Please check

defaults.json
```diff
@@ -1 +1,3 @@
-{}
+{
+  "reconnect_max_time": 600000
+}
```
samhed: The idea is that this should be empty. You will still get the same behavior since you added a default when calling `UI.getSetting()`.

defaults.json could be used by integrators of the noVNC app to provide other defaults.

app/ui.js
```diff
@@ -1121,6 +1125,23 @@ const UI = {
 
     reconnect() {
         UI.reconnectCallback = null;
+        const maxTime = UI.getSetting('reconnect_max_time') ?? 600000; // default to 10 minutes => 10 * 60 * 1000 ms
+
+        // Initialize first reconnect time if it's the first attempt
+        if (UI.firstReconnectTime === null) {
+            UI.firstReconnectTime = Date.now();
+        }
+        const elapsedTime = Date.now() - UI.firstReconnectTime;
```
samhed: It seems you're not using this variable.

app/ui.js
```diff
@@ -961,6 +962,9 @@ const UI = {
             UI.closePowerPanel();
         }
     },
+    resetFirstReconnection(){
```
samhed: missing space before {

app/ui.js
```diff
@@ -1154,6 +1175,8 @@ const UI = {
         }
         UI.showStatus(msg);
         UI.updateVisualState('connected');
+        // Here we can reset the retry count
```
samhed: misleading comment

app/ui.js
```diff
@@ -1121,6 +1125,23 @@ const UI = {
 
     reconnect() {
         UI.reconnectCallback = null;
+        const maxTime = UI.getSetting('reconnect_max_time') ?? 600000; // default to 10 minutes => 10 * 60 * 1000 ms
+
+        // Initialize first reconnect time if it's the first attempt
+        if (UI.firstReconnectTime === null) {
```
samhed, line 1132: `UI.firstReconnectTime` is never null, meaning your logic never works. Did you mean to check for 0?

app/ui.js
```diff
@@ -1121,6 +1125,23 @@ const UI = {
 
     reconnect() {
         UI.reconnectCallback = null;
+        const maxTime = UI.getSetting('reconnect_max_time') ?? 600000; // default to 10 minutes => 10 * 60 * 1000 ms
```
samhed: Perhaps this should be configured to seconds instead of milliseconds?

I.e:
```
const maxTime = UI.getSetting('reconnect_max_time') ?? 600; // default to 10 minutes => 10 * 60 seconds

...

// Check if we've exceeded the max reconnect time
if ((Date.now() - UI.firstReconnectTime) >= maxTime * 1000) {
```

samhed (changes requested):
I tested this and unfortunately, this breaks the reconnection functionality. No matter what I set `reconnect_max_time` to, the first time reconnect() is called, we abort the reconnection attempts. This is because `UI.firstReconnectTime` is never set.

I get the impression that you're not testing your code?

Ankit8969 (commented):
I have resolved the comments. Please review.

app/ui.js
```diff
@@ -1121,6 +1125,23 @@ const UI = {
...
+            UI.firstReconnectTime = Date.now();
+        }
+
+        // Check if we've exceeded the max reconnect time
+        if ((Date.now() - UI.firstReconnectTime) >= MAX_TIME_IN_MS) {
+            // hiding the previous status message
+            UI.hideStatus();
+            // Showing this message for long time, because if the connection is unstable and user is not around to see the message when the connection is lost, they will be able to see it when they will come back. (Showing for 3 hours)
+            UI.showStatus(_("Maximum reconnect attempts reached. Failed to connect to the server."), 'error', 180 * 60 * 1000);
+            UI.updateVisualState('disconnected');
```
samhed, line 1142: This should be `UI.cancelReconnect()` to avoid ending up in a broken state after reaching the maximum reconnect attempts.

app/ui.js
```diff
@@ -1121,6 +1125,23 @@ const UI = {
...
+        // Initialize first reconnect time if it's the first attempt
+        if (UI.firstReconnectTime === null) {
+            UI.firstReconnectTime = Date.now();
+        }
+
+        // Check if we've exceeded the max reconnect time
+        if ((Date.now() - UI.firstReconnectTime) >= MAX_TIME_IN_MS) {
+            // hiding the previous status message
+            UI.hideStatus();
+            // Showing this message for long time, because if the connection is unstable and user is not around to see the message when the connection is lost, they will be able to see it when they will come back. (Showing for 3 hours)
```
samhed, line 1140: It looks like the rest of the file avoids long lines like this one.

samhed (changes requested):
Thank you, I tested again, and it's getting closer. However, the state you end up in after reaching the threshold is broken.

Please, test yourself as well, so we don't need these long feedback cycles.

PR #2041, opened 2026-01-25
feat(rfb): add encoding, FPS and bandwidth event emissions

alarraz:
- Add encodingchange event when encoding type changes
- Add FPS counter event emitted every second
- Add bandwidth tracking in websock.js (bytes sent/received)
- Add bandwidth event with download/upload kbps every 2 seconds

### DIFF
--- a/core/rfb.js
+++ b/core/rfb.js
@@ -187,6 +187,12 @@ export default class RFB extends EventTargetMixin {
             encoding: null,
         };
 
+        // Encoding and FPS tracking
+        this._lastEncoding = null;
+        this._frameCount = 0;
+        this._lastFpsTime = performance.now();
+        this._lastBandwidthTime = performance.now();
+
         // Mouse state
         this._mousePos = {};
         this._mouseButtonMask = 0;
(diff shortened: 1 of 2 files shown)


core/rfb.js
```diff
@@ -2663,6 +2669,13 @@ export default class RFB extends EventTargetMixin {
                 this._FBU.encoding = this._sock.rQshift32();
                 /* Encodings are signed */
                 this._FBU.encoding >>= 0;
+
+                // Emit encoding change event for real encodings (not pseudo-encodings)
+                if (this._lastEncoding !== this._FBU.encoding && this._FBU.encoding >= 0) {
+                    this._lastEncoding = this._FBU.encoding;
+                    this.dispatchEvent(new CustomEvent("encodingchange",
+                        { detail: { encoding: this._FBU.encoding } }));
+                }
```
CendioOssman, line 2678: Encodings can change multiple times per update, so this is not a productive approach.

What information is it you want to present here?

core/rfb.js
```diff
@@ -2675,6 +2688,25 @@ export default class RFB extends EventTargetMixin {
...
 
+        // FPS counter
+        this._frameCount++;
+        const now = performance.now();
+        if (now - this._lastFpsTime >= 1000) {
+            this.dispatchEvent(new CustomEvent("fps",
+                { detail: { fps: this._frameCount } }));
+            this._frameCount = 0;
+            this._lastFpsTime = now;
+        }
```
CendioOssman, line 2699: Statistics is messy and people will likely want to measure things in different ways with different degrees of accuracy.

Can't we just expose the raw data and let the caller deal with how to aggregate that?

I.e. expose a frame counter and the number of transferred bytes.

core/websock.js
```diff
@@ -68,6 +68,34 @@ export default class Websock {
...
         };
+
+        // Bandwidth tracking
+        this._bytesReceived = 0;
+        this._bytesSent = 0;
+        this._bandwidthStartTime = performance.now();
+    }
+
+    // Bandwidth statistics
+    getBandwidthStats() {
```
CendioOssman, line 79: Please also update the documentation with all API changes.

PR #2042, opened 2026-02-09
Fix H264 hardware video decoding failure

kdxcxs:
### Summary

H264 decoding via WebCodecs `VideoDecoder` silently fails when hardware acceleration is used (e.g. Intel iGPU with D3D11 on Windows). The screen stays blank with no errors in the console. This PR fixes two issues:

1. **`self` vs `this` typo in `H264Context.decode()`** — `self` refers to `window` in browser context, so SPS parameters (profile, constraint set, level) were being set on the global object instead of the `H264Context` instance, preventing the decoder from ever being configured.

2. **Render queue deadlock with hardware video decoder** — The display render queue blocks on unready video frames (`ready = false`), which triggers the `_flushing` backpressure mechanism in `rfb.js`, stopping all VNC message processing. This starves the `VideoDecoder` pipeline of input. Hardware decoders (unlike software decoders) may buffer frames internally for reordering (e.g. H264 High profile allows B-frames), requiring continued input before producing output. The result is a deadlock:

   ```
   Decoder waiting for more input → render queue waiting for decoder output
   → _flushing stops VNC messages → no more input fed to decoder → deadlock
   ```

### Changes

- **`core/decoders/h264.js`**: Replace `self._profileIdc` / `self._constraintSet` / `self._levelIdc` with `this.*`
- **`core/display.js`**:
  - Video frames that aren't ready no longer block the render queue. Instead, they register an async callback to draw when the decoder produces output, allowing subsequent frames to continue being fed to the decoder.
  - The `flip` operation waits for all pending video frames to resolve before executing (`Promise.all`), preserving visual correctness.
  - Extract `_drawVideoFrame()` helper to deduplicate the frame drawing logic.

### Root cause analysis

Confirmed via `chrome://tracing` on Windows 10 with Intel UHD Graphics:

- The D3D11 hardware decoder accepted 3 H264 frames (1 keyframe + 2 delta) with `kOk` status
- `DoDecode` completed successfully, `CreatePictureBuffers` was called
- `BeginScopedWriteAccess` was called 3 times but `EndScopedWriteAccess` was never called — the decoder held frames in its internal buffer
- Zero `OutputResult` events, zero JS output callbacks
- After ~6 seconds of silence, `VideoDecoder::Shutdown` was called

The render queue's blocking pattern was originally designed for `Image` objects (Tight/TightPNG encoding), where each image decodes independently. The same pattern was applied to video frames when H264 support was added, but `VideoDecoder` is a pipeline that requires continuous input flow — blocking the queue cuts off the input and creates a deadlock.

### Tests

- [x] Verify H264 decoding works with hardware acceleration on Intel iGPU (Windows)
- [x] Verify H264 decoding still works with software decoding fallback
- [x] Verify no visual tearing or frame ordering issues during H264 playback

### DIFF
--- a/core/decoders/h264.js
+++ b/core/decoders/h264.js
@@ -196,9 +196,9 @@ export class H264Context {
             }
 
             if (parser.profileIdc !== null) {
-                self._profileIdc = parser.profileIdc;
-                self._constraintSet = parser.constraintSet;
-                self._levelIdc = parser.levelIdc;
+                this._profileIdc = parser.profileIdc;
+                this._constraintSet = parser.constraintSet;
+                this._levelIdc = parser.levelIdc;
             }
 
             if (this._decoder === null || this._decoder.state !== 'configured') {
@@ -206,12 +206,12 @@ export class H264Context {
                     Log.Warn("Missing key frame. Can't decode until one arrives");
                     continue;
                 }
-                if (self._profileIdc === null) {
+                if (this._profileIdc === null) {
                     Log.Warn('Cannot config decoder. Have not received SPS and PPS yet.');
                     continue;
                 }
-                this._configureDecoder(self._profileIdc, self._constraintSet,
-                                       self._levelIdc);
+                this._configureDecoder(this._profileIdc, this._constraintSet,
+                                       this._levelIdc);
             }
 
             result = this._preparePendingFrame(timestamp);
--- a/core/display.js
+++ b/core/display.js
@@ -16,6 +16,7 @@ export default class Display {
 
         this._renderQ = [];  // queue drawing actions for in-order rendering
         this._flushPromise = null;
+        this._pendingFrames = [];  // video frames awaiting decoder output
 
         // the full frame buffer (logical canvas) size
         this._fbWidth = 0;
@@ -479,6 +480,22 @@ export default class Display {
         }
     }
 
+    _drawVideoFrame(pendingFrame, rect) {
+        let frame = pendingFrame.frame;
+        if (!frame) {
+            return;
+        }
+        if (frame.codedWidth < rect.width || frame.codedHeight < rect.height) {
+            Log.Warn("Decoded video frame does not cover its full rectangle area. Expecting at least " +
+                      rect.width + "x" + rect.height + " but got " +
+                      frame.codedWidth + "x" + frame.codedHeight);
+        }
+        this.drawImage(frame,
+            0, 0, rect.width, rect.height,
+            rect.x, rect.y, rect.width, rect.height);
+        frame.close();
+    }
+
     _renderQPush(action) {
         this._renderQ.push(action);
         if (this._renderQ.length === 1) {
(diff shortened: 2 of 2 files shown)


core/decoders/h264.js
```diff
@@ -196,22 +196,22 @@ export class H264Context {
             }
 
             if (parser.profileIdc !== null) {
-                self._profileIdc = parser.profileIdc;
-                self._constraintSet = parser.constraintSet;
-                self._levelIdc = parser.levelIdc;
+                this._profileIdc = parser.profileIdc;
+                this._constraintSet = parser.constraintSet;
+                this._levelIdc = parser.levelIdc;
             }
```
CendioOssman, line 202: A bit embarrassing that this was overlooked.

It suggests that we are lacking one or more unit tests for this code. Is it something you could have a look at?

core/display.js
```diff
@@ -533,32 +564,20 @@ export default class Display {
...
-                        const dw = sw;
-                        const dh = sh;
-                        this.drawImage(frame, sx, sy, sw, sh, dx, dy, dw, dh);
-                        frame.close();
-                    } else {
+                    if (!a.frame.ready) {
+                        // Don't block the queue — the video decoder
+                        // pipeline needs continued input to produce
+                        // output. Register a callback to draw later,
+                        // and let the queue keep feeding the decoder.
```
CendioOssman, line 571: This gives us tearing, so I'm afraid it's not an approach we want to use.

B-frames should ideally be rare when used with VNC, as they will never be rendered and hence pointless to waste resources on.

But I can accept that you might not always have that control and we need to be prepared to deal with them.

Ideally, we queue them in the decoder rather than the display. Hopefully, we can detect these there?

CendioOssman (changes requested):
Thanks for your contribution!

core/display.js
kdxcxs: To clarify, this issue is not caused by B-frames. The stream I tested uses H264 High profile with only IDR and P-frames, yet Intel's hardware decoder still doesn't output frames immediately. It buffers a few frames of input before producing any output. It seems is Intel-specific; NVIDIA and AMD GPUs output frames without delay. Since Intel iGPUs are very common (laptops, office machines), I think it's worth handling.

I agree that handling this in the display layer is not ideal. I'll look into addressing it on the decoder side (`h264.js`) instead. I'll update the PR once I have a revised approach.

CendioOssman: That would be a violation of the protocol, and could cause all kinds of weird rendering effects. Frames need to be displayed immediately after they are received. It's incorrect to buffer them and display them later.

So we need to find some way to get the Intel decoder to spit out frames right away. Perhaps there is some "low latency mode"?

Failing that, we'd need to blacklist it. Can we detect it and fall back to software decoding?

kdxcxs: I've looked into what's available from the WebCodecs API:

- **`optimizeForLatency: true`** is already set in our `configure()` call, but Intel's decoder seems ignores it.
- **`VideoDecoder.flush()`** forces all pending outputs, but per spec it sets `[[key chunk required]] = true` afterwards, meaning the next `decode()` must provide a key frame. Not suitable for calling after every frame in a continuous stream with P-frames.

As far as I can tell, there's no other way from JavaScript to change Intel's buffering behavior. The only practical option is detection and fallback to software decoding via `hardwareAcceleration: "prefer-software"`. This does come with a performance cost, but at least it produces correct output.

For detection, we can query the GPU vendor via WebGL's `WEBGL_debug_renderer_info` extension (widely available, though blocked in some privacy-hardened browsers).

any1: According to spec, constrained baseline profile is prescribed, so B frames are explicitly not allowed.

PR #2043, opened 2026-02-18
Enable attaching the control bar to top and bottom

### LINKED ISSUE #1471: dock 'noVNC' control box on either top/bottom of the window rather than left/right only
I was bothered by the extensive box the 'noVNC' control box shows when I try to copy/paste in my BU SCC OnDemand Desktop session. And the control box seems can only be put on either left or right side of the window. 

I wonder if the following would be made possible: 
1. make the box can be put on top or bottom of the window
2. make the copy/paste text box resizable with scroll bar. 

No really

**Additional context**
I am attaching a screen shot for your reference. 
newFeatureReq20201002-ys

tobfah:
This PR enables attaching the control bar to the top and bottom positions, in addition to the traditional left and right positions.

Tested on 
- Fedora 42
  - Firefox 147
  - Chromium 144
- macOS 26.3
  - Safari 26.3
  - Chrome 144
- Win 11
  - Edge 145
- Android 13
  - Chrome 144
- iPad 26.3
  - Safari 26.3

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -37,6 +37,8 @@ const UI = {
 
     controlbarGrabbed: false,
     controlbarDrag: false,
+    controlbarMouseDownClientX: 0,
+    controlbarMouseDownOffsetX: 0,
     controlbarMouseDownClientY: 0,
     controlbarMouseDownOffsetY: 0,
 
@@ -110,8 +112,11 @@ const UI = {
         }
 
         // Restore control bar position
-        if (WebUtil.readSetting('controlbar_pos') === 'right') {
-            UI.toggleControlbarSide();
+        const pos = WebUtil.readSetting('controlbar_pos');
+        if (['left', 'right', 'top', 'bottom'].includes(pos)) {
+            UI.toggleControlbarSide(pos);
+        } else {
+            UI.toggleControlbarSide('left');
         }
 
         UI.initFullscreen();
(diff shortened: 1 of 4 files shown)


CendioHanli:
Tested the implementation on Epiphany v40.2 and Chromium v144 on Fedora 43 and found no issues. Regarding the decision to not use a relative offset for the `noVNC_control_bar` in the top and bottom positions look good. I found no instance where this caused any weird edge cases.
LGTM!

app/ui.js
```diff
@@ -586,14 +586,52 @@ const UI = {
         }
 
         const anchor = document.getElementById('noVNC_control_bar_anchor');
-        if (anchor.classList.contains("noVNC_right")) {
+        const anchorAndDescendants = [anchor, ...anchor.querySelectorAll('*')];
```
samhed: Varför behöver vi jobba med descendants här? Borde inte css-selectors lösa det?

vnc.html
samhed, line 1: Nämn gärna att denna commit är förberedande för att tillåta att flytta control bar

samhed (changes requested):
Funkar bra, har testat på Chrome (Linux) och Firefox (Android), mycket trevlig funktion!

Ifall slutresultatet är tänkt att ha uppdelade commits bör du nog nämna i alla commits vilken relation de har till huvudproblemet.

app/ui.js
tobfah: Löste det med css istället

vnc.html
tobfah: Förenklade uppdelningen av commits

tobfah:
I've restructured the PR a bit. A few changes since the initial one:

- The default left position returns as an implicit fallback for robustness (noVNC_left reverted)
- Rely less on local storage
- Separation of commits have been simplified and should make more sense on their own

Not sure why clipboard tests keep failing though.

tobfah:
The last force push fixed some alignment issues with the handle arrow when control bar is in horizontal modes (top/bottom).

I've ran the tests on a RHEL 10.1 with nodejs 22.22.0 and npm 10.9.4, and they all pass on my side, so I'm not sure what to do with those that keep failing here.

PR #2046, opened 2026-03-21
Fix typo

ArduFish:


CendioOssman:
Thanks for your contribution! However, `tlclient.pot` is an automatically generated file. It's `vnc.html` that needs to be changed instead.

"textarea" is also a rather technical term that is best avoided.

ArduFish:
> Thanks for your contribution! However, `tlclient.pot` is an automatically generated file. It's `vnc.html` that needs to be changed instead.
> 
> "textarea" is also a rather technical term that is best avoided.

Fixed.

PR #2047, opened 2026-03-28
feat: add configurable `ignore_keys` setting to filter key events

tjasko:
## Overview
Introduce a new `ignore_keys` setting that allows users to prevent specific keyboard inputs from being forwarded to the remote VNC host, while still allowing the browser to handle them locally.

This addresses common UX issues where browser level shortcuts (for example Escape to exit fullscreen) are also sent to the VM, potentially interrupting workflows or triggering unintended actions.

**Key behavior**:
- Keys listed in `ignore_keys` are handled locally and never sent to the server
- The default is empty, so no keys are ignored and existing behavior is unchanged
- The value is a comma separated list, with surrounding whitespace tolerated
- Aliases are accepted, for example "esc", "ctrl" and "cmd", and are mapped to canonical codes
- Matching is case insensitive

**Implementation details**:
- `supportedIgnoreKeys` centralizes the keys that can be ignored, along with their aliases
- `normalizeIgnoreKey()` maps an alias or label to its canonical form
- `shouldIgnoreKey()` checks a key code against the current setting
- `filterIgnoredKeys()` wraps `rfb.sendKey()` when a connection is set up, so that ignored keys are dropped before reaching the server
- `validateIgnoreKeysInput()` flags unrecognized entries as the setting is changed

**UI changes**:
- New "Ignored keys" text input under the "Advanced" section of the settings panel
- The supported keys are documented through the input's `title` attribute, using `&#013;` for line breaks
- Invalid entries are marked with a red border

**Tests**:
- New `tests/test.ui.js` covering `shouldIgnoreKey()`, verifying that aliases resolve to the right key and that keys outside `supportedIgnoreKeys` are never ignored
- `app/ui.js` added to the files served by karma, so that it can be imported from tests

**Known limitation**: filtering happens inside `rfb.sendKey()`, which is also the path used by the keyboard toolbar buttons. Ignoring `esc` therefore also disables the toolbar Esc button, and ignoring `ctrl`, `alt` or `del` breaks the Ctrl+Alt+Del button. Filtering in `Keyboard._handleKeyEvent()` instead would avoid this, at the cost of a change in core.

This change improves usability when interacting with fullscreen mode and other browser level shortcuts, while remaining backward compatible.

### DIFF
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -612,12 +612,15 @@ html {
     list-style: none;
     padding: 0px;
 }
-#noVNC_settings button,
 #noVNC_settings select,
 #noVNC_settings textarea,
 #noVNC_settings input:not([type=checkbox]):not([type=radio]) {
     margin-left: 6px;
     /* Prevent inputs in settings from being too wide */
+    max-width: calc(100% - 6px - var(--input-xpadding) * 2);
+}
+#noVNC_settings button:not(#noVNC_ignore_keys_help_button) {
+    margin-left: 6px;
     max-width: calc(100% - 6px - var(--input-xpadding) * 2);
 }
 
@@ -641,6 +644,83 @@ html {
 /* Connection controls */
 :root:not(.noVNC_connected) #noVNC_disconnect_button {
     display: none;
+}
+
+/* Help tooltips  */
+.noVNC_setting_with_help {
+    position: relative;
+    display: flex;
+    align-items: center;
+    gap: 0.4rem;
+}
+
+.noVNC_setting_with_help label {
+    flex: 0 1 auto;
+}
+
+#noVNC_settings .noVNC_tooltip_list {
+    margin: 0.4rem 0 0 0;
+    padding-left: 0.8rem;
+    list-style: disc;
+}
+
+.noVNC_tooltip_list li {
+    margin: 0.15rem 0;
+}
+
+#noVNC_setting_ignore_keys::placeholder {
+    color: var(--novnc-grey);
+    opacity: 0.7;
+}
+
+#noVNC_setting_ignore_keys.noVNC_invalid,
+#noVNC_setting_ignore_keys.noVNC_invalid:focus {
+    border-color: var(--novnc-red);
+    box-shadow: 0 0 2px rgba(var(--novnc-red-rgb), 0.5);
+}
+
+#noVNC_ignore_keys_help_button {
+    flex: 0 0 auto;
+
+    width: 1.6rem;
+    height: 1.6rem;
+    min-width: 1.6rem;
+    max-width: 1.6rem;
+
+    padding: 0;
+    margin: 0;
+
+    border-radius: 50%;
+    background: transparent;
+    border: 1px solid rgba(0,0,0,0.25);
+    cursor: pointer;
+
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+
+    line-height: 1;
+}
+
+#noVNC_ignore_keys_tooltip {
+    display: none;
+    position: absolute;
+    z-index: 1000;
+    left: 1rem;
+    padding: 0.5rem;
+    margin: 0;
+    border-radius: 0.4rem;
+    background: rgba(20, 20, 20, 0.96);
+    color: #fff;
+    box-shadow: 0 0.4rem 1rem rgba(0, 0, 0, 0.35);
+    font-size: 0.75rem;
+    line-height: .9rem;
+}
+
+#noVNC_ignore_keys_tooltip.noVNC_open {
+    display: flex !important;
+    flex-direction: column;
+    align-items: flex-start;
 }
 
 /* ----------------------------------------
--- a/app/styles/base.css
+++ b/app/styles/base.css
@@ -612,15 +612,12 @@ html {
     list-style: none;
     padding: 0px;
 }
+#noVNC_settings button,
 #noVNC_settings select,
 #noVNC_settings textarea,
 #noVNC_settings input:not([type=checkbox]):not([type=radio]) {
     margin-left: 6px;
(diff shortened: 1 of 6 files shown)


app/styles/base.css
```diff
@@ -643,6 +646,83 @@ html {
...
+}
+
+#noVNC_setting_ignore_keys::placeholder {
+    color: var(--novnc-grey);
+    opacity: 0.7;
+}
+
+#noVNC_setting_ignore_keys.noVNC_invalid,
+#noVNC_setting_ignore_keys.noVNC_invalid:focus {
+    border-color: var(--novnc-red);
```
samhed, line 678: I don't see why we need our own color for this, regular "red" looks fine.

app/ui.js
```diff
@@ -802,13 +811,28 @@ const UI = {
                 }
             }
         } else {
-            ctrl.value = value;
+            ctrl.value = value ?? '';
```
samhed, line 814: What is this?

app/ui.js
```diff
@@ -1593,6 +1631,25 @@ const UI = {
         UI.rfb.sendKey(keysym, code, down);
     },
 
+    wrapRfbSendKey() {
```
samhed, line 1634: `filterIgnoredKeys()` ?

app/ui.js
```diff
@@ -383,6 +385,13 @@ const UI = {
         UI.addSettingChangeHandler('logging', UI.updateLogging);
         UI.addSettingChangeHandler('reconnect');
         UI.addSettingChangeHandler('reconnect_delay');
+        UI.addSettingChangeHandler('ignore_keys');
+        const input = document.getElementById('noVNC_setting_ignore_keys');
+        if (input && !input.dataset.validationBound) {
```
samhed, line 390: Why have you added this `validationBound` attribute? I can't see any reason for why these two handlers need such a guard if no others do.

We should probably get rid of the "if" and the variable as well and follow the style of the other handlers.

app/ui.js
```diff
@@ -802,13 +811,28 @@ const UI = {
...
+            ctrl.value = value ?? '';
         }
     },
 
-    // Save control setting to cookie
-    saveSetting(name) {
-        const ctrl = document.getElementById('noVNC_setting_' + name);
+    // Update cookie and form control setting. If value is not set, then
+    // updates from control to current cookie setting.
+    saveSetting(nameOrEvent) {
```
samhed, line 820: The changes to this function makes no sense to me - `saveSetting()` doesn't ever seem to be called with an event?

vnc.html
```diff
@@ -220,6 +220,21 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br>VNC</h1>
                         </label>
                     </li>
                     <li><hr></li>
+                    <li class="noVNC_setting">
```
samhed, line 223: Most users won't need this, I think it fits best under the "advanced" section of the settings.

vnc.html
```diff
@@ -220,6 +220,21 @@ <h1 class="noVNC_logo" translate="no"><span>no</span><br>VNC</h1>
...
+                        <div class="noVNC_setting_with_help">
+                            <label for="noVNC_setting_ignore_keys">Ignored Keys:</label>
+
+                            <button id="noVNC_ignore_keys_help_button" type="button">?</button>
```
samhed, line 227: This large circled question mark doesn't fit the rest of our interface. I'd also like to avoid infrastructure for tooltips.

Couldn't we simply use the html "title" attribute to get almost the same? (With `&#013;` we can have line breaks in the title attribute.)

samhed (changes requested):
Thank you for contributing! I think this could be a useful feature in some cases.

I tested it in Chrome 145 on Fedora 43. Functionally, it works as advertised.

The code needs some work though, see the comments.

app/styles/base.css
tjasko: Agreed, there was no need for it. I dropped both the `--novnc-red-rgb` and `--novnc-red` variables and used plain `red` instead. I also removed the box-shadow, the border colour on its own is enough to signal an invalid entry.

app/ui.js
tjasko: Left over from an earlier iteration where the setting could still be undefined. It serves no purpose now, so I reverted it back to `ctrl.value = value`.

app/ui.js
tjasko: Renamed to `filterIgnoredKeys()`. While doing so I also dropped the `_ignoreKeysWrapped` guard, since the function only runs immediately after a fresh RFB object is created and so there was never anything to guard against.

app/ui.js
tjasko: You are right, there was no reason for it. I removed the guard, the variable and the two manual listeners. Validation is now registered with `UI.addSettingChangeHandler("ignore_keys", UI.validateIgnoreKeysInput)`, exactly like the other settings.

app/ui.js
tjasko: Correct, it never is. This was left over from the earlier validation handling, which at one point passed the event through. Since that is gone, I reverted `saveSetting()` and `updateSetting()` back to their original form.

vnc.html
tjasko: Agreed, moved it into the "Advanced" section, just below the Repeater ID field.

vnc.html
tjasko: Done. The help button, the tooltip element, its CSS and the JavaScript that built it are all gone, replaced by a plain `title` attribute on the input using `&#013;` for the line breaks. That also let me remove `initIgnoreKeysTooltip()`, `buildIgnoreKeysTooltipText()` and `buildIgnoreKeysPlaceholder()`. The supported keys are now listed directly in vnc.html, along with a short placeholder on the input itself.

PR #2052, opened 2026-05-12
Add optional username field for auth

Micket:
We see no reason not to support username.  Necessary for e.g. when using PAM auth with wayvnc. 

Disclaimer: I had this patched into our OpenOndemand repo, which still was on noVNC 1.3.0, so I haven't tested this with master branch, but the only change I saw was `WebUtil` -> `UI` and the changes are rather trivial.

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -1047,6 +1047,12 @@ const UI = {
             password = undefined;
         }
 
+        let username = UI.getSetting('username');
+
+        if (username === null) {
+            username = undefined;
+        }
+
         UI.hideStatus();
 
         UI.closeConnectPanel();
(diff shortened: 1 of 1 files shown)


samhed:
We already support username, see setCredentials():
https://github.com/novnc/noVNC/blob/5c2025d5a68d3f35cd1c06a8d88fcba06e0f8abb/app/ui.js#L1321

Micket:
@samhed 
You don't support usernames when calling  `connect` , it only passes password as the sole credential:
https://github.com/novnc/noVNC/blob/5c2025d5a68d3f35cd1c06a8d88fcba06e0f8abb/app/ui.js#L1084-L1088
which triggers when autoconnecting via url parameters:
```
vnc.html?autoconnect=true&path=hostname&password=12345678  <-- no username support?
```

I believe the code path you is when it's using it's built-in login widget

PR #2053, opened 2026-05-18
Replace top-level await with then() in browser.js

### LINKED ISSUE #2034: Endless loading screen on Android 16 with Chromium browsers
### **Describe the bug**
I encountered this problem the first time when I tried to use this docker container [jlesage/docker-jdownloader-2](https://github.com/jlesage/docker-jdownloader-2) 
It uses noVNC to provide a GUI view of JDownloader 2. 
This worked flawlessly and I was able to use every browser on PC and on Android until I recently changed my phone (OnePlus 13). 
Now I have the most recent Android version 16, but can't use noVNC on any Chromium browser. They won't get past the loading screen. 
Gecko browser still seem to work as expected. 

While researching this issue I found your demo at https://novnc.com/noVNC/vnc.html, which showed the same behaviour. So I don't think the docker container is causing this. 

---

### **To reproduce**
1. Use a chromium based browser on Android 16. 
2. Go to [noVNC Demo](https://novnc.com/noVNC/vnc.html)
3. See the endless loading screen

---

### **Expected behavior**
I would expect to see the noVNC logo and controls after some seconds. 
(On my Android 10 device, Huawei P30 Pro, it works just like that.)

---

### **Client:**
 - OS: Android 16; CPH2653 Build/BP2A.250605.015
 - Browser: chrome, brave, kiwi, ...
 - Chrome version: 143.0.7499.146
 - Brave version: 1.85.117, Chromium 143.0.7499.146

### **Server:**
 - noVNC version: your demo
 - VNC server: your demo
 - WebSocket proxy: your demo

---

### Additional Notes
Merry Christmas to all of you! 🎄
Thanks for developing such an awesome tool.

zk-wz:
On certain Android Chromium browsers, `Decoder.flush()` may hang indefinitely. As a result, the associated promise remains in a pending state, preventing the page from loading because the top‑level `await` blocks module evaluation.

### DIFF
--- a/core/util/browser.js
+++ b/core/util/browser.js
@@ -182,7 +182,10 @@ async function _checkWebCodecsH264DecodeSupport() {
 
     return true;
 }
-supportsWebCodecsH264Decode = await _checkWebCodecsH264DecodeSupport();
+
+_checkWebCodecsH264DecodeSupport().then(result => {
+    supportsWebCodecsH264Decode = result;
+});
 
 /*
  * The functions for detection of platforms and browsers below are exported
--- a/core/util/browser.js
+++ b/core/util/browser.js
@@ -183,7 +183,9 @@ async function _checkWebCodecsH264DecodeSupport() {
     return true;
 }
 
-_checkWebCodecsH264DecodeSupport().then(result => {
+// FIXME: Avoid top-level await due to a Chromium bug where Decoder.flush()
+// can hang indefinitely on some Android devices, blocking module evaluation.
+_checkWebCodecsH264DecodeSupport().then((result) => {
     supportsWebCodecsH264Decode = result;
 });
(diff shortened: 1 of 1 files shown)


samhed (changes requested):
Nice! This fixes #2034 for me. But since this is quite a strange change, we should mention the reason in a FIXME comment in the code. It does look like a Chromium bug after all.

CendioOssman:
I'm afraid this workaround introduces a race condition, so this is not the proper fix. :/

@samhed, can you reopen things and either revert or find a more robust fix?

CendioOssman:
#2061 might have a better diagnosis and fix?

samhed:
Tested and merged #2061 :+1:

PR #2061, opened 2026-05-29
Fix silent module load failure in Chrome 86-93 due to missing VideoDecoder.isConfigSupported

### LINKED ISSUE #2060: `_checkWebCodecsH264DecodeSupport()` causes silent ES module load failure on Chrome 86–93
**Describe the bug**

When `@novnc/novnc` is imported into a bundled web app (e.g. via Vite), the top-level `await` in `core/util/browser.js` causes the **entire ES module to silently fail to load** on Chrome versions 86–93. The result is a completely blank page with no console errors — making the bug extremely hard to diagnose.

**To reproduce**

1. Create a Vite project that statically imports `@novnc/novnc`
2. Build for production (`vite build`)
3. Open the app in Chrome 86–93 (e.g. on an Android phone that hasn't updated Chrome)
4. Observe: blank page, zero console errors, `#app` div is empty

**Expected behavior**

The app should load normally. `supportsWebCodecsH264Decode` should gracefully fall back to `false` on browsers where the WebCodecs API is partially implemented.

**Screenshots**

N/A — the symptom is a blank page with no visible error.

**Client:**
- OS: Android (Galaxy phone)
- Browser: Chrome
- Browser version: 86–93 (WebCodecs origin trial era)

**Server (noVNC):**
- noVNC version: 1.7.0
- VNC server: N/A (issue occurs before any VNC connection)
- WebSocket proxy: N/A

**Additional context**

Root cause is in `core/util/browser.js`:

```js
async function _checkWebCodecsH264DecodeSupport() {
    if (!('VideoDecoder' in window)) {
        return false;  // ✅ guards against missing VideoDecoder
    }

    // ...
    let support = await VideoDecoder.isConfigSupported(config);  // ❌ no guard here
```

`VideoDecoder.isConfigSupported()` was only added in **Chrome 94** when WebCodecs became stable. Chrome 86–93 shipped `VideoDecoder` in `window` (origin trial), but without `isConfigSupported`. So the `if (!('VideoDecoder' in window))` check passes, but the next line throws:

```
TypeError: VideoDecoder.isConfigSupported is not a function
```

Since this is called via top-level `await`, the error propagates to a **module load rejection**. ES module load failures are silent by default — `window.onerror` and `unhandledrejection` do not fire.

Suggested fix in `_checkWebCodecsH264DecodeSupport()`:

```js
// Option A: add a guard
if (!('VideoDecoder' in window) || typeof VideoDecoder.isConfigSupported !== 'function') {
    return false;
}

// Option B: wrap the isConfigSupported call in try/catch
try {
    let support = await VideoDecoder.isConfigSupported(config);
    if (!support.supported) return false;
} catch (e) {
    return false;
}
```

Note: this is separate from issue #1943 (bundler compatibility). That issue is about esbuild refusing to bundle top-level await at lower targets. This issue is a **runtime silent failure** on partially-implemented WebCodecs browsers.

YoongiKim:
## Summary

Fixes #2060

Chrome 86–93 shipped `VideoDecoder` as part of an origin trial but did **not** include `VideoDecoder.isConfigSupported` — that method was only added in Chrome 94. The existing guard in `_checkWebCodecsH264DecodeSupport()` checked for `VideoDecoder` in `window` but not for `isConfigSupported` on the constructor, causing a "is not a function" error at the top-level `await` call site. Because this happens during ES module evaluation, the entire module silently fails to load with no console error — just a blank page.

## Change

```js
// Before
if (!('VideoDecoder' in window)) {
    return false;
}

// After
if (!('VideoDecoder' in window) ||
    !('isConfigSupported' in VideoDecoder)) {
    return false;
}
```

This makes the function gracefully return `false` on browsers that have `VideoDecoder` but lack `isConfigSupported`, instead of throwing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/util/browser.js
+++ b/core/util/browser.js
@@ -107,7 +107,8 @@ export const hasScrollbarGutter = _hasScrollbarGutter;
 export let supportsWebCodecsH264Decode = false;
 
 async function _checkWebCodecsH264DecodeSupport() {
-    if (!('VideoDecoder' in window)) {
+    if (!('VideoDecoder' in window) ||
+        !('isConfigSupported' in VideoDecoder)) {
         return false;
     }
 


samhed (approved):
Thank you!

It works well in Chrome 152 on Android 16.

PR #2062, opened 2026-06-08
Fix control bar HTML structure in vnc.html

GurvaReddy2311:
when we are clicking on noVNC control bar we are getting an Exception popup because a div tag got missed in vnc.html we have tested it from our end after updating it is working fine could you please merge this change

### DIFF
--- a/vnc.html
+++ b/vnc.html
@@ -112,7 +112,9 @@
     <!-- noVNC control bar -->
     <div id="noVNC_control_bar_anchor" class="noVNC_crosscenter">
         <div id="noVNC_control_bar">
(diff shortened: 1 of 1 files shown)


GurvaReddy2311:
Handled popexception when clicked on novnc control bar

GurvaReddy2311:
Handled popup 
image
exception when clicked on novnc control bar

samhed:
That empty `<div></div>` is not a mistake.

I think your error came from mixed cached versions of JS and HTML. Clearing the browser cache often fixes issues like that.

PR #2063, opened 2026-06-08
Fix control bar handle HTML structure

GurvaReddy2311:
When clicked on control bar we are getting an error popup window


samhed:
Duplicate of #2062

PR #2065, opened 2026-06-08
Add trackpad mode for relative pointer input on touch devices

benjaminfrombe:
## Summary

noVNC maps touch input to absolute framebuffer coordinates, which is
awkward on phones/tablets: your finger covers the target, you can't
reach screen edges easily, and there is no way to zoom/pan locally.
This PR adds an opt-in **trackpad mode** that behaves like the iOS RDP
app or a laptop trackpad: one finger moves a virtual cursor relatively,
pinch zooms the view locally (without sending anything to the remote),
and two-finger drag either pans the magnified view or scrolls the remote.

### What it does

Toggle via `RFB.trackpadMode = true` or the new **"Trackpad mode (touch)"**
checkbox in the settings panel.

**One finger**
- Moves the virtual cursor relatively (scaled by `trackpadSensitivity`,
  default 1.5) — driven from raw `touchmove` events so there is no dead
  zone before movement is visible
- Tap = left click at the virtual cursor
- Long-press = right click (and drag = drag with button held)
- Tap immediately followed by drag ("tap-and-a-half") = drag with left
  button held

**Two fingers**
- Pinch = local magnification (1×–5×) via a CSS transform on the canvas
  (GPU-composited, smooth, the remote is never touched)
- While magnified: two-finger drag pans the view toward the fingers
- At fit (1×): two-finger drag sends mouse-wheel scroll events
- Zoom is anchored at the midpoint between the fingers (RDP-style)

**Cursor visibility**
- A dot cursor is always shown in trackpad mode (falls back to the dot
  when the server sends no visible cursor), so you can always see where
  the virtual cursor is

**iOS Safari**
- Blocks the browser's own pinch-to-zoom (which ignores `user-scalable=no`
  since iOS 10) by preventing the proprietary `gesturestart`/`gesturechange`
  events (detected via their `.scale` property, so noVNC's own synthetic
  gesture `CustomEvent`s are unaffected)

### Bug fix included

`gesturehandler.js` `_twoTouchTimeout`: the existing code chose between
PINCH and TWODRAG by comparing finger movement **per axis** against the
inter-finger distance change. A diagonal drag splits its movement across
both axes (~70% each), so it would lose to distance noise and be
misclassified as PINCH even when the fingers stayed parallel. Fixed by
comparing the 2D movement magnitude (`hypot`) instead.

### Implementation notes

- All two-finger and one-finger movement handling is in raw `touchstart/
  move/end` handlers that run alongside the gesture handler. The gesture
  handler retains responsibility for tap classification only.
- The CSS transform (`translate()+scale()`, origin 0 0) is purely visual;
  it does not affect the canvas layout size, so `absX`/`absY` still map
  pointer coordinates correctly to framebuffer pixels.
- The virtual cursor position is kept in untransformed canvas CSS pixels;
  `_trackpadUpdateCursor` maps it through the bounding rect's effective
  scale so the dot follows the logical position even while magnified.
- `trackpadMode` and `trackpadSensitivity` are public properties on `RFB`.
  Leaving trackpad mode resets the CSS transform.

## Test plan

- [ ] Enable "Trackpad mode (touch)" on a mobile browser and verify:
  - [ ] One-finger drag moves the cursor from the first pixel (no dead zone)
  - [ ] Tap clicks, two-finger tap right-clicks
  - [ ] Long-press holds the button; tap-then-drag does the same
  - [ ] Pinch zooms in/out around the midpoint between fingers
  - [ ] While zoomed: two-finger drag pans the view (including diagonally)
  - [ ] At fit: two-finger drag scrolls the remote
  - [ ] Dot cursor is visible without moving first
  - [ ] iOS Safari does not zoom the page on pinch
- [ ] Verify existing desktop behaviour is unaffected (trackpadMode defaults to false)
- [ ] `npm test` passes (654 tests)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/app/ui.js
+++ b/app/ui.js
@@ -191,6 +191,7 @@ const UI = {
         UI.initSetting('bell', 'on');
         UI.initSetting('view_only', false);
         UI.initSetting('show_dot', false);
+        UI.initSetting('trackpad_mode', false);
         UI.initSetting('path', 'websockify');
         UI.initSetting('repeaterID', '');
         UI.initSetting('reconnect', false);
@@ -378,6 +379,8 @@ const UI = {
         UI.addSettingChangeHandler('view_only', UI.updateViewOnly);
         UI.addSettingChangeHandler('show_dot');
         UI.addSettingChangeHandler('show_dot', UI.updateShowDotCursor);
+        UI.addSettingChangeHandler('trackpad_mode');
+        UI.addSettingChangeHandler('trackpad_mode', UI.updateTrackpadMode);
         UI.addSettingChangeHandler('keep_device_awake');
         UI.addSettingChangeHandler('keep_device_awake', UI.updateRequestWakelock);
         UI.addSettingChangeHandler('host');
@@ -1182,6 +1185,7 @@ const UI = {
         UI.rfb.qualityLevel = parseInt(UI.getSetting('quality'));
         UI.rfb.compressionLevel = parseInt(UI.getSetting('compression'));
         UI.rfb.showDotCursor = UI.getSetting('show_dot');
+        UI.rfb.trackpadMode = UI.getSetting('trackpad_mode');
 
         UI.updateViewOnly(); // requires UI.rfb
         UI.updateClipboard();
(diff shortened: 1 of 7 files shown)


tobfah:
Hi, thanks for your contribution. The first impression is that the modularization of this new trackpad feature is lacking -- see for instance how keyboard and cursor are integrated with rfb.

PR #2078, opened 2026-09-01
feat: enable subpath for vnc_lite.html

lufy90:
Enable custom subpath for `vnc_lite.html` entry.

.e.g novnc serve behind nginx reverse proxy like:

```
location ^~ /novnc/ {
    proxy_pass http://127.0.0.1:8888/;
    ...
```

`vnc.html` already works fine in such case.

### DIFF
--- a/vnc_lite.html
+++ b/vnc_lite.html
@@ -126,10 +126,10 @@
         document.getElementById('sendCtrlAltDelButton')
             .onclick = sendCtrlAltDel;
 
-        // Read parameters specified in the URL query string
-        // By default, use the host and port of server that served this file
-        const host = readQueryVariable('host', window.location.hostname);
-        let port = readQueryVariable('port', window.location.port);
+        // Use same logic as vnc.html, but to keep lite, without hash fragment,
+        // encrypt flag, .etc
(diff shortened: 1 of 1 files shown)


tobfah:
Perhaps vnc._lite.html and vnc.html should simply align completely in their URL handling (with the exception of 'encrypt' settings and the likes)?

lufy90:
> Perhaps vnc._lite.html and vnc.html should simply align completely in their URL handling (with the exception of 'encrypt' settings and the likes)?

Thanks for the suggest @tobfah . I've changed in the new force push:

Use same logic for generating websocket URL in `vnc_lite.html` with
    `vnc.html`, to ensure custom subpath works for `vnc_lite.html` entry
    as well.
    
    .e.g novnc serve behind nginx reverse proxy like:
    
    ```
    location ^~ /novnc/ {
        proxy_pass http://127.0.0.1:8888/;
        ...
    ```
    
    To keep lite, hash fragment, encrypt, .etc not included.

tobfah:
Thanks for the quick follow up. I'll just test this out on my end

tobfah:
Tested this out, appears to be working as intended

PR #2081, opened 2026-09-14
Read boolean query parameters properly in vnc_lite.html

### LINKED ISSUE #1662: vnc_lite.html issue when query parameter is set to view_only=false
**Describe the bug**
vnc_lite.html issue when query parameter is set to view_only=false. Can not move  the cursor acting like view_only=true. A work around is to set view_only to nothing eg view_only=

**To Reproduce**
Go to http://cloudtk.tcl-lang.org/view-only.html to see a live demo of the issue.

**Expected behavior**
view_only=false query parameter to allow controlling of the cursor.

 - OS: Linux64 Ubuntu
 - Browser: Firefox
 - Browser version: 101.0.1 (64-bit)

 - noVNC version: 1.3.0
 - VNC server: TigerVNC
 - WebSocket proxy: CloudTk

lbesecker195:
Fixes #1662

`readQueryVariable()` in `vnc_lite.html` returns query values as strings, and those strings went straight into `rfb.viewOnly` and `rfb.scaleViewport`. `?view_only=false` set `viewOnly` to the string `"false"`, and since `RFB` only checks `this._viewOnly` for truthiness, the session ended up view only. `?scale=0` had the same problem.

As @CendioOssman suggested in the issue, this adds a small `readQueryBoolean()` helper and uses it for `view_only` and `scale`. `"0"`, `"no"` and `"false"`, in any case, count as false, and any other value counts as true. That's the same set `UI.getSetting()` uses for checkbox settings in `vnc.html`. If the parameter isn't in the URL, the default applies as before.

## Testing

I loaded `vnc_lite.html` in headless Chromium through Playwright, with the files served from the checkout. There was no VNC server, so the connection fails, but both properties are set right after `RFB` is constructed. To capture the assigned values, I patched the `viewOnly` and `scaleViewport` setters of `core/rfb.js` in the served copy only.

| query | before | after |
|---|---|---|
| none | `false`, `false` | `false`, `false` |
| `view_only=true&scale=true` | `"true"`, `"true"` | `true`, `true` |
| `view_only=1&scale=1` | `"1"`, `"1"` | `true`, `true` |
| `view_only=false&scale=false` | `"false"`, `"false"` | `false`, `false` |
| `view_only=0&scale=0` | `"0"`, `"0"` | `false`, `false` |
| `view_only=no&scale=no` | `"no"`, `"no"` | `false`, `false` |
| `view_only=FALSE&scale=False` | `"FALSE"`, `"False"` | `false`, `false` |
| `view_only=yes&scale=on` | `"yes"`, `"on"` | `true`, `true` |

In the same setup, `vnc.html?autoconnect=true&view_only=...` sets `viewOnly` to the same booleans as the "after" column for all of these values.

`npm run lint` doesn't cover the HTML files. I ran ESLint with the repo config on the inline script, with the HTML indentation removed, and it is clean before and after the change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tobfah:
Thanks for your contribution. I'll have a look at this and test it out

